PR #65, opened 2020-09-18
Migrate tests and commit history from vc-examples

OR13:
- [x] delete old files
- [x] migrate vc-examples tests and commit history


msporny (approved):
I don't know what's going on here -- vc-examples was always supposed to be just example files, not a part of vc-http-api. That said, the work never got traction, so I'm happy to move it into a place that does have traction (the VC test suites).

OR13:
@msporny yep, vc examples was supposed to help people learn, by providing examples. this test suite is for this spec.

peacekeeper (approved):
Agree it makes sense to move this!

PR #89, opened 2021-01-26
OAuth2.0 Client Credentials Request Authorization Support

### LINKED ISSUE #86: Support for OAuth API protection for vendor API
Our (and I would assume other vendors) issuance and verification API's currently require OAuth bearer tokens for request authorization. Adding support to the test suite as an optional feature for vendors using this authorization strategy would prevent them from having to expose un-protected API's. A more advanced implementation might consider a client credentials flow implementation back to the authorization server to obtain the access token rather than it being supplied directly to the test runner. Happy to provide work on this feature if others are not opposed to its addition?

tplooker:
This PR adds basic support for obtaining an access_token for request authorization via the client credentials. Has been tested locally but our environment is not stood up yet so I have not checked in our vendor config that will use this feature. Ideally we need to support the injection of secrets for each vendors config via GH secrets rather than checking it into the repo directly.

To use this feature, add this config element to your vendor config

```
{
   ...,
  request_authorization_config: {
    type: "oauth2-client-credentials",
    token_endpoint: " https://idp.com/oauth/token",
    client_id: "xxxxxxxx",
    client_secret: "xxxxxxxxxx",
    audience: "optional-audience-constraint"
  }
}
```

Fixes #86

### DIFF
--- a/packages/plugfest-2020/.gitignore
+++ b/packages/plugfest-2020/.gitignore
@@ -1,0 +1,1 @@
+node_modules/
--- a/packages/plugfest-2020/help.js
+++ b/packages/plugfest-2020/help.js
@@ -1,10 +1,16 @@
 const fetch = require('node-fetch');
 
-const getJson = async url => {
+const getJson = async (url, requestAuthorization) => {
+  let headers = {
+    Accept: 'application/ld+json',
+  };
+
+  if (requestAuthorization && requestAuthorization.type === "oauth2-bearer-token") {
+    headers.Authorization = `Bearer ${requestAuthorization.accessToken}`
+  }
+
   const res = await fetch(url, {
-    headers: {
-      Accept: 'application/ld+json',
-    },
+    headers,
     method: 'get',
   });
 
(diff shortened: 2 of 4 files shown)


msporny (approved):
PR looks fine -- does this make OAuth2 mandatory? If so, we'd be a -1 -- all authz should be optional, but well defined -- there are use cases that don't require it.

OR13 (approved):

tplooker:
> PR looks fine -- does this make OAuth2 mandatory? If so, we'd be a -1 -- all authz should be optional, but well defined -- there are use cases that don't require it.

Nope def not, if your api's are making use of this as your request authorization layer then you set the config required to fetch a token with OAuth, if not just omit it. I've tried to keep it open through the usage of type to allow other AuthZ schemes to be supported in future

OR13:
These changes look good, this does not make oauth2 mandatory, but we are headed towards being able to allow vendors to show support for a secured set of http apis, and I would expect not all vendors would care to be included in that list.

msporny:
> These changes look good, this does not make oauth2 mandatory, but we are headed towards being able to allow vendors to show support for a secured set of http apis, and I would expect not all vendors would care to be included in that list.

Yep... that's what I'm concerned about is creating some sort of expectation that everyone in the cohort is going to have the cycles to implement this between now and the plugfest.

peacekeeper (commented):
Looks good to me too, but in the future maybe give people a bit more time to review before merging...

Issue #74, opened 2020-11-25
Checks in verify options


clehner:
In the test suite (added in #65), linked data proof options contains a property `"checks": ["proof"]`.This is also used in verification results. Should `checks` be added to `LinkedDataProofOptions` in the spec? Related: #62

OR13:
yes, imo this data model is terrible... here the caller should be asking for a specific set of checks to be run, including:

- predating
- postdating
- revocation
- expiration
- signature / proof
- schema validation

OR13:
Proposed JSON response for various checks:

```
{
      "valid": true,
      "validations": [
        {
          "valid": true,
          "title": "Signature",
          "description": "did:key:z6MkfMDGAbnhGxB9DCHeuhtFjCpfcDpfCjXTZWuvunCXec4P#z6MkfMDGAbnhGxB9DCHeuhtFjCpfcDpfCjXTZWuvunCXec4P"
        },
        {
          "valid": true,
          "title": "Activation",
          "description": "This credential activated 9 months ago"
        },
        {
          "valid": true,
          "title": "Expiration",
          "description": "This credential expires in 8 years"
        }
      ]
    }

```

peacekeeper:
+1 to supporting details about each individual validation/check. I think each entry in the list should also optionally have its own `error` (or similar) field to indicate why it failed.

David-Chadwick:
Putting myself in the position of a service provider, if I ask the Verifier to verify the VP I have just received, I would want full verification to take place by default, including the embedded VCs. I would expect any verifier that ignores the validity dates to be severely deficient. I don't want a reply that tells me all the checks that have been passed. I want one Yes that tells me all tests have been passed. Conversely, if the verification fails, then I might like errors that tell me which bits failed, such as VC has expired etc. So it seems  like the API is back to front if it is passing all the positive checks but not passing all the failed ones.

David-Chadwick:
> here the caller should be asking for a specific set of checks to be run, including:

I think this is the wrong approach. Why  would a caller only want partial verification to take place? The default must surely be to check everything

OR13:
@David-Chadwick is your counter proposal to infer checks from the proof type and credential structure?

- credentialSchema - check if present
- credentialStatus - check if present
- proof - check if present

Sound better?

OR13:
here we have this function: https://github.com/transmute-industries/vc.js/blob/master/packages/vc.js/src/vc-ld/validateCredential.spec.ts

[...]

PR #92, opened 2021-01-29
feat: add support for credentialStatus on credential issue

### LINKED ISSUE #90: Add support for Revocation List 2020
We need HTTP interface changes to support this.

We expect to need to see some examples to show how a credential which used this would be "issued" and "verified"

tplooker:
This PR proposes support for issuing a verifiable credential that includes a credential status, whereby one valid type of credential status implementation is [RevocationList2020](https://w3c-ccg.github.io/vc-status-rl-2020/).

The intended behaviour is that if the credentialStatusMethod property is present in the issue options section of the request, then the requestor is asking the issuer to issue a verifiable credential featuring a status using the method specified in the request.

Outstanding questions for review
- Are there options beyond just the credential status that should be communicated in the issuance request? E.g the initial state of the credentials status? Which revocation list (if applicable) the credential status should be issued against?
- How does a requestor know what credential status methods the issuer supports see [here](https://github.com/w3c-ccg/vc-http-api/issues/91) for an elaboration.

It is also of note, that this API currently assumes that the issuer co-ordinates much of the hidden detail required to issue the credential under a particular method, e.g for revocation list 2020 this could involve creating and hosting a new revocation list or allocating the issued credential against a previously created one.

Resolves #90

### DIFF
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -171,7 +171,7 @@ components:
           "verificationMethod": "did:example:123#z6MksHh7qHWvybLg5QTPPdG2DgEjjduBDArV9EF9mRiRzMBN",
           "proofPurpose": "assertionMethod",
           "created": "2020-04-02T18:48:36Z",
-          "domain": "example.com",
+          "domain": "revocation.example",
           "challenge": "d436f0c8-fbd9-4e48-bbb2-55fc5d0920a8",
           "credentialStatusMethod": "RevocationList2020Status",
         }
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -162,7 +162,7 @@ components:
           description: A challenge provided by the requesting party of the proof.
         domain:
           type: string
-          description: The domain of the proof.
+          description: The intended domain of validity for the proof. For example: website.example
         credentialStatusMethod:
           type: string
           description: The method of credential status to issue the credential including. If omitted credential status will be included.
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -156,7 +156,7 @@ components:
           description: The purpose of the proof. Default: "assertionMethod".
         created:
           type: string
-          description: The date of the proof. If omitted system time will be used.
+          description: The date and time of the proof (with a maximum accuracy in seconds). Default: current system time.
         challenge:
           type: string
           description: The challenge of the proof.
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -153,7 +153,7 @@ components:
           description: The URI of the verificationMethod used for the proof. Default assertionMethod URI.
         proofPurpose:
           type: string
-          description: The purpose of the proof. If omitted "assertionMethod" will be used.
+          description: The purpose of the proof. Default: "assertionMethod".
         created:
           type: string
           description: The date of the proof. If omitted system time will be used.
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -153,19 +153,23 @@ components:
           description: The URI of the verificationMethod used for the proof. Default assertionMethod URI.
         proofPurpose:
           type: string
-          description: The purpose of the proof. Default: "assertionMethod".
+          description: The purpose of the proof. Default 'assertionMethod'.
         created:
           type: string
-          description: The date and time of the proof (with a maximum accuracy in seconds). Default: current system time.
+          description: The date and time of the proof (with a maximum accuracy in seconds). Default current system time.
         challenge:
           type: string
-          description: A challenge provided by the requesting party of the proof. For example: 6e62f66e-67de-11eb-b490-ef3eeefa55f2
+          description: A challenge provided by the requesting party of the proof. For example 6e62f66e-67de-11eb-b490-ef3eeefa55f2
         domain:
           type: string
-          description: The intended domain of validity for the proof. For example: website.example
-        credentialStatusMethod:
-          type: string
+          description: The intended domain of validity for the proof. For example website.example
+        credentialStatus:
+          type: object
           description: The method of credential status to issue the credential including. If omitted credential status will be included.
+          properties:
+            type:
+              type: string
+              description: The type of credential status to issue the credential with
       example:
         {
           "verificationMethod": "did:example:123#z6MksHh7qHWvybLg5QTPPdG2DgEjjduBDArV9EF9mRiRzMBN",
@@ -173,7 +177,9 @@ components:
           "created": "2020-04-02T18:48:36Z",
           "domain": "revocation.example",
           "challenge": "d436f0c8-fbd9-4e48-bbb2-55fc5d0920a8",
-          "credentialStatusMethod": "RevocationList2020Status",
+          "credentialStatus": {
+            "type": "RevocationList2020Status"
+          }
         }
     PresentCredentialOptions:
       type: object
@@ -182,19 +188,19 @@ components:
       properties:
         verificationMethod:
           type: string
-          description: The URI of the verificationMethod used for the proof. If omitted a default assertionMethod will be used.
+          description: The URI of the verificationMethod used for the proof. Default assertionMethod URI.
         proofPurpose:
           type: string
-          description: The purpose of the proof. If omitted "assertionMethod" will be used.
+          description: The purpose of the proof. Default 'assertionMethod'.
         created:
           type: string
-          description: The date of the proof. If omitted system time will be used.
+          description: The date and time of the proof (with a maximum accuracy in seconds). Default current system time.
         challenge:
           type: string
-          description: The challenge of the proof.
+          description: A challenge provided by the requesting party of the proof. For example 6e62f66e-67de-11eb-b490-ef3eeefa55f2
         domain:
           type: string
-          description: The domain of the proof.
+          description: The intended domain of validity for the proof. For example website.example
       example:
         {
           "verificationMethod": "did:example:123#z6MksHh7qHWvybLg5QTPPdG2DgEjjduBDArV9EF9mRiRzMBN",
@@ -210,19 +216,19 @@ components:
       properties:
         verificationMethod:
           type: string
-          description: The URI of the verificationMethod used for the proof. If omitted a default assertionMethod will be used.
+          description: The URI of the verificationMethod used for the proof. Default assertionMethod URI.
         proofPurpose:
           type: string
-          description: The purpose of the proof. If omitted "assertionMethod" will be used.
+          description: The purpose of the proof. Default 'assertionMethod'.
         created:
           type: string
-          description: The date of the proof. If omitted system time will be used.
+          description: The date and time of the proof (with a maximum accuracy in seconds). Default current system time.
         challenge:
           type: string
-          description: The challenge of the proof.
+          description: A challenge provided by the requesting party of the proof. For example 6e62f66e-67de-11eb-b490-ef3eeefa55f2
         domain:
           type: string
-          description: The domain of the proof.
+          description: The intended domain of validity for the proof. For example website.example
       example:
         {
           "verificationMethod": "did:example:123#z6MksHh7qHWvybLg5QTPPdG2DgEjjduBDArV9EF9mRiRzMBN",
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -159,7 +159,7 @@ components:
           description: The date and time of the proof (with a maximum accuracy in seconds). Default: current system time.
         challenge:
           type: string
-          description: A challenge provided by the requesting party of the proof.
+          description: A challenge provided by the requesting party of the proof. For example: 6e62f66e-67de-11eb-b490-ef3eeefa55f2
         domain:
           type: string
           description: The intended domain of validity for the proof. For example: website.example
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -143,7 +143,67 @@ components:
           "proofPurpose": "assertionMethod",
           "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..YtqjEYnFENT7fNW-COD0HAACxeuQxPKAmp4nIl8jYAu__6IH2FpSxv81w-l5PvE1og50tS9tH8WyXMlXyo45CA",
         }
-    LinkedDataProofOptions:
+    IssueCredentialOptions:
+      type: object
+      additionalProperties: false
+      description: Options for specifying how the LinkedDataProof is created.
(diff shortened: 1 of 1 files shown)


docs/vc-http-api.yml
```diff
@@ -143,7 +143,67 @@ components:
...
+      example:
+        {
+          "verificationMethod": "did:example:123#z6MksHh7qHWvybLg5QTPPdG2DgEjjduBDArV9EF9mRiRzMBN",
+          "proofPurpose": "assertionMethod",
+          "created": "2020-04-02T18:48:36Z",
+          "domain": "example.com",
+          "challenge": "d436f0c8-fbd9-4e48-bbb2-55fc5d0920a8",
+          "credentialStatusMethod": "RevocationList2020Status",
+        }
+    PresentCredentialOptions:
```
msporny, line 178: Hmm, PR says it's just adding support for credential status, but many more options seem to be being added for issuance and presentation.

docs/vc-http-api.yml
```diff
@@ -143,7 +143,67 @@ components:
...
+        domain:
+          type: string
+          description: The domain of the proof.
+      example:
+        {
+          "verificationMethod": "did:example:123#z6MksHh7qHWvybLg5QTPPdG2DgEjjduBDArV9EF9mRiRzMBN",
+          "proofPurpose": "assertionMethod",
+          "created": "2020-04-02T18:48:36Z",
+          "domain": "example.com",
+          "challenge": "d436f0c8-fbd9-4e48-bbb2-55fc5d0920a8",
```
msporny, line 204: Same comments as above apply to this block.

msporny (changes requested):
PR seems like it's doing more than the PR title suggests? Provided feedback.

msporny:
> Outstanding questions for review
> 
> * Are there options beyond just the credential status that should be communicated in the issuance request? E.g the initial state of the credentials status? Which revocation list (if applicable) the credential status should be issued against?

Probably not the initial state (I mean, why issue a VC that's immediately revoked?).

Which revocation list might be useful, but then that may be a leaky abstraction -- the calling system shouldn't really know/care about how the issuer is managing that stuff -- it couples the two systems together unnecessarily.

> * How does a requestor know what credential status methods the issuer supports see [here](https://github.com/w3c-ccg/vc-http-api/issues/91) for an elaboration.

We may want to expose a configuration endpoint... but really, the software documentation should probably state this. I'd err on not having a config endpoint right now.

> It is also of note, that this API currently assumes that the issuer co-ordinates much of the hidden detail required to issue the credential under a particular method, e.g for revocation list 2020 this could involve creating and hosting a new revocation list or allocating the issued credential against a previously created one.

Yes, we should keep these details internal to the issuing software for now. Everyone is probably going to have their own way of doing this and it doesn't seem like something we need to interoperate on.

tplooker:
Thanks @msporny, for clarification the API swagger definition was previously sharing one definition for the "options" across both the issue and verify apis so I broke these up so they can have independent options as I expect them to evolve differently.

docs/vc-http-api.yml
```diff
@@ -143,7 +143,67 @@ components:
...
+          type: string
+          description: The method of credential status to issue the credential including. If omitted credential status will be included.
+      example:
+        {
+          "verificationMethod": "did:example:123#z6MksHh7qHWvybLg5QTPPdG2DgEjjduBDArV9EF9mRiRzMBN",
+          "proofPurpose": "assertionMethod",
+          "created": "2020-04-02T18:48:36Z",
+          "domain": "example.com",
+          "challenge": "d436f0c8-fbd9-4e48-bbb2-55fc5d0920a8",
+          "credentialStatusMethod": "RevocationList2020Status",
```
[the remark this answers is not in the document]
tplooker: Yeah me neither, got an alternative in mind, `credentialStatusType` or `credentialStatus`

docs/vc-http-api.yml
tplooker: See comment below really just breaking up the old shared DTO definition

tplooker:
> Probably not the initial state (I mean, why issue a VC that's immediately revoked?).

Agree for that simple case, but just thinking ahead, what about a more advanced status where its not simply binary and expressing revocation? If the answer is we dont really care at this point im happy, was just raising it as an outstanding question.

> Which revocation list might be useful, but then that may be a leaky abstraction -- the calling system shouldn't really know/care about how the issuer is managing that stuff -- it couples the two systems together unnecessarily.

Yeah main question I think is if we make `credentialStatus` a map then we have room for this to evolve, whereas if we make it a string then we are locked to that expression and would have to create other props at the options level to express other things related to the credential status?

> We may want to expose a configuration endpoint... but really, the software documentation should probably state this. I'd err on not having a config endpoint right now.

Yeah would not be advocating for it to be mandatory, kind of like openid metadata, however when it is present it allows clients to auto configure which can be handy.

> Yes, we should keep these details internal to the issuing software for now. Everyone is probably going to have their own way of doing this and it doesn't seem like something we need to interoperate on.

+1

OR13:
@msporny you requested changes but I can't determine what exactly you want to see here.

Are you objecting to the idea that the api should manage this list?

Can you refine your change request to make it clearer?

docs/vc-http-api.yml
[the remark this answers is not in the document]
OR13: personally I like `credentialStatusMethod` or `credentialStatus`

docs/vc-http-api.yml
OR13: these are the json schema for options... this probably should have happened in a separate PR, but i am greatful it happened at all.

docs/vc-http-api.yml
```diff
@@ -179,12 +239,18 @@ components:
...
           type: array
           description: Warnings
+          items:
+            type: string
         errors:
           type: array
           description: Errors
+          items:
+            type: string
       example: { "checks": ["proof"], "warnings": [], "errors": [] }
```
OR13, line 254: I continue to hate this

docs/vc-http-api.yml
```diff
@@ -418,7 +494,7 @@ components:
         verifiableCredential:
           $ref: "#/components/schemas/VerifiableCredential"
         options:
-          $ref: "#/components/schemas/LinkedDataProofOptions"
+          $ref: "#/components/schemas/VerifyOptions"
```
OR13, line 497:
```suggestion
          $ref: "#/components/schemas/VerifyCredentialOptions"
```

docs/vc-http-api.yml
```diff
@@ -427,6 +503,6 @@ components:
         verifiablePresentation:
           $ref: "#/components/schemas/VerifiablePresentation"
         options:
-          $ref: "#/components/schemas/LinkedDataProofOptions"
+          $ref: "#/components/schemas/VerifyOptions"
```
OR13, line 506:
```suggestion
          $ref: "#/components/schemas/VerifyPresentationOptions"
```

OR13 (changes requested):
Blocking:
Not related to revocation list, but I would like to see the the option objects uniquely named.


Non blocking comments:

I like the idea of internalizing the list management, a caller should follow these steps:

1. authenticate
2. issue credential with revocation

credential list should be updated automatically.

peacekeeper:
I think this goes into the right direction, but please give people more time to review (i.e. not 15 hours as in the OAuth 2.0 Client Credentials one https://github.com/w3c-ccg/vc-http-api/pull/89). I know you want to make progress super-fast, but this shouldn't be done by giving as few people as possible the chance to review.

docs/vc-http-api.yml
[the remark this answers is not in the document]
peacekeeper: I've always found it strange how the verification methods in the Security Vocabulary are named. Why are they called `authentication` and `assertionMethod` and `keyAgreement`. Shouldn't it be consistent, i.e. either:

- `authentication` and `assertion` and `keyAgreement`, or
- `authenticationMethod` and `assertionMethod` and `keyAgreementMethod`

Anyway, in this particular case, I think my preference would be to just re-use `credentialStatus` instead of inventing something new.

tplooker:
> I think this goes into the right direction, but please give people more time to review (i.e. not 15 hours as in the OAuth 2.0 Client Credentials one #89). I know you want to make progress super-fast, but this shouldn't be done by giving as few people as possible the chance to review.

Noted, Apologies about the haste on the previous PR

docs/vc-http-api.yml
peacekeeper: Related: https://github.com/w3c-ccg/vc-http-api/issues/74. Each check should be able to express its own error. I can maybe try to proposa a separate PR for that.

msporny: Agree... `items` and `data` are the ultimate "I gave up trying to pick a good name". :P

msporny:
> I think this goes into the right direction, but please give people more time to review (i.e. not 15 hours as in the OAuth 2.0 Client Credentials one #89). I know you want to make progress super-fast, but this shouldn't be done by giving as few people as possible the chance to review.

Agree -- even though I'm doing a review, it's really rushed, I'm sure I'm missing things and I'm worried that whatever rushed thing we do here is going to get locked in because it's going to be memorialized in a test suite. It can happen even when everyone's intention is that we don't want that to happen. See the comment above about "assertionMethod" wrt. how this happens in practice: https://github.com/w3c-ccg/vc-http-api/pull/92#discussion_r566846885

Just because I'm doing a review, doesn't mean the direction has been vetted or that I'm thinking deeply. I'm rushing through it because people in the DHS SVIP cohort say they need features to meet their customer needs and I'm taking that at face value and trying to help. The time to have proposed these changes was months ago... not 45 days before an interop test.

docs/vc-http-api.yml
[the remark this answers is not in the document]
msporny: Also... I was talking more about `RevocationList2020Status` than `credentialStatusMethod`. I'm afraid the name of the "Revocation List 2020" is wrong. It should probably be "Binary Status List 2020", and it should define properties for publishing lists that are for "suspension", "revocation", "activation", etc...

We should probably make this a map, so:

```
credentialStatus: {
  type: "RevocationList2020Status"
}
```

That would allow us to provide other options to the specific type in the future if we wanted to do so.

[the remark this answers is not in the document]
peacekeeper: I would be in favor of pushing for consistency, since I haven't shipped this to customers yet :)

docs/vc-http-api.yml
msporny: Thanks, I missed that these were JSON Schema definitions.

docs/vc-http-api.yml
msporny: The comments above still apply... remove "If omitted" language to "Default" language. The descriptions need the same sort of treatment as the suggestions I made previously.

docs/vc-http-api.yml
tplooker: Understood however I was just fixing a syntax error in this section and would prefer not to expand the scope of this PR to addressing these issues, lets fix with #74

tplooker:
> So, while we're doing reviews -- we don't like this work mode at all. We're going to have to figure out a different way to do this next time around... like requiring that the API is frozen 3 months before the interop test

Understood I think we are all under time pressure with this stuff at the moment and I'm not trying to compound things merely provide solutions. This PR is adding an optional feature so that those that want to support this can, those that don't wont be marked as failing the associated tests, instead shown as opted not to implement.

OR13:
@tplooker please address my comments and request a re-review and I will approve.

It looks like @msporny clarified his change request here: https://github.com/w3c-ccg/vc-http-api/pull/92#discussion_r566860481

I would consider him to continue to block until that feedback is addressed.

docs/vc-http-api.yml
```diff
@@ -143,7 +143,67 @@ components:
...
+      properties:
+        verificationMethod:
+          type: string
+          description: The URI of the verificationMethod used for the proof. If omitted a default assertionMethod will be used.
+        proofPurpose:
+          type: string
+          description: The purpose of the proof. If omitted "assertionMethod" will be used.
+        created:
+          type: string
+          description: The date of the proof. If omitted system time will be used.
```
TallTed, line 191: trying to help get @msporny's point across...
```suggestion
        verificationMethod:
          type: string
          description: The URI of the verificationMethod used for the proof. Defaults to ToDo____________ToDo.
        proofPurpose:
          type: string
          description: The purpose of the proof. Defaults to "assertionMethod".
        created:
          type: string
          description: The date of the proof. Defaults to system time.
```

docs/vc-http-api.yml
```diff
@@ -143,7 +143,67 @@ components:
...
+          description: The URI of the verificationMethod used for the proof. Default assertionMethod URI.
+        proofPurpose:
+          type: string
+          description: The purpose of the proof. Default: "assertionMethod".
+        created:
+          type: string
+          description: The date and time of the proof (with a maximum accuracy in seconds). Default: current system time.
+        challenge:
+          type: string
+          description: A challenge provided by the requesting party of the proof.
```
msporny, line 162:
```suggestion
          description: A challenge provided by the requesting party of the proof. For example: 6e62f66e-67de-11eb-b490-ef3eeefa55f2
```

docs/vc-http-api.yml
mprorock: note, if the property can be omitted, please mark as `nullable: true` so that a null can correctly be passed in the object or parameter and still pass complete validation checking

mprorock (commented):
Looks like a few changes to go.  @tplooker I know you have been slammed - any help we can offer here? we are hoping to get this merged to be able to fully leverage credential status sooner rather than later

docs/vc-http-api.yml
TallTed: @mprorock - I think that comment should be posted somewhere with broader readership, than those interested in my change suggestion (which I believe has been obviated by others).

mprorock: good catch - i will likely create an issue on this and a few other specifics related to openapi as best practices

OR13:
This blocking us.

docs/vc-http-api.yml
tplooker: This options object is currently shared between the credential verify and presentation verify endpoint

docs/vc-http-api.yml
tplooker: Thanks @Tallted suggestions have been adopted in subsequent commits

docs/vc-http-api.yml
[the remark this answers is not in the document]
tplooker: Updated to credentialStatus as an object rather than string type

msporny (approved):
There are still some minor punctuation issues, but ignoring those so we can get this PR in.

OR13 (approved):

OR13:
13 days open 2 approvals.

Issue #88, opened 2021-01-21
Need a mechanism to verify multiple certificates attached to an agricultural shipment 


mprorock:
When processing a shipment of agricultural goods, multiple VCs (sharing a common parent type) need to be verified with a single request for each item type and packaging material (e.g. there will be multiple certificates attached to one shipping container of blueberries).  How is this best approached with the api?

mprorock:
@OR13 any thoughts on how you have approached multiple verifications in the past?

OR13:
@mprorock we talked about this in the previous phase. 

Its possible to make a verifiable presentation which contains no proof, and multiple credentials... (its basically a stanard bucket of credentials with no replay attack prevention).

in order to keep the api simple I would recommend that approach.... however if folks want the `verifyPresentation` api operation to REQUIRE a proof.... this won't work and we would need another endpoint like `verifyCredential` that could do multiple at once.

OR13:
It looks like todays APIs require `proof`.... https://w3c-ccg.github.io/vc-http-api/#/Verifier/verifyPresentation

so the options are:

1. overload verifyPresentation to allow verification of presentations that have no proof.
2. create `verifyCredentials` which can check multiple credentials at once.

@dlongley any preference here?

bumblefudge:
The supply-chain sub-cohort discussion group discussed this today. All of us support an extension to the API in time for the next plugathon. We had no strong objections to 1 OR 2, but have a slight preference for 2 since it minimizes the chance of breaking changes for any existing implementations.

dlongley:
It seems like providing an option to `verifyPresentation` that would not reject presentations lacking a proof is simpler than creating an entirely separate API that does the same thing. My vote would be for #1.

OR13:
This was solved when we enabled verifying presentations without proofs

OR13:
This issue should be closed.

msporny:
We discussed this on the 2021-11-16. @OR13 reported in that a test has been added in PR #113 to address this issue. No further action needed. Closing.

PR #95, opened 2021-02-03
feature/verify many

mprorock:
a way of handling #88 

it became clear to me when working this up and testing some things that there is an additional issue in the way that resources / controllers are named as there is a clearer way to structure the end points in compliance with [the restful api guidelines](https://restfulapi.net/resource-naming/).  i will open a separate issue on this for discussion.  

as is this PR does not break any existing implementations against `credentials/verify` and simply adds the ability to verifyMany by passing in an array of objects, and returning back an array of results that matches the array that was passed into the new method


docs/vc-http-api.yml
```diff
@@ -89,6 +89,30 @@ paths:
           description: invalid input!
         "500":
           description: error!
+  /credentials/verifyMany:
```
msporny, line 92: This is an anti-pattern that raises a number of architectural concerns:

1. Why do you need this, why can't you scale horizontally? We're able to handle 1,000s of verifications per second w/o this API... what scaling issues are you having?
2. This is a DoS attack vector that requires implementations to put DoS mitigations in their application code, rather than protecting themselves at the edges of a network.

To be clear -- feel free to define this as an optional API and implement it for your Phase I... but I expect there to be strong push back in a future WG that attempts to standardize this. You will note that Twitter, Facebook, Github, Pinterest, etc. don't provide batch APIs for their most heavily utilized endpoints -- there is a reason for this.

docs/vc-http-api.yml
```diff
@@ -89,6 +89,30 @@ paths:
...
+      description: Verifies a list of verifiableCredentials and returns a list of index matched verificationResults in the response body.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/VerifyCredentialsRequest"
+        description: Parameters for verifying a list of verifiableCredentials.
+      responses:
+        "200":
+          description: Verifiable Credentials successfully verified!
```
msporny, line 107:
```suggestion
          description: The array of Verifiable Credentials was processed. NOTE: Some verifications might have failed; successful processing does not mean successful verification.
```

msporny: No need to use exclamation marks in API descriptions. None of this should be a surprise or surprising. :)

docs/vc-http-api.yml
```diff
@@ -89,6 +89,30 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/VerifyCredentialsResponse"
+        "400":
+          description: invalid input!
```
msporny, line 113: Needs a more thorough description.

docs/vc-http-api.yml
```diff
@@ -89,6 +89,30 @@ paths:
...
+        "500":
+          description: error!
```
msporny, line 115: When are 500 errors thrown? Under what situations?

msporny (changes requested):
-1 to batch processing APIs. It needs to be demonstrated that the API, as it stands today, is not capable of achieving the desired processing rates.

docs/vc-http-api.yml
mprorock: totally understand, just replicated what was in the existing api for `credentials/verify`.  have updated this.  Also really like adding the additional clarification to the 200 response and have incorporated this

docs/vc-http-api.yml
mprorock: matched existing API as before.  Added a bit of clarity there

docs/vc-http-api.yml
mprorock: typically a system error - noted in f55515f

OR13 (changes requested):
@mprorock can you implement the proposal made here: https://github.com/w3c-ccg/vc-http-api/issues/88#issuecomment-768608983

mprorock:
@OR13 that can be done, the issue is that the swagger UI does not handle `oneOf` very well

see swagger (note the "no parameters":
Screenshot from 2021-02-03 16-02-50

vs redoc (handles properly):
Screenshot from 2021-02-03 16-03-07

This kindof thing can also be done with `nullable: true` on parameters that can be set to explicitly to `null` but then the docs do not clearly mark the parameter as optional or nullable and it is easier for developers to make mistakes.  pr #99 introduces redoc as an option and makes the `oneOf` option doable, which is much more clear to developers coming into a spec than a `nullable` that has a variety of UI presentation handlings

docs/vc-http-api.yml
```diff
@@ -89,6 +89,30 @@ paths:
...
+        description: Parameters for verifying a list of verifiableCredentials.
+      responses:
+        "200":
+          description: The array of Verifiable Credentials was processed. NOTE - Some verifications might have failed; successful processing does not mean successful verification of each credential.  The results array must be inspected for individual results
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/VerifyCredentialsResponse"
+        "400":
+          description: invalid input not matching required parameters, e.g. passing a single object rather than array, or a malformed set of parameters
```
TallTed, line 113:
```suggestion
          description: Invalid input not matching required parameters, e.g., passing a single object rather than array, or a malformed set of parameters
```

OR13:
@mprorock `oneOf` is valid JSON Schema.... if the built in render fails, but redoc succeeds and the JSON Schema is valid.... seems like redoc is a better long term choice... the important part is that the API contract is expressed clearly.... If holder is present... proof must be present.

docs/vc-http-api.yml
mprorock: issue discussed at length in #88 welcome comments on that issue.

OR13: This argument feels stronger for verification than it does for issuance.

factories, tend to make more than 1 thing at a time.

mprorock:
@OR13 removed verifyMany from the PR based on proposal in #88 and also in reference to discussion on #97

docs/vc-http-api.yml
msporny: > factories, tend to make more than 1 thing at a time.

Yes, and they scale via parallelization (more assembly lines), not by making the one machine in the middle of the factory floor bigger... but the analogy is strained... we're talking about computing architecture here, which does things in ways that aren't done on factory floors.

If folks want to insist that we need batch APIs, let's insist that they demonstrate that keeping simple/atomic APIs that can be parallelized don't work. 

Moving discussion to #88.

OR13 (approved):
This change set appears to solve the user story:

"As a developer I want to verify multiple credentials at once"

msporny:
> This change set appears to solve the user story

It does it by adding unnecessary complexity... it hasn't been shown why the current APIs are not capable of addressing the use case.

-1 to this feature, more documentation here: https://github.com/w3c-ccg/vc-http-api/issues/100

msporny (changes requested):
@OR13 wrote:
> This change set appears to solve the user story

It does it by adding unnecessary complexity... it hasn't been shown why the current APIs are not capable of addressing the use case.

-1 to this feature, more documentation here: #100

peacekeeper (changes requested):
I tend to agree with @mprorock that verifying a presentation with multiple credentials isn't really "batch processing", so I think this use case should be supported by the API. But I would comment that:

1. Looking at the current PR, it sounds a bit weird if the endpoint is called "/presentations/verify", but it would now allow a client to pass in something that's considered a "presentation" but not a "verifiable presentation". How do you verify a non-verifiable presentation? (I know what you mean, just saying that this could confuse people; maybe there's a better way of modeling the same thing).
2. Perhaps add a comment which says that implementations may choose to only accept presentations with a single credential (e.g. to prevent Denial-of-Service attacks or other problems mentioned in https://github.com/w3c-ccg/vc-http-api/issues/100).

mprorock:
> 1. Looking at the current PR, it sounds a bit weird if the endpoint is called "/presentations/verify", but it would now allow a client to pass in something that's considered a "presentation" but not a "verifiable presentation". How do you verify a non-verifiable presentation? (I know what you mean, just saying that this could confuse people; maybe there's a better way of modeling the same thing).
yeah, this is one of the issues of multiple layers of "verify"  in this case the api is a controller that acts on "presentations".  in some cases those are presentations with proof, e.g. "verifiable presentation" and in other cases, just a "presentation".

open to suggestions on better language here, perhaps "validate", however the issue then, is that existing implementations of the API would have to be updated.

> 2. Perhaps add a comment which says that implementations may choose to only accept presentations with a single credential (e.g. to prevent Denial-of-Service attacks or other problems mentioned in #100).
this is very good idea.  I will add a comment to the descriptor

msporny:
@peacekeeper wrote:
> I tend to agree with @mprorock that verifying a presentation with multiple credentials isn't really "batch processing",

Well, yeah... totally agree with that -- but that's not where this PR started (see the state of the PR when I first reviewed it). This PR started as sending in an unbounded array of VCs and getting an unbounded set of results. -1 to that.

If what we're talking about here is allowing one presentation containing a number of embedded VCs to be submitted and getting back whether or not all/some failed (as a list of errors, which I think we support) -- then 100% onboard with that. +1 to being able to verify VPs that contain multiple VCs (even if the VP isn't signed). Presentations don't need to be signed... they only need to be signed to be "Verifiable", so we can add an option to note that VP proof check can be skipped.

We do want to state that implementations may impose a size/number restriction on the VP being provided... For example, "If you submit a 1MB VP, or one that contains 100s of VCs, go pound sand -- we're not going to process that, even if you're authorized." ... but that can be up to implementations to draw that line.

docs/vc-http-api.yml
```diff
@@ -110,9 +112,13 @@ paths:
               schema:
                 $ref: "#/components/schemas/VerifyPresentationResponse"
         "400":
-          description: invalid input!
+          description: Invalid or malformed input 
+        "413":
+          description: Payload too large
```
msporny, line 117: Yes! +1 to this -- spec should say that implementations can set payload limits.

docs/vc-http-api.yml
```diff
@@ -93,14 +93,16 @@ paths:
...
-      description: Verifies a verifiablePresentation and returns a verificationResult in the response body.
+      description: Verifies a verifiablePresentation and returns a verificationResult in the response body.  Given the possibility of denial of service, buffer overflow, or other style attacks, an implementation is permitted to rate limit or restrict requests against this API endpoint to those requests that contain only a single credential with a 413 or 429 error code as appropriate.
       requestBody:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/VerifyPresentationRequest"
+              oneOf:
+                - $ref: "#/components/schemas/VerifyPresentationRequest"
+                - $ref: "#/components/schemas/ProoflessVerifyPresentationRequest"
```
msporny, line 105:
```suggestion
                - $ref: "#/components/schemas/ProoflessVerifyPresentationRequest"
```

We shouldn't create a new endpoint... this is just a `.../presentation/verify` endpoint that allows the caller to say "Don't check the proof on the presentation".

docs/vc-http-api.yml
```diff
@@ -295,7 +301,8 @@ components:
           description: The JSON-LD type of the presentation.
         "holder":
           type: object
-          description: The holder
+          description: The holder - will be ignored if no proof is present since there is no proof of authority over the credentials
+          nullable: true
         "verifiableCredential":
           type: array
```
msporny, line 307: hmm, why is this an array? Why not just take in a single VP? -1 to unbounded arrays as inputs and outputs.

docs/vc-http-api.yml
```diff
@@ -295,7 +301,8 @@ components:
           description: The JSON-LD type of the presentation.
         "holder":
           type: object
-          description: The holder
+          description: The holder - will be ignored if no proof is present since there is no proof of authority over the credentials
+          nullable: true
         "verifiableCredential":
```
msporny, line 306: Why is this a 'verifiableCredential' -- shouldn't it be a 'verifiablePresentation' (with the ability to turn off the presentation proof check)?

docs/vc-http-api.yml
```diff
@@ -428,5 +435,10 @@ components:
           $ref: "#/components/schemas/VerifiablePresentation"
         options:
           $ref: "#/components/schemas/LinkedDataProofOptions"
+    ProoflessVerifyPresentationRequest:
+      type: object
+      properties:
+        presentation:
+          $ref: "#/components/schemas/Presentation"
```
msporny, line 442: Let's get rid of the `Proofless` and plain 'ol `Presentation` language -- these are all "VerifiablePresentations" --- https://w3c.github.io/vc-data-model/#presentations-0  -- we were clear in the spec that you don't need to have a proof on a VP for it to be a valid VP... you just can't verify the presentation (but you can still verify the VCs in the presentation.

msporny (changes requested):
I do still have concerns about this API, but am now supportive of it since it's clear it's no longer a Batch HTTP API... we're just trying to verify presentations that can contain multiple VCs. 

We should just upgrade the presentation/verify endpoint to support this use case and I think we're good.

docs/vc-http-api.yml
mprorock: yep, I think this should be helpful ;)

docs/vc-http-api.yml
mprorock: good question, I matched what was already in the code there and did not modify it.  @OR13 any comment as to why that was originally defined as an array? personally I am fine either way as long as it is clear that an implementation may choose to only take a single element

docs/vc-http-api.yml
mprorock: This is not a new endpoint, the endpoint is just `/presentations/verify`, that is the path reference for the openapi spec, the `#` indicates a local reference so that the unique payload can be identified, in this case in the yaml.  We need a distinct id path in the file for openapi to pick it up as an option that can be generated and identified in both code and at the ui side.  an option with other API specs is to use proof as nullable or optional, but the openAPI spec does not clearly show this to the user in a documentation ui when that is the case, so often it is better to distinguish at the request level so that you can clearly show that there are different ways of calling the api, with different payload examples.  

screenshot from above that shows some of the odd behaviors you can get into with nullable/optional params vs clearly defining different payloads.  note in both cases the payload is a `Presentation` object, one with proof, one without.

Screenshot from 2021-02-03 16-02-50

vs redoc (handles properly):
Screenshot from 2021-02-03 16-03-07

docs/vc-http-api.yml
mprorock: that is the array of credential(s) (on line 306) inside the `Presentation` object

docs/vc-http-api.yml
mprorock: so the actual object passed is a `Presentation` object.  Proofless vs Presentation is just how that unique element is identified in the schema description to identify the unique combination of parameters.  I will work up an option here and propose something.  

@msporny My goal here was to implement in such a way that the existing YAML still held with no changes - am I fine to propose a change to the YAML, or could that delay getting this capability in?  

if we can change existing elements, this section can be restructured a bit for clarity, and at that point I would also issue a separate PR to clean up error handling conventions and docs embedded in the spec as well.  Big thing is, that I want to make sure we get all changes in in time for the impending code freeze, in a manner that does not change any existing implementations for vendors.  If we can make minor changes to parameter naming in the docs this whole pr could have been cleaner from the start.

msporny: Happy to fix this later... don't want to block people that need it.

My point was that we're starting to use language that we either specifically decided not to use during the VCWG, or we're inventing new language that might cause people to go to the VC spec to look up the concept only to find nothing there wrt. class being described. We can bike shed the names later as long as all of us understand that it needs to be done and the current names we're using are just placeholders.

docs/vc-http-api.yml
msporny: My bad... that's fine then.

Looking at this stuff feels like peering through a straw at times. 

The W3C PRs have a "Preview" mode where they generate the entire human-readable document and you can read through it w/o having to interpret the HTML in your head. I don't know if we'd be able to do the same here -- since this is taking YAML -> index.html.

mprorock: yeah, this stuff can be painful to mull though - i typically like to run through redoc, swagger, or similar on the cli othewise things get can get confusing.

for example:
```
~/devel/oss/vc-http-api$ redoc-cli serve docs/vc-http-api.yml 
{
  ssr: undefined,
  title: undefined,
  watch: undefined,
  templateFileName: undefined,
  templateOptions: {},
  redocOptions: {}
}

Server started: http://127.0.0.1:8080
```
will validate the yaml and serve it up local to make sure things are sane

docs/vc-http-api.yml
mprorock: agreed entirely to this.  there are a few naming things that would be very helpful to revisit, as well as structural items int he openapi spec that we could address that will make sure things are clean, enterprise friendly, and self documenting.

docs/vc-http-api.yml
mprorock: @msporny happy to change this to a single item, however, the issue then is that it will not be backwards compatible anymore - should we hold this until a solid refactor pass after plugfest?

msporny: I'ts fine as is, no need to change it.

msporny (approved):

mprorock:
@peacekeeper some quick comments regarding your open questions

> 1. Looking at the current PR, it sounds a bit weird if the endpoint is called "/presentations/verify", but it would now allow a client to pass in something that's considered a "presentation" but not a "verifiable presentation". How do you verify a non-verifiable presentation? (I know what you mean, just saying that this could confuse people; maybe there's a better way of modeling the same thing).

Agree that the language defining schema elements in the yaml can be improved here - but not without introducing potential conflict with existing implementations.  This is related to some conversation above with @msporny where there are some better and more clear ways of approaching some of this in the openapi spec, but this would probably be better approached post code freeze on a separate branch, and then merged in a future major version (to allow for possibly non backwards compatible changes).

> 2. Perhaps add a comment which says that implementations may choose to only accept presentations with a single credential (e.g. to prevent Denial-of-Service attacks or other problems mentioned in #100).

agreed, and this has been added around line 98: `Given the possibility of denial of service, buffer overflow, or other style attacks, an implementation is permitted to rate limit or restrict requests against this API endpoint to those requests that contain only a single credential with a 413 or 429 error code as appropriate.`  Is this sufficient?

If both of these are resolved, can we get this approved and merged up so that we and others that require this can begin testing against a fixed set of merged API updates?

mprorock:
ping @peacekeeper - per previous comment - what changes are you looking to see here that will not break current api spec?  as noted we will be revisiting the api more broadly after plugfest to extend and deal with some of these language items that might be confusing to some.

OR13:
@mprorock this PR now has a conflict.

mprorock:
@OR13 resolved

OR13:
Our team is blocked by this, and its consistent with the VC Data Model:

https://w3c.github.io/vc-data-model/#presentations-0

> (proof) If present, the value of the proof property ensures that the presentation is verifiable. For details related to the use of this property, see Section § 4.7 Proofs (Signatures).

By blocking this PR, we are blocking vendors from implementing the standard.

mprorock:
Still blocked by this.  Can we get this in, or are there actual changes requested?

mprorock:
@OR13 @msporny we need this one merged and I do not see any actual changes requested at this point

msporny:
@mprorock wrote:
> @OR13 @msporny we need this one merged and I do not see any actual changes requested at this point

Can you rebase to fix the merge conflicts? Or, if you're ok w/ history being squashed, we can merge right now.

mprorock:
@msporny totally fine either way - if you would prefer a rebase happy to do so.  I don't think at this point we are going to go down the collections of credentials path at a top level, so squashing is probably fine

OR13:
@peacekeeper  can you re-review this?

msporny:
> @peacekeeper can you re-review this?

I'm going to hold for 24 hours for a review by @peacekeeper ... if we don't get one, let's merge.

msporny:
@mprorock wrote:
> @msporny totally fine either way - if you would prefer a rebase happy to do so.

I always prefer to rebase -- it keeps a chronological record of what we tried/changed/did/etc. so we can refer to it days, weeks, or years from now. It also protects all of us from an Intellectual Property perspective as if someone comes in with a patent years from now, we have a record of when a feature was clearly in the public domain (before the patent was filed). You can always do an interactive rebase and squash away the things you don't feel need to be saved. I'd rather the PR authors do this an the Editors. Food for thought.

mprorock:
> I always prefer to rebase

Makes total sense - will push shortly

peacekeeper (approved):
Sorry for the delay, looks good now!

peacekeeper:
Still waiting to rebase, or do you want us to squash and merge?

mprorock:
@peacekeeper i did pull and merge upstream, and then rebase the branch - so in theory all history should be persevered.  please do verify though

peacekeeper:
@mprorock hmm it still says "This branch cannot be rebased due to conflicts"..

mprorock:
@peacekeeper - mind if i close this, branch fresh since so many changes have gone through, and reopen a new pull?

TallTed:
@peacekeeper @mprorock 

There seems to be something weird happening on this PR. 

@peacekeeper is told "This branch cannot be rebased due to conflicts", but I see "This branch has no conflicts with the base branch".

If @msporny doesn't get a conflict warning, I think merging this PR is the best option, because of the volume of changes and complexity of change history (31 commits from several authors).  I'm not sure how best to handle that if the PR needs recreation.

mprorock:
@TallTed I see a no conflicts message as well - but want to make sure that nothing got rolled back.  we have had quite a few merges into master since this pr originally went up

peacekeeper:
@mprorock and @TallTed it only says there are conflicts if I try "Rebase and merge". For "Squash and merge" there are no conflicts.. So I could just do that, but as @msporny says this doesn't really preserve the history of the individual commits (I think this would be equal to your "close and branch fresh" suggestion).

I guess if we can't figure out how to resolve the "rebase and merge" conflicts, we just go ahead and "squash and merge".

mprorock:
@peacekeeper does it show where the rebase conflicts are?

mprorock:
when I run a [compare](https://github.com/w3c-ccg/vc-http-api/compare/master...mesur-io:feature/verifyMany) it shows one file changed, with only correct changes and no conflicts.... @msporny or @OR13 any insights as to why this is showing a conflict for @peacekeeper ?

msporny:
> when I run a [compare](https://github.com/w3c-ccg/vc-http-api/compare/master...mesur-io:feature/verifyMany) it shows one file changed, with only correct changes and no conflicts.... @msporny or @OR13 any insights as to why this is showing a conflict for @peacekeeper ?

I'm seeing the same rebase conflicts that @peacekeeper is seeing... because there are rebase conflicts. :P

The PR is now destroyed beyond repair due to the merge (never do a multi-head merge into a branch you're trying to get a clean rebase on). :) -- git can be unforgiving. I could fix it if I had access to measur-io's repo and/or enough time, but I don't.

If @mprorock doesn't want to do the surgery to save this branch (at this point a hard reset to b77b7d7e71b6a636a442f0766f86c6c9d86865f6 on his branch, followed with a rebase on top of master (with appropriate rebase conflict fixes along the way), followed with an interactive rebase to merge commits to clean up commit history), I suggest we just squash and merge and move on.

mprorock:
Let's just squash it.

msporny:
> Let's just squash it.

By Odin's Beard. Done.

PR #94, opened 2021-01-31
refactor: test suite

tplooker:
This PR starts the refactor of the current plugfest test suite over to a new package that features a more generalized testing infrastructure that is runnable on an HTTP server to allow for on demand testing of register vendor APIs. Still very much a WIP and is based on the hard work of @OR13 on the [did-core-test-suite](https://github.com/w3c/did-test-suite).

Changes of note
- No net change in core test cases, the tests that were being performed before are still the same
- Generalized fixtures, with the plugfest test suite, the fixtures were vendor specific when they should really be shared to ensure there aren't divergent implementations
- New interop test cases added, for instance with the VC verify endpoint a set of tests will run against a filtered list of the fixtures based on things like the did methods the vendor expresses support for in their config.

Whilst converting over to this format I was only able to get results from Digital Bazaars and Transmutes APIs at present, the rest appeared to be offline?

I still have quite a bit to do on this PR, but looking for early feedback primarily from @OR13 due to his familiarity with it.

PR #109, opened 2021-02-11
refactor: further changes to test suite

tplooker:
This PR is a continuation of #94 completing the following tasks

- Refactoring how the test fixtures are loaded, these are now dynamically loaded from the directory.
- Removed two problematic test fixtures that were bad.
- Split out the httpClient related functions from help.js into its own file
- Renamed help.js -> services/utilities.js
- Re-encapsulated the jest based testing infrastructure to be composed of the following
     - A wrapper to run the jest CLI programatically `services/jest/cli.js`
     - A wrapper around the HTML based test reporter using [jest-html-reporters]() `services/jest/testReporter.js`
     - A runner that executes 1 to many test runs `services/jest/testRunner.js` (based on the supplied config) and reports the results out to the following directory structure `test_results/${testRun.name}` (where the name is usually the vendors name) 
- Removed test case 6 from the verifyPresentations API conformance tests as it was incomplete and was already covered by previous test cases
- Removed Transmutes support for did:v1 as a supported didMethod for now as their infrastructure was not passing when attempting to verify these test fixtures? (@OR13 can you confirm?)


OR13 (approved):

tplooker:
Because this is a continuation of #94 and makes no net changes to functionality of the test suite other than removing a duplicate test case, I'm going to merge with the 1 approval.

PR #110, opened 2021-02-11
feat: add update credential status endpoint

tplooker:
This PR adds a credential issuer API for managing an update to the status of an issued credential.

*Outstanding Questions*

- ~~How should the status property relate to the type property in the request? Are we assuming that the type sets the valid enumeration of values for status?~~ Yes
- ~~Is the credentialId the id of the issued credential?~~ Yes

### DIFF
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -58,7 +58,7 @@ paths:
         "200":
           description: Credential status successfully updated!
         "400":
-          description: invalid input!
+          description: Malformed input.
         "404":
           description: credential not found!
         "500":
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -59,6 +59,8 @@ paths:
           description: Credential status successfully updated!
         "400":
           description: invalid input!
+        "404":
+          description: credential not found!
         "500":
           description: error!
   /presentations/prove:
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -56,13 +56,13 @@ paths:
         description: Parameters for updating the status of the issued credential.
       responses:
         "200":
-          description: Credential status successfully updated!
+          description: Credential status successfully updated
         "400":
-          description: Bad Request!
+          description: Bad Request
         "404":
-          description: Credential not found!
+          description: Credential not found
         "500":
-          description: Internal Server Error!
+          description: Internal Server Error
   /presentations/prove:
     post:
       tags:
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -62,7 +62,7 @@ paths:
         "404":
           description: credential not found!
         "500":
-          description: error!
+          description: Internal server error.
   /presentations/prove:
     post:
       tags:
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -158,13 +158,13 @@ components:
               type:
                 type: string
               status:
-                type: number
+                type: string
       example:
         { 
           "credentialId": "urn:uuid:45a44711-e457-4fa8-9b89-69fe0287c86a",
           "credentialStatus": [{
             "type": "RevocationList2020Status",
-            "status": 0
+            "status": "0"
           }]
         }
     LinkedDataProof:
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -41,6 +41,26 @@ paths:
           description: invalid input!
         "500":
           description: error!
+  /credentials/updateStatus:
+    post:
+      tags:
+        - Issuer
+      summary: Updates the status of an issued credential
+      operationId: updateCredentialStatus
+      description: Updates the status of an issued credential.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/UpdateCredentialStatus"
+        description: Parameters for updating the status of the issued credential.
+      responses:
+        "200":
+          description: Credential status successfully updated!
+        "400":
+          description: invalid input!
+        "500":
+          description: error!
   /presentations/prove:
     post:
       tags:
@@ -122,6 +142,24 @@ paths:
 
 components:
   schemas:
+    UpdateCredentialStatus:
+      type: object
+      description: Request for updating the status of an issued credential.
+      oneOf:
+        - type: object
+          properties:
+            credentialId:
+              type: string
+            type:
+              type: string
+            status:
+              type: number
+      example:
+        { 
+          "credentialId": "urn:uuid:45a44711-e457-4fa8-9b89-69fe0287c86a",
+          "type": "RevocationList2020Status",
+          "status": 0
+        }
     LinkedDataProof:
       type: object
       description: A JSON-LD Linked Data proof.
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -41,7 +41,7 @@ paths:
           description: invalid input!
         "500":
           description: error!
-  /credentials/updateStatus:
+  /credentials/status:
     post:
       tags:
         - Issuer
@@ -58,11 +58,11 @@ paths:
         "200":
           description: Credential status successfully updated!
         "400":
-          description: Malformed input.
+          description: Bad Request!
         "404":
-          description: credential not found!
+          description: Credential not found!
         "500":
-          description: Internal server error.
+          description: Internal Server Error!
   /presentations/prove:
     post:
       tags:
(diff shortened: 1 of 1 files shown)


OR13 (approved):

OR13:
> * Is the credentialId the id of the issued credential?

Is the credentialId the id of the issued credential?

Lets do whatever will be most confusing for developers :)

OR13:
> * ....Are we assuming that the type sets the valid enumeration of values for status?

IMO, Yes....

Support for a type is sufficient to convey support for the enumeration of values.... thats the whole point of creating unqiue type names :)

OR13:
@tplooker if you can provide some sample JSON / curls that would be awesome.

tplooker:
@OR13, can you elaborate on what you mean? Is this sufficient? https://github.com/w3c-ccg/vc-http-api/pull/110/files#diff-74dfbc770d6e21cb974f297ba8f1bc7015ece556537a6202e87b78622728a0cfR154

OR13:
@tplooker I guess I was hoping for test suite JSON examples... but maybe better in a separate PR...

tplooker:
Ah gotcha, yes fixtures and tests will be added after this PR.

tplooker:
Would love some feedback from @msporny and @peacekeeper on this one, not trying to rush the process, just aware that this is time critical

mprorock:
we are blocked by this, both functionally, and from a purpose of being able to run a cleanup and consistency pass on the YAML definition file, and on beginning to finalize documentation on a more complete version of the API that matches our bare minimum use case needs.

docs/vc-http-api.yml
```diff
@@ -41,6 +41,28 @@ paths:
...
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/UpdateCredentialStatus"
+        description: Parameters for updating the status of the issued credential.
+      responses:
+        "200":
+          description: Credential status successfully updated!
+        "400":
+          description: invalid input!
```
msporny, line 61:
```suggestion
          description: Malformed input.
```

docs/vc-http-api.yml
```diff
@@ -41,6 +41,28 @@ paths:
...
+        "404":
+          description: credential not found!
+        "500":
+          description: error!
```
msporny, line 65:
```suggestion
          description: Internal server error.
```

docs/vc-http-api.yml
msporny: We really need to describe when these sorts of errors can occur.

docs/vc-http-api.yml
msporny: We really need to describe when these sorts of errors can occur.

docs/vc-http-api.yml
```diff
@@ -116,6 +138,24 @@ paths:
...
+              type: string
+            type:
+              type: string
+            status:
+              type: number
+      example:
+        { 
+          "credentialId": "urn:uuid:45a44711-e457-4fa8-9b89-69fe0287c86a",
+          "type": "RevocationList2020Status",
+          "status": 0
```
msporny, line 157:
```suggestion
          "id": "urn:uuid:45a44711-e457-4fa8-9b89-69fe0287c86a",
          "credentialStatus": [{
              "type": "RevocationList2020Status",
              "statusValue": 0
           }]
```

I'm concerned about both of these proposals for the object.

@dlongley, thoughts?

msporny (changes requested):

docs/vc-http-api.yml
msporny: What error is thrown if the status list type is not supported?

docs/vc-http-api.yml
tplooker: @msporny agree with the functionality you are trying to enable, but how do we unambiguously refer to the different credential statuses in the array you have defined? I.e are we assuming that a credential will never be issued with more than one credential status of the same type?

docs/vc-http-api.yml
tplooker: Agree we need more details on the error cases, will add some now, I would assume it would be a 400 because you are making a request that the server does not understand.

docs/vc-http-api.yml
tplooker: Another approach would be to have this endpoint only update a single status associated to the credential and require the requestor to make multiple requests to update all the credential statuses?

msporny: > are we assuming that a credential will never be issued with more than one credential status of the same type?

I believe so -- unless someone has a use case where they are publishing two revocation lists associated with the same credential.

> Another approach would be to have this endpoint only update a single status associated to the credential and require the requestor to make multiple requests to update all the credential statuses?

I don't think we have to do that... the current form allows you to update all credential statuses associated with the VC in one shot.

tplooker: > I believe so -- unless someone has a use case where they are publishing two revocation lists associated with the same credential.

Ok, I agree at this stage it would be a strange application just pointing out that this mechanism for identifying the associated credential status to update is not 100% explicit and also that it is valid from the VC data model perspective to have multiple statues of the same type in a single credential.

> Unless I'm missing something, I don't think we have to do that... the current form allows you to update all credential statuses associated with the VC in one shot.

Agree im not against this style of bulk/multi status update just calling that out as our other option.

tplooker:
Another outstanding question for this endpoint is should it be `credentials/status` or `credentials/updateStatus`

OR13:
`credentials/status`

docs/vc-http-api.yml
```diff
@@ -41,6 +41,28 @@ paths:
           description: invalid input!
         "500":
           description: error!
+  /credentials/updateStatus:
```
OR13, line 44:
```suggestion
  /credentials/status:
```

OR13 (changes requested):
Thanks for this PR!

docs/vc-http-api.yml
```diff
@@ -41,6 +41,28 @@ paths:
...
+    post:
+      tags:
+        - Issuer
+      summary: Updates the status of an issued credential
+      operationId: updateCredentialStatus
```
OR13, line 49: https://restfulapi.net/resource-naming/

note that the operation has the word "update" but the route should not... this route is an example of a `controller` style endpoint.

docs/vc-http-api.yml
OR13: See https://github.com/w3c-ccg/vc-http-api/pull/110/files#r576914083

tplooker:
Ready for re-review @msporny @OR13 @peacekeeper

tplooker:
Also I have local test cases for revocation almost ready so if we can get this reviewed soon then I will file the next PR

docs/vc-http-api.yml
```diff
@@ -116,6 +138,29 @@ paths:
...
+              type:
+                type: string
+              status:
+                type: number
+      example:
+        { 
+          "credentialId": "urn:uuid:45a44711-e457-4fa8-9b89-69fe0287c86a",
+          "credentialStatus": [{
+            "type": "RevocationList2020Status",
+            "status": 0
```
OR13, line 161: feels like there should be an index in here....

updateCredentialStatus = (credentialId:string, statusIndex:number, statusFlag: boolean)

OR13: @tplooker ^

tplooker: I would have assumed that the `credentialId` would allow you to look that up? Having to supply it is exposing too much complexity to the client IMO

peacekeeper: Requiring `status` to be a number feels very specific to RevocationList2020Status, no? Other credential status mechanisms in the future may have other concepts of status. Maybe allow status to be any string (and in the specific case of RevocationList2020Status the only valid string values are "0" and "1")?

tplooker: +1 @peacekeeper I have updated this

tplooker: @OR13 I have addressed @peacekeeper's query but will leave this open until you have had a chance to reply on your original query

OR13 (approved):

tplooker:
@peacekeeper @msporny looking for re-reviews here PR been open 8 days, made changes yesterday as per @OR13 review and now approval, anything else from you needed before approval and merging?

docs/vc-http-api.yml
```diff
@@ -41,6 +41,28 @@ paths:
...
+        description: Parameters for updating the status of the issued credential.
+      responses:
+        "200":
+          description: Credential status successfully updated!
+        "400":
+          description: Bad Request!
+        "404":
+          description: Credential not found!
+        "500":
+          description: Internal Server Error!
```
msporny, line 65: No exclamation marks in error conditions! :P 

Also, these descriptions are not very descriptive. We need to fix this (not now, but later)

msporny (approved):

docs/vc-http-api.yml
tplooker: But errors are exceptional and hence should be emphasised :) will update

tplooker:
Multiple reviews and approvals 8 days open, merging.

PR #116, opened 2021-02-17
feat: add vaccination certificate test fixture

tplooker:
Adds sample fixtures from the [vaccination vocabulary](https://w3id.org/vaccination)


tplooker:
Just FYI to anyone wanting to add fixtures @OR13 @msporny @mprorock @peacekeeper this is an example PR on how to do it, the test suite will then automatically run these cases against which vendors have support for things like the issuer did method.

OR13 (approved):

tplooker:
Non substantive change to the test suite just adding in another example fixture for interop testing, 1 approval, merging.

PR #128, opened 2021-02-23
feat: add bbs fixtures

tplooker:
PR dependent on #124 being merged ahead.

This PR adds optional test fixtures for the verifyCredential endpoint test cases for those who support BbsBlsSignature2020

PR #129, opened 2021-02-25
feat: signature suite agility

tplooker:
PR dependent on #128 being merged ahead.

Adds in the ability to formally recognise the signature suite that the credential will be issued under for a configured issuer. Also adds in multiple new negative test cases for the verifyCredential interop test cases.

### DIFF
--- a/packages/vc-http-api-test-server/__tests__/issueCredential.spec.js
+++ b/packages/vc-http-api-test-server/__tests__/issueCredential.spec.js
@@ -19,6 +19,7 @@ if (suiteConfig.issueCredentialConfiguration) {
             const res = await httpClient.postJson(value.endpoint, body, {});
             expect(res.status).toBe(201);
             expect(res.body.proof).toBeDefined();
+            expect(res.body.proof.type).toEqual(value.proofType);
         });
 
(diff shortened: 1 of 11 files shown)


peacekeeper (approved):
Hmm I am fine with this, but I wonder..

1. Could you make an argument that it should be possible for the signature suite to be requested by the client as part of [`IssueCredentialOptions`](https://github.com/w3c-ccg/vc-http-api/blob/master/docs/vc-http-api.yml#L197), instead of being hardcoded in the issuer configuration?

2. Or isn't the signature suite that the issuer uses already determined by the `verificationMethod` in the `IssueCredentialOptions`? E.g. if your `verificationMethod` is a `Ed25519VerificationKey2018`, then the proof type must be `Ed25519Signature2018`, no?

So I feel like either 1. or 2. is true, and that explicitly hardcoding a proof type for a combination of issuer endpoint and verification method shouldn't really be necessary, but I guess it can make the test suite more robust?

Or is there some reason I am missing, why this is absolutely required here?

OR13:
@tplooker merge conflicts :(

OR13:
@peacekeeper yes.... it's also interesting to consider if the relationship between the issuer verification method, the requested suites and defaults.

OR13 (approved):

tplooker:
Two approvals, open 4 days, no objections merging.

PR #142, opened 2021-03-06
Change "assertionMethod" option to "verificationMethod" in issueCredential test.

peacekeeper:
"assertionMethod" doesn't exist as an option for `/credentials/issue`.

See https://github.com/w3c-ccg/vc-http-api/blob/master/docs/vc-http-api.yml#L221.

Fixes https://github.com/w3c-ccg/vc-http-api/issues/112.

### DIFF
--- a/packages/vc-http-api-test-server/__tests__/issueCredential.spec.js
+++ b/packages/vc-http-api-test-server/__tests__/issueCredential.spec.js
@@ -69,7 +69,7 @@ if (suiteConfig.issueCredentialConfiguration) {
             expect(res.status).toBe(400);
         });
 
-        it(`6. The Issuer's Issue Credential HTTP API MUST reject if the value of "options.assertionMethod" in the body of the POST request does not exist.`, async () => {
+        it(`6. The Issuer's Issue Credential HTTP API MUST reject if the value of "options.verificationMethod" in the body of the POST request does not exist.`, async () => {
             const body = {
               credential: {
                 ...credentials[0].data,
@@ -77,7 +77,7 @@ if (suiteConfig.issueCredentialConfiguration) {
               },
               options: {
                 ...value.options[0],
-                assertionMethod: 'foo',
+                verificationMethod: 'foo',
               },
             };
             const res = await httpClient.postJson(value.endpoint, body, {});


brianorwhatever:
Thanks for taking a look at this Markus. On a similar not I was noticing that the main issue credential test doesn't pass in the options (https://github.com/w3c-ccg/vc-http-api/blob/23a75e7514b49049cb387dd21331789dfdf07830/packages/vc-http-api-test-server/__tests__/issueCredential.spec.js#L17). This is resulting in there being no way to configure a different proof type which I believe is planning on being explored as the proofType was recently introduced into the config files. The only way there would currently be to issue a different signature scheme would be a separate issuing endpoint which I don't believe is the desired result.

I propose we add the options into the linked test which will allow the API to issue using a desired signature scheme. Please let me know if I am misunderstanding something.

brianorwhatever:
Sorry I found my misunderstanding and that the issuer config is setup in the id option in the config which I was previously configuring with just a did. Will implement in that way.

OR13:
We originally had a requirement that options were "optional"... I have found this frustrating, we should tackle it seperatly.

OR13 (approved):

mprorock (approved):

troyronda (approved):

packages/vc-http-api-test-server/__tests__/issueCredential.spec.js
```diff
@@ -69,15 +69,15 @@ if (suiteConfig.issueCredentialConfiguration) {
...
+        it(`6. The Issuer's Issue Credential HTTP API MUST reject if the value of "options.verificationMethod" in the body of the POST request does not exist.`, async () => {
             const body = {
               credential: {
                 ...credentials[0].data,
                 issuer: value.id
               },
               options: {
                 ...value.options[0],
-                assertionMethod: 'foo',
+                verificationMethod: 'foo',
```
msporny, line 80: I'm a bit concerned here... at present, the only verification relationship that can issue a VC is "assertionMethod". Why are we generalizing here? What's the use case?

msporny (changes requested):
Why are we generalizing here? What is the driving use case?

packages/vc-http-api-test-server/__tests__/issueCredential.spec.js
OR13: Because the test currently does not match the spec: https://github.com/w3c-ccg/vc-http-api/blob/master/docs/vc-http-api.yml#L228

Alternative solution would be to update the spec.

peacekeeper: Yes, this PR is just to fix an inconsistency with the current spec.

clehner:
This is a good fix I think since the existing test is potentially confusing. I just want to point out that the existing version of the test could be considered testing that unknown options are rejected, corresponding to the schema's use of `additionalProperties: false`. I don't know how necessary or useful that is, but if it was intended to test the behavior of `additionalProperties: false` then maybe that could be addressed again separately.

peacekeeper:
+1 to @clehner, that was my conclusion as well. I'm pretty sure the intention of the test to check for a non-existent verification method, not to check if "additionalProperties" is supported.

This PR fixes this.

peacekeeper:
(And on a separate note, I don't think unknown properties should be rejected, so we might want to remove the `additionalProperties: false` statement, but that's another issue).

OR13 (changes requested):
While I agree that this change should go in, we should hold off until after plugfest.

peacekeeper:
> we should hold off until after plugfest.

We're still merging other PRs, so why not this one :)

packages/vc-http-api-test-server/__tests__/issueCredential.spec.js
msporny: Right, sorry -- looks like the test is wrong and needs to be fixed. Agree with @OR13 that it's a bit late to do that.

I'm confused as to why DB is passing this test as-is... we should be failing it, no? /cc @gannan08?

peacekeeper:
> this is exactly why we run plugfests

+1 yes I'm planning to talk about this a bit during the plugfest...

OR13 (approved):

msporny (approved):

OR13:
IMO the correct behavior is to be explicit and NOT overload options.... so callers should provide options like this:

```
{
  "proofPurpose": "assertionMethod",
  "verificationMethod": "did:example:123#key-0"
}
```

The fact that "all VCs are issued from `assertionMethod`s".... will lead to a more confusing API, and it might not be true in the future.

also after issuance, the proof object will always look like:

```
  "proof": {
    "type": "BbsBlsSignature2020",
    "created": "2021-02-23T19:31:12Z",
    "proofValue": "qPrB+1BLs...",
    "proofPurpose": "assertionMethod",
    "verificationMethod": "did:key:zUC7..."
  }
```

I believe this PR should be merged.

OR13:
@msporny please merge when ready.

msporny:
> @msporny please merge when ready.

Done.

PR #144, opened 2021-03-08
add nonce option to DeriveCredential API

troyronda:


docs/vc-http-api.yml
```diff
@@ -282,6 +282,18 @@ components:
...
           "challenge": "d436f0c8-fbd9-4e48-bbb2-55fc5d0920a8",
         }
+    DeriveCredentialOptions:
+      type: object
+      additionalProperties: false
+      description: Options for specifying how the derived credential is created.
+      properties:
+        nonce:
+          type: string
+          description: A nonce provided by the holder of the credential to be included into the LinkedDataProof. For example 6e62f66e-67de-11eb-b490-ef3eeefa55f2
```
troyronda, line 292: If we want to be able to generate a nonce, we might need a bit more nuance in this description or another option.

tplooker (approved):

OR13 (approved):

docs/vc-http-api.yml
troyronda: To make the default less ambiguous, we can either state that the default is to generate a nonce or that the default is empty nonce.

In the case where the default is a generated nonce, we would also want to specify the behavior of an empty nonce option being provided (use empty nonce; generate nonce; disallow).

Note: the [BBS+ Signatures spec](https://w3c-ccg.github.io/ldp-bbs2020/#nonce) currently states:
> A linked data document featuring a BBS+ proof of knowledge linked data proof MAY contain a nonce attribute.

OR13: another example of harmful MAY... why is this not a MUST and be of min length ... etc... @tplooker can we make this a MUST ?

OR13: https://github.com/w3c-ccg/ldp-bbs2020/pull/43

docs/vc-http-api.yml
```diff
@@ -282,6 +282,18 @@ components:
...
+      example:
+        {
+          "nonce": "d436f0c8-fbd9-4e48-bbb2-55fc5d0920a8",
```
troyronda, line 295: As we need to include the nonce bytes into the proof, perhaps we should specify that the nonce option is base64 encoded. @tplooker @OR13

msporny: As a general rule these days, if you catch your self picking a base-encoding format, you're probably wrong. :)

A few things to keep in mind:

1. Multibase exists, if you want to base-encode something, we should use multibase for the value. 
2. Protocols can limit the type of base-encoding used (e.g. base58btc only). It's the easiest way to not get locked into a base encoding that doesn't work for some subset of the community.
2. The nonce value should never be optional and never be "default empty" in a protocol that is susceptible to replays. I think I saw something to this effect in a PR that flew through my inbox... just noting that it'll be a hard objection from us if folks want to default to empty nonces.

That said, the nonce value being encoded as pure bytes really doesn't matter... don't think we need to base-encode it... you just need enough randomness to prevent replays.

troyronda: @msporny 
> Multibase exists, if you want to base-encode something, we should use multibase for the value.

The context for mentioning base64 (or encoding, in general) comes from the BBS+ Signatures implementations and spec examples using it. As the output uses encoded bytes, I was curious if the input should use the same so they end up being the same.

I agree that multibase is good.

> The nonce value should never be optional and never be "default empty" in a protocol that is susceptible to replays.

Yup. In this case the output is a derivation of the VC (and is, itself a VC), but I would imagine the common intention is to use the derivation once. @OR13 has opened a PR on BBS+ Signatures spec regarding the MAY vs MUST nonce topic (https://github.com/w3c-ccg/ldp-bbs2020/pull/43).

> That said, the nonce value being encoded as pure bytes really doesn't matter

Note: the holder and the verifier both include the pure bytes into their proof computation. The output BBS+ LD Proof nonce property uses encoded bytes.

Example from [BBS+ Signatures spec](https://w3c-ccg.github.io/ldp-bbs2020/):
> "nonce": "6i3dTz5yFfWJ8zgsamuyZa4yAHPm75tUOOXddR6krCvCYk77sbCOuEVcdBCDd/l6tIY="

Of course, the derive service can transform from the input to bytes for the proof computation and do the output encoding. (and, as currently described in this PR, this is what would have to happen).

OR13: @tplooker I agree with this thread, and am bothered by the base64 use in that suite :) we should move to multibase in the next version.

OR13 (approved):

peacekeeper (approved):

OR13:
This is spec only, and limited support for it in the test suite, propose we merge before cut, since its already an optional API.

PR #168, opened 2021-03-14
Add credential exchange APIs for holders

OR13:
- Notify a holder of an available presentation
- Present to a holder (telling them to store vs  the "verifyPresentation" operation which is not stateful, and not a holder API)

### DIFF
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -113,6 +113,65 @@ paths:
           description: invalid input!
         "500":
           description: error!
+  /presentations/available:
+    post:
+      tags:
+        - Holder
+      summary: Notifies a holder of an available presentation.
+      operationId: notifyPresentationAvailable
+      description:
+        This API is the first part of a holder to holder credential exchange over http.
+        The server MAY reply with a query for the client to proceed with a presentation.
+      requestBody:
+        description: Details the client provides to the server to help it decide if the presentation should be made.
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/NotifyPresentationAvailableRequest"
+
+      responses:
+        "200":
+          description: Proceed with presentation
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/NotifyPresentationAvailableResponse"
+        "400":
+          description: Request for presentation is malformed
+        "501":
+          description: Not implemented
+        "500":
+          description: internal error
+  /presentations/submissions:
+    post:
+      tags:
+        - Holder
+      summary: Provide a presentation for a holder to store.
+      operationId: storePresentation
+      description: Present from a client holder to a server holder, for the purpose of verification AND storage.
+      requestBody:
+        description: Details the client provides to the server to help it decide if the presentation should be made.
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/StorePresentationRequest"
+      responses:
+        "202":
+          description: Presentation accepted
+        "400":
+          description: Presentation is malformed
+        "401":
+          description: Presentation did not contain a proof
+        "402":
+          description: Payment required
+        "403":
+          description: Presentation verification failed
+        "425":
+          description: Server is unwilling to risk processing a request that might be replayed
+        "501":
+          description: Not implemented
+        "500":
+          description: internal error
   /credentials/verify:
     post:
       tags:
@@ -592,6 +651,91 @@ components:
           $ref: "#/components/schemas/PresentCredentialOptions"
     ProvePresentationResponse:
       $ref: "#/components/schemas/VerifiablePresentation"
+    NotifyPresentationAvailableRequest:
+      type: object
+      properties:
+        query:
+          type: object
+          description: See https://w3c-ccg.github.io/vp-request-spec/#format
+          properties:
+            type:
+              type: string
+              description: "The type of query the server should reply with."
+            credentialQuery:
+              type: object
+              description: "Details of the client's available presentation"
+
+      example:
+        {
+          "query":
+            [
+              {
+                "type": "RequestQueryByFrame",
+                "credentialQuery":
+                  [
+                    {
+                      "type":
+                        [
+                          "VerifiableCredential",
+                          "CommercialInvoiceCertificate",
+                        ],
+                      "reason": "Wallet XYZ is ready to selectively disclose new credentials.",
+                    },
+                  ],
+              },
+            ],
+        }
+
+    NotifyPresentationAvailableResponse:
+      type: object
+      properties:
+        query:
+          type: object
+          description: See https://w3c-ccg.github.io/vp-request-spec/#format
+        domain:
+          type: string
+          description: See https://w3id.org/security#domain
+        challenge:
+          type: string
+          description: See https://w3id.org/security#challenge
+      example:
+        {
+          "query":
+            [
+              {
+                "type": "QueryByFrame",
+                "credentialQuery":
+                  {
+                    "frame":
+                      {
+                        "@context":
+                          [
+                            "https://www.w3.org/2018/credentials/v1",
+                            "https://w3id.org/traceability/v1",
+                            "https://w3id.org/bbs/v1",
+                          ],
+                        "type":
+                          [
+                            "VerifiableCredential",
+                            "CommercialInvoiceCertificate",
+                          ],
+                        "credentialSubject":
+                          {
+                            "@explicit": true,
+                            "type": ["CommercialInvoice"],
+                            "purchaseDate": {},
+                            "destinationCountry": {},
+                          },
+                      },
+                  },
+              },
+            ],
+          "domain": "jobs.example.com",
+          "challenge": "3182bdea-63d9-11ea-b6de-3b7c1404d57f",
+        }
+    StorePresentationRequest:
+      $ref: "#/components/schemas/VerifiablePresentation"
+
     VerifyCredentialRequest:
       type: object
       properties:
--- a/docs/vc-http-api.yml
+++ b/docs/vc-http-api.yml
@@ -121,7 +121,7 @@ paths:
       operationId: notifyPresentationAvailable
       description:
         This API is the first part of a holder to holder credential exchange over http.
-        The server MAY reply with a query for the client to proceed with a presentation.
+        Notifies this server-holder that the client-holder (caller) is prepared to deliver a presentation compliant with the provided Verifiable Presentation Request query. This is the first part of a presentation exchange over this HTTP API. This server-holder MUST return the required 'domain' and 'challenge' parameters for the Presentation. This server-holder MAY reply with an alternate Verifiable Presentation Request query if it requires something different than the one proposed by the client-holder. The client-holder MUST return a Presentation in accordance with the response parameters.
       requestBody:
         description: Details the client provides to the server to help it decide if the presentation should be made.
         content:
@@ -148,7 +148,7 @@ paths:
         - Holder
       summary: Provide a presentation for a holder to store.
       operationId: storePresentation
-      description: Present from a client holder to a server holder, for the purpose of verification AND storage.
+      description: "Delivery endpoint for a client-holder to provide a presentation in compliance with the Verifiable Presentation Request provided in the Notification call. This server-holder MUST match the 'domain' and 'challenge' to a 'domain' and 'challenge' previously sent in a Notification Response, and not already received. The server-holder MUST verify the Presentation. The server MUST store the Presentation if verification passes."
       requestBody:
         description: Details the client provides to the server to help it decide if the presentation should be made.
         content:


kimdhamilton:
General comment: those of us not in the SVIP program should also be privy to reasons behind PRs. What are some use cases for this?

OR13:
@kimdhamilton initially this API was only about internal "issuer" and "verifier" for the vc data model, which made working with verifiable presentations hard. We added the presentation endpoints to help with that, and the derive endpoint for bbs+.

These endpoints offer a similar flow to CHAPI, but over http.

We need these APIs as a way to exchange credentials between HTTP APIs in an enterprise context.

OR13:
@msporny @mprorock @mkhraisha @bumblefudge @mavarley any comments on this proposed API for doing "CHAPI" over HTTP?

msporny:
> @msporny @mprorock @mkhraisha @bumblefudge @mavarley any comments on this proposed API for doing "CHAPI" over HTTP?

Yes -- there's not enough here to understand what you're going for. Or rather, what you're doing looks inverted to what I was thinking we'd do... but I don't know if I'm misreading the API or not. Having a call and presentation on the proposal would be helpful. My expectation was:

1. As a holder, you can initiate a certain known credential flow from the requesting party (issuer or verifier).

... but what you seem to be proposing is:

2. As an issuer, you can initiate a "hey, I've got a presentation for you" flow.

So, I'd like to meet and discuss... ideally with a presentation so that we can all understand the use cases and flows possible with this API.

mprorock:
Like the general direction of this, but think there is more to fill out.  Agree with @msporny on setting up a discussion around this.  There are some great lessons learned from chapi that are worth considering.   Perhaps a topic for an upcoming CCG call?

mprorock:
Adding a bit more commentary, as an issuer (or anywhere else in the chain), there are often cases where we need to notify another party to come get a VP or other set of credentials, and this proposal handles that case pretty well.

aniltj:
> So, I'd like to meet and discuss... ideally with a presentation so that we can all understand the use cases and flows possible with this API.

I would recommend having this conversation w/in the context of the CCG rather than an SVIP cohort only call. This is something that I think has broad enterprise applicability and as such could benefit from the community discussion.

msporny:
> I would recommend having this conversation w/in the context of the CCG rather than an SVIP cohort only call. This is something that I think has broad enterprise applicability and as such could benefit from the community discussion.

Yes, agreed.

bumblefudge:
+1 for CCG as venue, if they've got space on the calendar.  Maybe one quick cohort call just to plan for a more public call, i.e. brainstorm enough that a few opinionated parties could make a slide or two for some straw men?

mavarley:
+1 to a sequence diagram and use case description (even a concrete example use case is fine to get rolling). +1 to CCG discussion.

OR13:
Yes, pictures will help.

For the perspective of VC Data Model roles, I think these APIs are only about Holders, NOT Issuers or Verifiers.

The intention is to allow Holder A to present credentials to Holder B for the purpose of Holder B retaining the credentials.

The fact that Holder B probably verifies them is irrelevant, and confusing given the we already have "verifier" apis.

If it helps to think of Holder A as a web hook client, and Holder B as a web hook server, I think thats compatible with the proposed endpoints.

OR13:
Can I get actual code review on this PR?

Are folks saying they will only review / approve or request changes after we "have a call"? 

IMO, we should merge this as is, and iterate on it.

vsnt:
Do y'all want to discuss/kick off discussion at the CCG call tomorrow/March 24?

mprorock (approved):
This gives us a good starting place to improve on

msporny:
> Are folks saying they will only review / approve or request changes after we "have a call"?

Yes. The design of the API is currently unclear.

To put it more bluntly: I object to merging a PR where it's not clear which use cases it's attempting to address. :)

docs/vc-http-api.yml
```diff
@@ -113,6 +113,65 @@ paths:
           description: invalid input!
         "500":
           description: error!
+  /presentations/available:
+    post:
+      tags:
+        - Holder
+      summary: Notifies a holder of an available presentation.
```
Therecanbeonlyone1969, line 120: Can we replace "holder" with "potential holder" since the notification can be ignored and language matters. @OR13

TallTed:
I'm a bit concerned about referring to API calls as specific to Holders or Issuers or Verifiers, as if there are no calls that more than one of these would need, and as if an entity can only be one of these at any given time.  To my mind, any entity should be able to make any API call that's relevant to their usage scenario, which may not have been included in our UCR because we were unaware of it (or it didn't exist until after the UCR was published), and an entity might be any combination of one or more of Holder, Issuer, and Verifier, and therefore need to use any or all API calls (which should not be broken up into CommonApi, HolderApi, IssuerApi, VerifierApi, etc.)...

docs/vc-http-api.yml
```diff
@@ -113,6 +113,65 @@ paths:
...
+        "400":
+          description: Presentation is malformed
+        "401":
+          description: Presentation did not contain a proof
+        "402":
+          description: Payment required
+        "403":
+          description: Presentation verification failed
+        "425":
+          description: Server is unwilling to risk processing a request that might be replayed
```
Therecanbeonlyone1969, line 170: @OR13 does this mean there is no replay protection? If yes, then "Presentation is not replay protected" might be better. If not, why use this error if the only other reason for rejection would be that the holder is not trusted? Or can the submission come from a different endpoint/DID than the FYI notification to the potential Holder?

Therecanbeonlyone1969 (approved):

Therecanbeonlyone1969:
A couple of things to note:

1. This is the initiation of a "Pull" flow. A CURRENT Holder (Alice) notifies a POTENTIAL FUTURE Holder (Bob) that there is a VP that is relevant to them. Bob can choose to ignore this FYI.
2. Bob then CHOOSES to respond and in such a way that there is replay protection, letting Alice know where to drop the VP
3. Alice then submits the VP to BOB who verifies it AND stores it 

The simplest example in the eCommerce world is an Order from Alice to Bob of a sensitive product Alice bought from Charles to resale e.g. high-performance chipset. Bob will only accept the order if

- Alice can prove that she has been KYCd/AMLd e.g. has a bond by a third party e.g. a Bank (1st VC)
- Charles has undergone  KYC/AML by Alice (2nd VC) and is provably not from a bad actor nation as certified by another trusted party (3rd VC linked in the 2nd VC)
- the product has been certified by Alice to be compliant with all of Bob's requirements (4th VC)
- Order meets the commercial requirements as laid out in the MSA between Alice and Bob (5th VC which links all 4 other VCs together)

So this is a super common use case that currently is not possible with the current version of this API and this PR addresses it. We can tweak it if we can clearly define, where this PR is NOT meeting all currently identified needs.

mkhraisha (approved):

mkhraisha:
I have some minor quibles with the error codes (401 for malformed should be unauth) but this is absolutely not a blocker

TallTed:
@Therecanbeonlyone1969 said --
> The simplest example in the eCommerce world is an Order from Alice to Bob of a sensitive product Alice bought from Charles to resale e.g. high-performance chipset. Bob will only accept the order if

Something's not right here... and clarity is paramount in discussions like these.

* Is Alice (re)selling to Bob? If so, the Order is from Bob to Alice; the Shipment would be from Alice to Bob.

* `e.g.` should always be preceded by some punctuation (most commonly a comma; sometimes a semicolon, an open-parenthesis, or something else), and followed by a comma. so `resale e.g.` should become `resale, e.g.,`

* `to resale,` should be `to resell,` or `for resale,`

(And all of this should be transformed into a Use Case and added to that document, if it's not there already -- and if it *is* there, we should be working against and/or clarifying the Requirements that stem from that Use Case.)

OR13:
Some other related issues:

- https://github.com/w3c-ccg/universal-wallet-interop-spec/issues/75
- https://github.com/w3c-ccg/universal-wallet-interop-spec/issues/76

^ note that these may not be relevant depending on the "holder's wallet implementation".

msporny:
> So this is a super common use case that currently is not possible with the current version of this API and this PR addresses it.

The PR might be layering the wrong things in the wrong order, and folks should probably take a step back and see that this sort of API can be designed in a few different ways before settling on this design. 

For example, one alternative is to use a ping-back pattern and then use the "pull-based mechanism" I outlined above. If we were to do that, we'd have a cohesive design for both the pull-based AND the push-based approaches, instead of just designing for one w/o considering the other. One of the concerns I have here is a proliferation of a large number of endpoints that are specific to certain use cases rather than trying to minimize the number of endpoints we support. If we design for both pull and push at the same time, we increase our chances of reusing endpoints and thus reducing endpoint proliferation and complexity.

I also get it that the Supply Chain folks have a valid use case... but you can't just design these APIs around only that use case.

Therecanbeonlyone1969:
> @Therecanbeonlyone1969 said --
> 
> > The simplest example in the eCommerce world is an Order from Alice to Bob of a sensitive product Alice bought from Charles to resale e.g. high-performance chipset. Bob will only accept the order if
> 
> Something's not right here... and clarity is paramount in discussions like these.
> 
> * Is Alice (re)selling to Bob? If so, the Order is from Bob to Alice; the Shipment would be from Alice to Bob.
> * `e.g.` should always be preceded by some punctuation (most commonly a comma; sometimes a semicolon, an open-parenthesis, or something else), and followed by a comma. so `resale e.g.` should become `resale, e.g.,`
> * `to resale,` should be `to resell,` or `for resale,`
> 
> (And all of this should be transformed into a Use Case and added to that document, if it's not there already -- and if it _is_ there, we should be working against and/or clarifying the Requirements that stem from that Use Case.)

@TallTed ... sorry meant quote but wrote order ... changed

TallTed:
@Therecanbeonlyone1969 - Please note that my earlier comment had three bullet points (and a related parenthetical observation). You addressed the first bullet point. Please take another look at the rest.

docs/vc-http-api.yml
OR13: I just included this error code since it seemed like it might bee relevant to certain veracity requirements... and because I knew it would spark a debate :)

even if the holder is authenticated, the holder might be presenting credentials which are "too young / unstable"... think finality requirements... its a thought I could live without it.

docs/vc-http-api.yml
OR13: can you make the proposed change on the PR and I will accept?

OR13:
@msporny the use case is Alice wants to send Bob a verifiable presentation for storage (not verification).

You may have another use case in mind, but that shouldn't block the most basic of "Alice send Bob a VP"... are asserting that we need something more complicated than this to start? 

For context this flow here: https://w3c-ccg.github.io/vp-request-spec/#did-authentication-request

is essentially Identical to this proposal.

msporny:
> @msporny the use case is Alice wants to send Bob a verifiable presentation for storage (not verification).

I'm going to assert that the use case above is new, and it's not the following use case:

> For context this flow here: https://w3c-ccg.github.io/vp-request-spec/#did-authentication-request
> 
> is essentially Identical to this proposal.

I think we're miscommunicating, so I'm going to assert that it's not essentially identical and get more pedantic below (in an effort to clear up the miscommunication).

> You may have another use case in mind, but that shouldn't block the most basic of "Alice send Bob a VP"... are asserting that we need something more complicated than this to start?

Let's take a look at a pull-based mechanism:

1. PartyA initiates a credential flow with PartyB (by hitting some HTTP API endpoint).
2. PartyB sends challenge and Verifiable Presentation Request to PartyA.
3. PartyA responds with digitally signed Verifiable Presentation with to PartyB.

Now, let's take a look at what could be a push-based mechanism:

0. PartyB asks PartyA to initiate a credential flow (by hitting some HTTP API endpoint).
1. PartyA initiates a credential flow with PartyB (HTTP API endpoint sent in previous message).
2. PartyB sends challenge and Verifiable Presentation Request to PartyA.
3. PartyA responds with digitally signed Verifiable Presentation with to PartyB.

Note that the only thing that changed with the push-based mechanism was the addition of one more step (a pingback, or notification). The rest stays the same. I just want to make sure this API is aligned with the model above. I can't tell from just the PR if it is.

mavarley:
> Note that the only thing that changed with the push-based mechanism was the addition of one more step (a pingback, or notification). The rest stays the same. I just want to make sure this API is aligned with the model above. I can't tell from just the PR if it is.

I would also like to know more about the "pull based" mechanism's "Step 0." -- it seems like some bespoke or out-of band arrangement. How does PartyA "know" to tell PartyB a VP is ready for delivery? Does PartyB have any pre-established relationship with PartyA (maybe not in large supply chain cases, probably some session in personal identity cases...). What if PartyA wants to provide "more" to PartyB as either updates to previous Presentations or as new content, is there state at PartyB that needs to be maintained? (Yes No Maybe/out of Scope are all valid answers but impact the specification of the API).

docs/vc-http-api.yml
```diff
@@ -113,6 +113,65 @@ paths:
...
           description: error!
+  /presentations/available:
+    post:
+      tags:
+        - Holder
+      summary: Notifies a holder of an available presentation.
+      operationId: notifyPresentationAvailable
+      description:
+        This API is the first part of a holder to holder credential exchange over http.
+        The server MAY reply with a query for the client to proceed with a presentation.
```
mavarley, line 124: Attempt at clearer description:
"Notifies this holder that the caller is prepared to deliver a presentation compliant with the provided Verifiable Presentation Request query. This is the first part of a presentation exchange from over this HTTP API. This holder (server) MUST return the required 'domain' and 'challenge' parameters for the Presentation. This holder (server) MAY reply with an alternate Verifiable Presentation Request query if it requires something different than the one proposed by the caller. The caller MUST return a Presentation in accordance with the response parameters."

docs/vc-http-api.yml
```diff
@@ -113,6 +113,65 @@ paths:
...
+              $ref: "#/components/schemas/NotifyPresentationAvailableRequest"
+
+      responses:
+        "200":
+          description: Proceed with presentation
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/NotifyPresentationAvailableResponse"
+        "400":
```
mavarley, line 139: Do we need a 401 Unauthorized (I don't know who you are, caller) and a 404 (I don't know what this presentation you are trying to send me is, caller)?

docs/vc-http-api.yml
```diff
@@ -113,6 +113,65 @@ paths:
...
+          description: Not implemented
+        "500":
+          description: internal error
+  /presentations/submissions:
+    post:
+      tags:
+        - Holder
+      summary: Provide a presentation for a holder to store.
+      operationId: storePresentation
+      description: Present from a client holder to a server holder, for the purpose of verification AND storage.
```
mavarley, line 151: There are confusing terms here; are they callers/holders/clients/this/server? I am fine will "client holder" and "server holder" - although not sure why the server needs to be a Holder - is it a requirement of this API endpoint that the server _not_ verify the presentation?

Attempt at clearer spec:
"Delivery endpoint for a caller to provide a presentation in compliance with the Verifiable Presentation Request provided in the Notification call. This holder (server) MUST match the 'domain' and 'challenge' to a 'domain' and 'challenge' previously send in a Notification Response, and not already received."

Any other processing rules here? what about the caller-holder's DID?

docs/vc-http-api.yml
```diff
@@ -113,6 +113,65 @@ paths:
...
+          application/json:
+            schema:
+              $ref: "#/components/schemas/StorePresentationRequest"
+      responses:
+        "202":
+          description: Presentation accepted
+        "400":
+          description: Presentation is malformed
+        "401":
+          description: Presentation did not contain a proof
```
mavarley, line 164: What if this presentation does not match a previous Notification?

docs/vc-http-api.yml
```diff
@@ -113,6 +113,65 @@ paths:
...
+        "402":
+          description: Payment required
+        "403":
+          description: Presentation verification failed
```
mavarley, line 168: Is verification by this holder (server) required or optional?

docs/vc-http-api.yml
mavarley: @OR13 can you clarify? Do you mean "may be a replay of a previous credential" or "a request which may be replayed by the caller to update a presentation with a new state?" -- and if you mean the latter, that doesn't seem like a 400 error, but like a "201 Accepted" or "Continue" response, like "I got your message but I'm not donna deal with it now..."

docs/vc-http-api.yml
```diff
@@ -592,6 +651,91 @@ components:
           $ref: "#/components/schemas/PresentCredentialOptions"
     ProvePresentationResponse:
       $ref: "#/components/schemas/VerifiablePresentation"
+    NotifyPresentationAvailableRequest:
+      type: object
+      properties:
+        query:
+          type: object
+          description: See https://w3c-ccg.github.io/vp-request-spec/#format
+          properties:
```
mavarley, line 660: Shouldn't the caller DID be in this message? (or maybe it is somewhere else in the request butt I missed it).
Seems like a good opportunity to DID-Auth to this holder (server) - is that required? desired? anti-pattern?

docs/vc-http-api.yml
```diff
@@ -592,6 +651,91 @@ components:
...
+                    },
+                  ],
+              },
+            ],
+        }
+
+    NotifyPresentationAvailableResponse:
+      type: object
+      properties:
+        query:
```
mavarley, line 692: optional?
What if a client cannot meet the requirements of the new query, is there a need for a follow-up call? How does the server protect against spam notifications?

docs/vc-http-api.yml
```diff
@@ -592,6 +651,91 @@ components:
...
+          type: object
+          description: See https://w3c-ccg.github.io/vp-request-spec/#format
+        domain:
+          type: string
+          description: See https://w3id.org/security#domain
+        challenge:
+          type: string
+          description: See https://w3id.org/security#challenge
+      example:
```
mavarley, line 701: I was thinking there needs to be some sort of 'session' provided here, so there can be a co-relation on the Presentation endpoint; but then there is suddenly (potentially) a lot of state for this holder (server) to manage. @OR13 is there a model you have in mind how this holder (server) would link a presentation to a notification, or is that important for your use case? (It is important for some personal credential use cases).

bumblefudge: forgive me if this is an anti-pattern or just a dumb idea for 13,152 reasons, but might an authN token be enough of a correlation or session to manage? if it's a bad idea or deliberately out of scope, some non-normative guidance might be good to make sure it's "at your own risk" to bind challenges to authN tokens or other kinds of session tracking :D

mavarley: auth token could work, and depending on its structure (bearer token) can alleviate the need for server side session state. But I agree, we need clarity on if this is in scope for this API or not, and then language that states these assumptions.

TallTed:
@mavarley - In several of your notes above, you've used the term **DID**, where I think you should use **URI** or **IRI**. There is no VC/VP/VR requirement that DIDs be used at any point -- only URIs/IRIs -- which do *include* DIDs, CIDs, and identifiers of other kinds, so DIDs *may* be used at each point you suggest, but it's important not to imply such a requirement where there is none.

mavarley:
> @mavarley - In several of your notes above, you've used the term **DID**, where I think you should use **URI** or **IRI**. There is no VC/VP/VR requirement that DIDs be used at any point -- only URIs/IRIs -- which do _include_ DIDs, CIDs, and identifiers of other kinds, so DIDs _may_ be used at each point you suggest, but it's important not to imply such a requirement where there is none.

A couple points:
1. totally 100% understand and agree with your statement that VC/VPs do not require DIDs (thank-you for catching this!) 
2. I did mean to use the tern DIDs (and the properties of DIDs) as an entity identifier when establishing a secure session, because of the identifier->key relationship this affords.

So, either we do specify an API that relies on DIDs as a means of securely identifying parties (using DIDAuth or something else) - which seems like a 10-megaton landmine to avoid - or we can indicate that URIs must be processed for the 'server holder' to authenticate the 'caller-client- holder' (require the parameter and leave that as deliberately vague as to how) or be clear and descope this check from this part of the API.... 

so @TallTed happy to adjust the notes above to be URIs, but the real question I was driving at remains (for everyone): does the server need a URI from the client to establish the session/relationship, or is that out of scope here?

Thanks!

docs/vc-http-api.yml
mavarley: > Shouldn't the caller DID be in this message? 
Oops, I meant presenter's id URI.
If it is included, is there some sort of authentication required associated with this id uri that the spec needs to define?

TallTed:
@mavarley - I think I see what you're aiming at. 

Please note that DIDs are not the only identifiers that can provide an identifier->key relationship. WebIDs often do this as well. I'm sure there are other examples, though none leap to my mind just now. 

Therefore, rather than specifying DIDs just to get one/some of its feature(s), I think it would be best if things were reworded to focus on the required feature(s), which allows for use of other classes of identifiers (such as WebIDs), and so helps future-proof this API.

docs/vc-http-api.yml
OR13: out of scope, the same issue regarding spam exists for all the other endpoints... and there is no authorization defined... which would appear to be a prerequisite for spam mitigation.

I pulled query from https://w3c-ccg.github.io/vp-request-spec/#query-by-example

which is exactly how CHAPI works,

Imo if you return a 400+ as the server to a notification request, each of those cases would seem to be need to be defined.... including https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

docs/vc-http-api.yml
OR13: again these are pulled from the VC Data Model, and would seem to be non negotiable, regardless of additional transport of message level security mechanisms.

OR13:
I don't see any objections to this PR, or changes requested and its been up for 12 days... Acknowledging that git lets us make revisions, and changes to the spec, any objections to merging this as is, and taking any further changes in a subsequent PR?

OR13:
@TallTed I agree with your comment regarding identifiers, but it does not apply to this PR, its applies to the entire spec... 

Is this API meant to work for any IRI, or just DIDs? 

If we want it to work for other identifiers, we should probably have some language and examples of that, or we should be clear that we only want this spec to cover DIDs and VCs... in either case, this is a great discussion for an independent issue.

docs/vc-http-api.yml
OR13: https://w3c-ccg.github.io/vp-request-spec/#example-1-a-sample-query

```
recipients: [
    // an optional key agreement key for encrypting the response if
    // this is supported
  ]
```

docs/vc-http-api.yml
OR13: @mavarley can you either convert this to an issue or a suggested change, that I can approve.

docs/vc-http-api.yml
OR13: Probably we need a lot fo 400 level status codes... prefer to tackle them in a separate PR, please open an issue to discuss further.

docs/vc-http-api.yml
OR13: @mavarley the entire point of this API is that the server will be a holder.

The client is sending data to the server that the server is expected to store.
There are no other APIs that explicitly indicate that the server will store / be a holder.

> although not sure why the server needs to be a Holder - is it a requirement of this API endpoint that the server not verify the presentation?

I would see verification as a strong SHOULD for this API, but its clearly not a must like the verifyPresentation.

Again, please convert this suggestion to something I can approve or an issue.

docs/vc-http-api.yml
OR13: some kind of 400.

docs/vc-http-api.yml
OR13: I'd recommend it be required.

docs/vc-http-api.yml
OR13: https://tools.ietf.org/html/rfc8470#section-5.2

```
  A 425 (Too Early) status code indicates that the server is unwilling
   to risk processing a request that might be replayed.

   User agents that send a request in early data are expected to retry
   the request when receiving a 425 (Too Early) response status code.  A
   user agent SHOULD retry automatically, but any retries MUST NOT be
   sent in early data.

   In all cases, an intermediary can forward a 425 (Too Early) status
   code.  Intermediaries MUST forward a 425 (Too Early) status code if
   the request that it received and forwarded contained an Early-Data
   header field.  Otherwise, an intermediary that receives a request in
   early data MAY automatically retry that request in response to a 425
   (Too Early) status code, but it MUST wait for the TLS handshake to
   complete on the connection where it received the request.

   The server cannot assume that a client is able to retry a request
   unless the request is received in early data or the Early-Data header
   field is set to "1".  A server SHOULD NOT emit the 425 status code
   unless one of these conditions is met.

   The 425 (Too Early) status code is not cacheable by default.  Its
   payload is not the representation of any identified resource.
```

docs/vc-http-api.yml
OR13: I'm trying real hard not to invent new data models here, so only using things in the vc data model or vp request spec.

docs/vc-http-api.yml
mavarley:
```suggestion
        Notifies this server-holder that the client-holder (caller) is prepared to deliver a presentation compliant with the provided Verifiable Presentation Request query. This is the first part of a presentation exchange over this HTTP API. This server-holder MUST return the required 'domain' and 'challenge' parameters for the Presentation. This server-holder MAY reply with an alternate Verifiable Presentation Request query if it requires something different than the one proposed by the client-holder. The client-holder MUST return a Presentation in accordance with the response parameters.
```

docs/vc-http-api.yml
mavarley:
```suggestion
      description: "Delivery endpoint for a client-holder to provide a presentation in compliance with the Verifiable Presentation Request provided in the Notification call. This server-holder MUST match the 'domain' and 'challenge' to a 'domain' and 'challenge' previously sent in a Notification Response, and not already received."
```

docs/vc-http-api.yml
mavarley: @OR13  above you said "strong SHOULD", here you say "required" - I'll go with required...

docs/vc-http-api.yml
mavarley: Ok, it seems like there is alignment to put the concept of "session" or authorization etc... in a different PR?

OR13: @mavarley I think there is alignment that authorization for all these endpoints need to be addressed consistently. I'd suggest we keep that discussion in issues until we have a clear path forward.

msporny:
> I don't see any objections to this PR, or changes requested and its been up for 12 days... Acknowledging that git lets us make revisions, and changes to the spec, any objections to merging this as is, and taking any further changes in a subsequent PR?

Yes, hard objection. OBJECTION OBJECTION OBJECTION.

https://github.com/w3c-ccg/vc-http-api/pull/168#issuecomment-805064448
https://github.com/w3c-ccg/vc-http-api/pull/168#issuecomment-805276841

TallTed:
@OR13 -- `<echoing @msporny>` OBJECTION! OBJECTION! OBJECTION!

This is the `vc-http-api` -- not the `did-http-api` -- not the `vc-did-api` -- NOTHING that says `DID` as presumptive in any way.

"Merging and iteratively refining" works when general consensus is found on the general frame of what's being merged, and when failure to refine wouldn't lead to OBJECTION -- which is not the case here!

I want to see the specific Use Case, and attendant Requirements, which are found in the VC-HTTP-API UCR, and which this PR is intended to address.  Such citation has not yet been provided, and without it I do not believe we're going to achieve consensus on merging this PR.

OR13:
@TallTed You can use the "request change feature" to block a pull request.... neither you nor @msporny have done that.

__This PR does not impact the use of DIDs with this test suite... and therefore this PR should be be required to fix that issue.__

@msporny here is an end to end example of this API working with the VP Request Spec:

https://github.com/OR13/simple-presentations#simple-presentations

I suggest you use the "request changes feature" and provide concrete path forward... The comments you linked at not actionable.

> A push-based mechanism (issuer notifies holder that they have VCs for them with challenge, holder responds with VP to requesting party).

^ this is completely unrelated to this PR, this PR is about Peer to Peer Presentations, nothing to do with issuers what so ever.

> A pull-based mechanism (holder initiates flow, requesting party sends challenge and VP request, holder responds with VP to requesting party).

This is the flow that was requested of the supply chain folks.... we have agreement on it, thats why I raised this PR....  

The people who are on these calls and what to us this API don't use the word "pull-based" for this flow... because its initiated by the holder who is sending the VP... hence we describe it as a "push based flow"...

Maybe its not clear because this PR is just to a spec, so I implemented the entire thing as an integration test.... It would work over CHAPI or this HTTP API... it's the same exact data model.

If this section of the VP Request spec had anything useful, we would have used it: https://w3c-ccg.github.io/vp-request-spec/#peer-to-peer

An alternative path forward would be for us to abandon the vp request spec and this work item for peer to peer interactions.

The question is not "will folks use http to exchange presentations in a peer to peer manner"... the question is will they use this spec (or the vp request spec) to do it.... and the answer is "obviously not" if there is no way to actually do peer to peer interactions over http.

mprorock (approved):

Therecanbeonlyone1969:
I 100% agree with @OR13 ... @TallTed @msporny ... please, provide actionable changes using the change request feature ... simply saying no is NOT an option, as it is NOT constructive.

msporny (changes requested):
@OR13 wrote:
> I suggest you use the "request changes feature" and provide concrete path forward... The comments you linked are not actionable.

Here are concrete actions anyone pushing for this API could take:

1. Respond to the questions that have been raised throughout the thread (many of the questions have gone unanswered).
2. Define exactly what pull-based vs. push-based means, define the roles in the exchange and their relationship to one another map it to the VC ecosystem, write it down, and get consensus around those definitions.
3. Hold a call to explain the architecture around this API (as a number of people have requested throughout this thread).

You are getting push back because people don't understand the proposal to the depth necessary to believe that the design is solid.

msporny:
> simply saying no is NOT an option, as it is NOT constructive.

That is a mis-characterization of what is going on here. I have requested that concrete things be done here: https://github.com/w3c-ccg/vc-http-api/pull/168#pullrequestreview-622784131

Spamming PRs with "+1, merge it!" when there isn't consensus does not result in the PR getting merged.

peacekeeper:
Like others on this thread, I don't fully understand the proposal either, even though I've been part of SVIP. I thought this is an API for Issuers and Verifiers, but this PR seems to be about holder-to-holder exchange? I thought holder interaction was handled elsewhere (CHAPI, DIDComm, OIDC, etc.)

BTW I'd also prefer if the Issuer API and Verifier API were separate (as it was in the beginning; not sure why they got merged in the first place). If then there's need for a Holder API ("CHAPI over HTTP?"), then that could be a third, separate API.

msporny:
> I don't fully understand the proposal either, even though I've been part of SVIP.

Thinking about this more... one thing occurred to me... this thing (the VC HTTP API) we're talking about doesn't have an actual specification with documented use cases and requirements. All we have is an OAS file to divine details from (and this PR is an example of the fragility of that approach).

We've gotten away without needing that for a while now because we were building off of stuff that was kinda-sorta in the VC specification... but now we're in new territory (kinda sorta), and everyone has slightly different mental models of what's going on and how to accomplish it.

I think @OR13 has the beginnings of use cases here: https://github.com/OR13/simple-presentations#simple-presentations

Thank you @OR13 for putting that together. The code here is also helpful:

https://github.com/OR13/simple-presentations/blob/main/example/src/__tests__/integration.mitm.test.ts

... but raises even more questions. Is this JSON or JSON-LD? Why are we assuming a wallet? Why are these requests stored in a wallet? Why do we need to specify that? Authorized Flows looks intriguing, but doesn't have enough documentation to clearly articulate who produces that? Is there something simpler we could do that the use case above could layer on top?

These are not questions being asked to be obstructionist... they're questions that need to be answered to make sure we're all on the same page when building this API into our products (and again, /the supply chain use cases are valid use cases that I expect everyone on this thread wants to address/).

I think we need at least the following to understand how this API fits in w/ more market verticals, namely:

1. Clearly documented use cases (from more than one market vertical).
2. Clearly documented roles.
3. Clearly documented information flows, for both "pull" and "push" (ideally with diagrams).
4. How the information flows are realized using this PR.

This information might exist across a bunch of issues, PRs, markdown files. It doesn't need to be in ReSpec form -- markdown or a Google Doc would be fine for now. I suggest Google Doc to enable all of us to collaborate.

Thoughts?

TallTed (changes requested):
> @TallTed You can use the "request change feature" to block a pull request.... neither you nor @msporny have done that.

I generally use the "request change feature" when there's a specific change I want. But since you've insisted...

peacekeeper:
> this thing (the VC HTTP API) we're talking about doesn't have an actual specification with documented use cases and requirements

I agree.. Besides the links you mentioned, the architecture page also addresses some of the questions about requirements and scope: https://github.com/w3c-ccg/vc-http-api/blob/main/docs/architecture.md

OR13:
@TallTed To be clear, anyone "requesting changes" is blocking the PR. In the CCG and in the W3C we don't merge over change requests unless there is extreme consensus that the changes are impossible to implement, or address, and there is consensus to adopt them from the rest of the group.

Requesting changes MUST be actionable... one form of this is to say something like "I don't want this feature to ever be supported by this spec".

Requesting changes and making your position clear, so that your concerns can be addressed is a requirement for effective collaboration. Requesting changes is not a bad thing, it is very helpful.

@msporny I'm not sure that raising use cases on a PR that adds endpoints is the best approach. I would propose we tackle use cases as separate issue, and that we clearly define them for all the endpoints in the spec, not just the "new ones".

We already have had a release of this api spec: https://github.com/w3c-ccg/vc-http-api/releases/tag/v0.0.2.plugfest-march-2021

There were no use cases defined in this release ^ 

There are currently 0 integration tests for `provePresentation` and `deriveCredential`.

Perhaps we should start with a Google Doc for use cases for `vp-request-spec` since thats actually what we are talking about here.... thats not this work item, but this work item is being blocked by a lack of use case documentation for vp-request-spec.

I suggest @msporny @dmitrizagidulin @mavarley create a use case document for https://w3c-ccg.github.io/vp-request-spec/#introduction so we can point to it from this API... you can see me trying to do that in this PR already: https://github.com/w3c-ccg/vc-http-api/pull/168/files#diff-74dfbc770d6e21cb974f297ba8f1bc7015ece556537a6202e87b78622728a0cfR659

bumblefudge:
Hey guys, I think we can all agree that a few use-cases and diagrams in a google doc is the most time-efficient starting point. 

My question, just for the sake of starting that google doc, is what is the **target** of this conversation.  Is the expected output:
1. a PR that splits the [architecture file](https://github.com/w3c-ccg/vc-http-api/blob/main/docs/architecture.md) into an architecture file and a UCR, or
2. a new UCR just for this API, or 
3. some use-cases for the VP Request Spec? 

I don't think it will be too hard or controversial to get a _straw man_ with usecases from multiple industries into a form that could be debated on a CCG call or CCG topic call, but it would be a stitch in time for me to know where this conversation is headed so I know how to shape the straw.

troyronda:
I am generally supportive of Holder APIs, including for exchange. I wrote another holder exchange scenario here: https://github.com/w3c-ccg/vc-http-api/issues/176

peacekeeper (changes requested):
I propose that `/presentations/prove` and `/credentials/derive` and the new endpoints in this PR be moved into a separate API spec (could be in the same repo here, but separate file).

I think putting Issuer, Verifier, AND Holder functionality all into a single API spec doesn't make much sense. In practice there will be very few use cases that require all three roles in a single deployment.

OR13:
@peacekeeper  @msporny your wish is my command:

https://github.com/OR13/traceability-api

msporny:
> @peacekeeper @msporny your wish is my command:
> 
> https://github.com/OR13/traceability-api

Forking the VC HTTP API into a competing specification focused only on Traceability is *definitely not* what any of us were asking for.

> We'll be announcing a new work item shortly.

That seems unnecessary and counter-productive... and will lead to objections.

The right course of action here was to call a meeting with the parties that these endpoints affect (as has been requested multiple times in this thread) and have the discussion being requested in this thread. Doing this fork without that discussion just creates unnecessary work and process burden for everyone.

vsnt:
I've added this to the 4/13 CCG agenda and tracking under community issues. https://github.com/w3c-ccg/community/issues/190

OR13:
I agree with @peacekeeper we should make the *vc*-http-api focused on Verifiable Credentials ONLY, and not Verifiable Presentations.

peacekeeper:
> Forking the VC HTTP API into a competing specification focused only on Traceability is _definitely not_ what any of us were asking for.

Right.. Just for the record, that's not what I was suggesting.

> I agree with @peacekeeper we should make the vc-http-api focused on Verifiable Credentials ONLY, and not Verifiable Presentations.

I think I didn't suggest this either.

PR #191, opened 2021-06-03
Added required element on VerifiableCredentials

nissimsan:
Making VC mandatory attrs required on the schema too.

If some are required on the (non-verifiable) Credential, they can be made mandatory there instead.

### DIFF
--- a/docs/versions/v0.0.2/vc-http-api.yml
+++ b/docs/versions/v0.0.2/vc-http-api.yml
@@ -406,6 +406,13 @@ components:
     VerifiableCredential:
       type: object
       description: A JSON-LD Verifiable Credential with a proof.
+      required:
+        - id
+        - type
+        - credentialSubject
+        - proof
+        - issuer
+        - issuanceDate
       allOf:
         - $ref: "#/components/schemas/Credential"
         - type: object
--- a/docs/versions/v0.0.2/vc-http-api.yml
+++ b/docs/versions/v0.0.2/vc-http-api.yml
@@ -407,7 +407,6 @@ components:
       type: object
       description: A JSON-LD Verifiable Credential with a proof.
       required:
-        - id
         - type
         - credentialSubject
         - proof


docs/versions/v0.0.2/vc-http-api.yml
```diff
@@ -406,6 +406,13 @@ components:
     VerifiableCredential:
       type: object
       description: A JSON-LD Verifiable Credential with a proof.
+      required:
+        - id
```
msporny, line 410: `id` isn't required for a VC

msporny (changes requested):
`id` isn't required for a VC, other than that LGTM.

docs/versions/v0.0.2/vc-http-api.yml
OR13: I was surprised to learn this was correct... I have never seen an example of a VC without an `id`, and its pretty rare that it not be a URL....

This is one of those things we should be clearer on in VC Data Model 2.0

msporny: Any sort of short-lived, privacy-preserving "bearer" VC might have this quality. It also helps to compress these sorts of VCs to CBOR-LD if you omit an `id`... saves you between 16-32 bytes for most number-based identifiers... more for URL-based identifiers.

OR13: It also means that `id` is not a safe index or query parameter in an API or database... what primary index would you use for a bearer VC?

msporny: Yes, correct... and you probably wouldn't track bearer VCs, you'd put a short expiration date on them (minutes to weeks at most)... and then fire and forget. This is what we do for the Age Tokens in the TruAge program.

OR13: sure on the "issuer side"... but on the holder side, do you index bearer tokens? how?

msporny: We bundle the bearer VCs together into another VC that does have an `id` (the container VC for a bunch of bearer VCs), IIRC... and then keep track of which bearer tokens have been used in metadata (array index of spent tokens vs. unspent tokens). This is a general pattern that we want to pass by the broader community once we're sure the design works at scale (billions of age tokens over multiple years)... we should be ready to circulate the proposal in the next couple of months. It requires smarter management of VCs in a wallet, and we're not sure if everyone is going to be comfortable with that level of management... or if this is a general pattern that people will want to apply to other use cases. We're trying to end up with a generalized solution instead of a one-off solution for a specific VC type... we're trying to avoid per-VC custom code in wallets.

OR13: > This is a general pattern that we want to pass by the broader community once we're sure the design works at scale (billions of age tokens over multiple years)... 

yes! we have the same issue, and we've been discussing formalizing this meta data in the universal wallet interop spec, while we wait for a better place for it long term.

msporny: Good to know... we've got a bunch of other stuff on our plate before trying to tackle community feedback on this particular proposal (as I'm sure everyone else is busy with more pressing things right now as well). It's useful to know that this seems to be a use case that keeps coming up, and that we probably want to handle it at the "wallet layer"... more in a couple of months. :)

nissimsan: Hi Manu!  :) 
Thanks for the pointer. Really good discussion it spawned.
`id` is removed from the required list.

msporny (approved):

PR #195, opened 2021-06-08
Fix spec links

OR13:
This PR corrects the github pages respec rendering issue, and is paired with changes the the interop spec which points to resources hosted in this repo:

https://w3c-ccg.github.io/traceability-interop/#/issuer

Noting that we ought to use path based versioning for specs in the future, so that external links can be pinned to a stable version

Changes https://w3c-ccg.github.io/vc-http-api:

Screen Shot 2021-06-08 at 1 50 28 PM

to 

Screen Shot 2021-06-08 at 1 57 50 PM


msporny (changes requested):
This PR continues to conflate test reports, with specs, with versioned specs and test reports. I'd rather the group decide how they're going to structure this work than flail at a solution that doesn't have group buy-in.

OR13:
I opened https://github.com/w3c-ccg/vc-http-api/issues/196 to track stable versioning for both the spec and the api defintion files... All this PR does is make the root director return something other than this:

```
<!DOCTYPE html>
<html>
  <head>
    <title>W3C CCG - vc-http-api</title>
    <meta charset="utf-8"/>
  </head>
  <body>
    <p>VC HTTP APIs</p>
    <ul>
      <li><a href="issuer.html">VC Issuer HTTP API</a></li>
      <li><a href="verifier.html">VC Verifier HTTP API</a></li>
      <li><a href="holder.html">VC Holder HTTP API</a></li>
    </ul>
  </body>
</html>
```

^ this is mostly useless. I suggest we do something like this:

https://github.com/w3c-ccg/traceability-interop/blob/main/docs/sections/api-spec.html

WE agreed to split the API files up, but we never agreed to 3 separate spec html source files, in fact I recall there being agreement that we would not do that.

OR13:
Here is an example of the kind of structure I thought we agreed to:

http://w3id.org/traceability/interoperability#/issuer
https://w3c-ccg.github.io/traceability-interop/#/issuer
https://github.com/w3c-ccg/vc-http-api/blob/main/docs/issuer.yml

msporny:
> All this PR does is make the root director return something other than this:

I rolled back the change, as requested. We are left with something not very useful (which was where we were before). This PR avoids the point of contention by putting something in place, which is just going to be moved/changed again (maybe). So, rather than putting a band-aid over the issue, I suggest the group discuss the repo structure that they want, provide a concrete plan that addresses all of the concerns, and then act on it.

OR13:
> We are left with something not very useful (which was where we were before).

agreed, we can leave this PR unmerged and keep it that way while the group discussed the path forward... The PR should be reviewed by the group.

OR13:
This is compatible with my stable version proposal on https://github.com/w3c-ccg/vc-http-api/issues/196

OR13:
Opened https://github.com/perma-id/w3id.org/pull/2211

msporny (approved):
Merging based on https://lists.w3.org/Archives/Public/public-credentials/2021Jun/0184.html

Issue #47, opened 2020-03-12
The API should be able to support multiple proof types


OR13:
My preference would be to use `assertionMethod` by itself for JWT and LD Proofs, however as is noted here: https://github.com/w3c-ccg/vc-issuer-http-api/pull/30#discussion_r391363396

Its technically possible to use a VerificationKey of one type with multiple signature suites... although it is IMO a terrible idea :) 

Consider `ethereumAddress` and secp256k1... does the assertionMethod point to an `ethereumAddress` for use with ECRecover, to a publicKeyHex, a publicKeyBase58, or a publicKeyJwk?

Should all of these verificationMethods work with EcdsaSecp256k1Signature2019?

First some basics about JWA LD Proof Interop... You need to agree to a JWA to use... so if ethereum developers are using `ES-256K-R` and LD Proof devs are using `ES-256K`... you will have problems...

If both JWT and VC Data Model folks agree to a JWT alg we are done, because:

Ed25519Siganture2018 uses EdDSA, so Ed25519VerificationKey2018 => EdDSA.

JsonWebSignature2020 uses EdDSA, so JwsVerificationKey2020 => EdDSA

Obviously if you point to a verification method that is stored as publicKeyBase58, you may be required to do some conversion in order to issue a JWT using off the shelf JOSE libraries...

.For all Verification Methods, if the method uses a `type` which relies on a JWT alg, then you can ask for the a credential as either a JWT or LD-Proof, and `assertionMethod` is all that is required.

I am not aware of any LD Signature Suite that supports multiple signature formats, but if one were to exist, say for RSA, this would be problematic... however, IMO it would be best for such a suite NOT to exist, because letting someone decide to encode their signature as RS256, PS256, or base64url raw bytes is not a good idea for a signature suite.

OR13:
@tplooker I think this may be partially addressed by some of your work. can you cross link anything thats relevant... I would prefer not to start on this, if its going to conflict with what you are doing.

OR13:
I think we should close this issue.

OR13:
We will address support for multiple suites in other issues/

PR #197, opened 2021-06-11
Add proof type issue/present option

clehner:
Signature suite [Ethereum EIP712 Signature 2021](https://uport-project.github.io/ethereum-eip712-signature-2021-spec/) (proposed work item: https://github.com/w3c-ccg/community/issues/194) does not define a new verification method but only a signing method (proof type), and is meant to be used with existing verification methods `EcdsaSecp256k1VerificationKey2019` or `EcdsaSecp256k1RecoveryMethod2020`. For proof creation with `vc-http-api`, the `verificationMethod` option of the issue/present credential options is then no longer sufficient to identify the type of proof to create, since there is not a one-to-one correspondance of verification method type to proof type. This PR adds a type option for use in situations such as this where there is more than one type of proof that could be issued.

Edit: relevant issue #47.
This also could be useful if one wants to choose between using e.g. `Ed25519Signature2018` or `Ed25519Signature2020` when the issuer/holder is using `did:key`.

### DIFF
--- a/docs/components/IssueCredentialOptions.yml
+++ b/docs/components/IssueCredentialOptions.yml
@@ -17,6 +17,9 @@ components:
       additionalProperties: false
       description: Options for specifying how the LinkedDataProof is created.
       properties:
+        type:
+          type: string
+          description: The type of the proof. Default is an appropriate proof type corresponding to the verification method.
         verificationMethod:
           type: string
           description: The URI of the verificationMethod used for the proof. Default assertionMethod URI.
@@ -41,6 +44,7 @@ components:
               description: The type of credential status to issue the credential with
       example:
         {
+          "type": "Ed25519Signature2018",
           "verificationMethod": "did:example:123#z6MksHh7qHWvybLg5QTPPdG2DgEjjduBDArV9EF9mRiRzMBN",
           "proofPurpose": "assertionMethod",
           "created": "2020-04-02T18:48:36Z",
--- a/docs/components/PresentCredentialOptions.yml
+++ b/docs/components/PresentCredentialOptions.yml
@@ -17,6 +17,9 @@ components:
       additionalProperties: false
       description: Options for specifying how the LinkedDataProof is created.
       properties:
+        type:
+          type: string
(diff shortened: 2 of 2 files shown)


docs/components/IssueCredentialOptions.yml
```diff
@@ -41,6 +44,7 @@ components:
               description: The type of credential status to issue the credential with
       example:
         {
+          "type": "https://w3id.org/security#Ed25519Signature2018",
```
msporny, line 47: Now the question becomes... why is `type` a fully qualified URL and `assertionMethod` is not. :)

We could go two ways on this -- include an `@context` (probably overkill), or require full URLs in `options` (probably annoying).

docs/components/PresentCredentialOptions.yml
```diff
@@ -34,6 +37,7 @@ components:
           description: The intended domain of validity for the proof. For example website.example
       example:
         {
+          "type": "https://w3id.org/security#Ed25519Signature2018",
```
msporny, line 40: Same issue as above.

docs/components/IssueCredentialOptions.yml
clehner: Amended to use the compact form, for consistency with the other options and existing conventions for proof objects.

docs/components/PresentCredentialOptions.yml
clehner: Amended to use the compact form, for consistency.
https://github.com/w3c-ccg/vc-http-api/compare/322fba6..173c26b0df3ca1a1406f7b8d4469ef5f15b3afcb

clehner:
@OR13 I think in #47 you were suggesting to not use a key of one type with multiple signature suites; I wonder if that still your opinion, or do you think this is a good approach?

OR13:
@clehner I am no longer certain that `proof.type` -> `verificationMethod.type` should be a 1-1 binding, but this work item is probably not the place to sort this out... from a security perspective, its possible to verify a VC  where this is not a 1-1 mapping, I tend to treat that as the rule of law....

msporny (approved):

PR #200, opened 2021-06-17
Suggested improvements on the Presentations schema

nissimsan:
- make credentials plural
- refer to the VC schema, not just an object

### DIFF
--- a/docs/versions/v0.0.2/vc-http-api.yml
+++ b/docs/versions/v0.0.2/vc-http-api.yml
@@ -468,11 +468,11 @@ components:
           type: object
           description: The holder - will be ignored if no proof is present since there is no proof of authority over the credentials
           nullable: true
-        "verifiableCredential":
+        "verifiableCredentials":
           type: array
           description: The Verifiable Credentials
           items:
-            type: object
+            $ref: "#/components/schemas/VerifiableCredential"
       example:
         {
           "@context":
--- a/docs/versions/v0.0.2/vc-http-api.yml
+++ b/docs/versions/v0.0.2/vc-http-api.yml
@@ -468,7 +468,7 @@ components:
           type: object
           description: The holder - will be ignored if no proof is present since there is no proof of authority over the credentials
           nullable: true
-        "verifiableCredentials":
+        "verifiableCredential":
           type: array
           description: The Verifiable Credentials
           items:


docs/versions/v0.0.2/vc-http-api.yml
```diff
@@ -468,11 +468,11 @@ components:
           type: object
           description: The holder - will be ignored if no proof is present since there is no proof of authority over the credentials
           nullable: true
-        "verifiableCredential":
+        "verifiableCredentials":
```
msporny, line 471: This would be a violation of the VC Data Model, wouldn't it? https://w3c.github.io/vc-data-model/#presentations-0

OR13: yep, linked data "convention" is to use "singular" for arrays, and camelCase for field names... 2 boats that have sailed long ago, in a direction of less readability :)

none the less, the convention (and spec conformance) MUST be preserved.


> verifiableCredential
> If present, the value of the verifiableCredential property MUST be constructed from one or more verifiable credentials, or of data derived from verifiable credentials in a cryptographically verifiable format.


technically `verifiableCredential` is not required and `verifiableCredentials` is not forbidden, but we can all see how terrible pulling on that thread would be.

OR13:
```suggestion
        "verifiableCredential":
```

OR13 (changes requested):
see suggested changes

OR13 (approved):

msporny (approved):

OR13 (approved):

peacekeeper:
I agree with the change itself, but is this the right file to change? I think this file is the "stable" version 0.0.2 that shouldn't be changed anymore.

Instead, we should make that change in the current unstable version, i.e. here: https://github.com/w3c-ccg/vc-http-api/blob/main/docs/components/Presentation.yml#L40

Or am I missing something?

OR13 (changes requested):
@nissimsan @peacekeeper is correct, this change should be made here:

https://github.com/w3c-ccg/vc-http-api/tree/main/docs <- (latest / next)

(not to a fixed version / v0.0.2)

nissimsan:
Of course, thanks @peacekeeper. 
Skipping the git gymnastics and creating a new PR for this.

Issue #207, opened 2021-06-25
Verifier API should return verified VCs


msporny:
@David-Chadwick wrote:
> Looking at your latest version of the verifier API, I have the following suggestion to make for when the input is a signed VP. The 200 OK code should return the set of validated credentials in standard W3C format, without their proofs.

> In the case where the VP for verification is encoded as a JWT, with the encapsulated VCs also being JWT proofed, each JWT VC has removed several fields from the W3C VC and put them as claims in the JWT. Thus it would be good if the verifier API, after validating the JWTs, could put these claims back into the W3C VC, and return this as a standard W3C formatted credential without any proof. This would then mirror what the issuer does, as it is given a standard W3C credential and it turns it into a JWT. In this way, whether the Verifier API verifies a JWT VP, a LD-Proof VP, or an mDL VP, they will all return standard W3C credentials without any proofs and the API caller will have a consistent response regardless of how the VP was formatted

mavarley:
I agree that having one expected output format from the API would be a very good idea.

Is this ready for PR?

msporny:
[...]

Issue #42, opened 2020-03-12
Define JWT Encoding of VCs


OR13:
Per the spec: https://www.w3.org/TR/vc-data-model/#jwt-encoding

We should eliminate a good chunk of the optionality here, at least for the first version of the spec that actually supports JWTs...

> For backward compatibility with JWT processors, the following JWT-registered claim names MUST be used instead of, or in addition to, their respective standard verifiable credential counterparts: exp, iss, sub, et...

IMO the first version of this spec should duplicate these so that the credential that is built looks the same when decoded. (ONLY support "addition to").

This will also allow the simpler API to use the same payload for LD Proof and JWT Proof variants.... The alternative will be to use the more complex endpoints...

OR13:
This issue is being addressed in a new work item:

https://github.com/w3c-ccg/community/issues/198

PR #208, opened 2021-06-25
Support JWT

clehner:
Add support for [JWT](https://www.w3.org/TR/vc-data-model/#json-web-token) [proof format](https://www.w3.org/TR/vc-data-model/#proof-formats) in the API.

The `VerifiableCredential` and `VerifiablePresentation` types are changed to allow a `string` value, using `oneOf`. (Changes in the  associated yaml files in the `object` section are indentation only).

An option for issuing or presenting a credential ~~is~~ could be (re-)added, `proofFormat`, with possible values "jwt" or "ldp". ~~Would there be more idiomatic existing terms to use here?~~

This API change assumes that the existing routes and options handling LDPs can be used with JWTs. ~~Should it be documented how that should work? e.g. the `domain` option could correspond to the `aud` claim.~~ How this works, in addition to what VC Data Model specifies, could be determined by the `type` option in a issue/present/verify call.

~~Should there be an issue/present option for "additional JWT claims" that the caller may want to add outside the VC/VP?~~

Related issues: #42, #207

### DIFF
--- a/docs/components/VerifiableCredential.yml
+++ b/docs/components/VerifiableCredential.yml
@@ -13,40 +13,45 @@ paths:
 components:
   schemas:
     VerifiableCredential:
-      type: object
-      description: A JSON-LD Verifiable Credential with a proof.
-      allOf:
-        - $ref: "./Credential.yml#/components/schemas/Credential"
+      oneOf:
+        - type: string
+          description: A JWT Verifiable Credential
+          example:
+            eyJhbGciOiJFZERTQSIsImtpZCI6ImRpZDprZXk6ejZNa2lWcHdBMjQxZ3VxdEtXQWtvaEhwY0FyeTdTOTRRUWI2dWtXM0djQ3N1Z2JLI3o2TWtpVnB3QTI0MWd1cXRLV0Frb2hIcGNBcnk3Uzk0UVFiNnVrVzNHY0NzdWdiSyJ9.eyJpc3MiOiJkaWQ6ZXhhbXBsZToxMjMiLCJuYmYiOjE2MjQ2MzU0MTYsImp0aSI6Imh0dHA6Ly9leGFtcGxlLmdvdi9jcmVkZW50aWFscy8zNzMyIiwic3ViIjoiZGlkOmV4YW1wbGU6MTIzIiwidmMiOnsiQGNvbnRleHQiOlsiaHR0cHM6Ly93d3cudzMub3JnLzIwMTgvY3JlZGVudGlhbHMvdjEiLCJodHRwczovL3d3dy53My5vcmcvMjAxOC9jcmVkZW50aWFscy9leGFtcGxlcy92MSJdLCJ0eXBlIjpbIlZlcmlmaWFibGVDcmVkZW50aWFsIiwiVW5pdmVyc2l0eURlZ3JlZUNyZWRlbnRpYWwiXSwiY3JlZGVudGlhbFN1YmplY3QiOnsiaWQiOiJkaWQ6ZXhhbXBsZToxMjMiLCJkZWdyZWUiOnsibmFtZSI6IkJhY2hlbG9yIG9mIFNjaWVuY2UgYW5kIEFydHMiLCJ0eXBlIjoiQmFjaGVsb3JEZWdyZWUifX0sImlzc3VhbmNlRGF0ZSI6IjIwMjEtMDYtMjVUMTU6MzY6NTZaIn19.FnYoavlqTQSbaVZ15TNo3mzlo9nbKtzzZkSHjwngD6yR-Z-9iXXSZ4ZZu17bv90Z279RUGzY0M89-fIHrmP5DA
         - type: object
-          properties:
-            proof:
-              $ref: "./LinkedDataProof.yml#/components/schemas/LinkedDataProof"
-      example:
-        {
-          "@context":
-            [
-                "https://www.w3.org/2018/credentials/v1",
-                "https://www.w3.org/2018/credentials/examples/v1",
-            ],
-          "id": "http://example.gov/credentials/3732",
-          "type": ["VerifiableCredential", "UniversityDegreeCredential"],
-          "issuer": "did:example:123",
-          "issuanceDate": "2020-03-16T22:37:26.544Z",
-          "credentialSubject":
+          description: A JSON-LD Verifiable Credential with a proof.
+          allOf:
+            - $ref: "./Credential.yml#/components/schemas/Credential"
+            - type: object
+              properties:
+                proof:
+                  $ref: "./LinkedDataProof.yml#/components/schemas/LinkedDataProof"
+          example:
             {
-              "id": "did:example:123",
-              "degree":
+              "@context":
+                [
+                    "https://www.w3.org/2018/credentials/v1",
+                    "https://www.w3.org/2018/credentials/examples/v1",
+                ],
+              "id": "http://example.gov/credentials/3732",
+              "type": ["VerifiableCredential", "UniversityDegreeCredential"],
+              "issuer": "did:example:123",
+              "issuanceDate": "2020-03-16T22:37:26.544Z",
+              "credentialSubject":
                 {
-                  "type": "BachelorDegree",
-                  "name": "Bachelor of Science and Arts",
+                  "id": "did:example:123",
+                  "degree":
+                    {
+                      "type": "BachelorDegree",
+                      "name": "Bachelor of Science and Arts",
+                    },
                 },
-            },
-          "proof":
-            {
-              "type": "Ed25519Signature2018",
-              "created": "2020-04-02T18:28:08Z",
-              "verificationMethod": "did:example:123#z6MksHh7qHWvybLg5QTPPdG2DgEjjduBDArV9EF9mRiRzMBN",
-              "proofPurpose": "assertionMethod",
-              "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..YtqjEYnFENT7fNW-COD0HAACxeuQxPKAmp4nIl8jYAu__6IH2FpSxv81w-l5PvE1og50tS9tH8WyXMlXyo45CA",
-            },
-        }
+              "proof":
+                {
+                  "type": "Ed25519Signature2018",
+                  "created": "2020-04-02T18:28:08Z",
+                  "verificationMethod": "did:example:123#z6MksHh7qHWvybLg5QTPPdG2DgEjjduBDArV9EF9mRiRzMBN",
+                  "proofPurpose": "assertionMethod",
+                  "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..YtqjEYnFENT7fNW-COD0HAACxeuQxPKAmp4nIl8jYAu__6IH2FpSxv81w-l5PvE1og50tS9tH8WyXMlXyo45CA",
+                },
+            }
--- a/docs/components/VerifiablePresentation.yml
+++ b/docs/components/VerifiablePresentation.yml
@@ -13,58 +13,63 @@ paths:
 components:
   schemas:
     VerifiablePresentation:
-      type: object
-      description: A JSON-LD Verifiable Presentation with a proof.
-      allOf:
-        - $ref: "./Presentation.yml#/components/schemas/Presentation"
+      oneOf:
+        - type: string
+          description: A JWT Verifiable Presentation
+          example:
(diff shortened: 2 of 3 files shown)


msporny:
I expect this to be an unpopular opinion... but with the new work on BBS+, the lack of interoperable examples of JWT VCs, and a lack of examples of interoperable tests for JWT VCs... I'm concerned about the work this might create for the VC HTTP API group.

> An option for issuing or presenting a credential is added, proofFormat, with possible values "jwt" or "ldp". Would there be more idiomatic existing terms to use here?

How many implementers are going to support the JWT format? Are there implementers that would like to never support the JWT format, if possible? Are there other proofFormats in the works? Can we just express the proofFormat via the `type`?

> This API change assumes that the existing routes and options handling LDPs can be used with JWTs. Should it be documented how that should work? e.g. the domain option could correspond to the aud claim.

Feels like adding a lot of work to the group at this stage... especially given that none of the implementers to date have requested that the API support these features, and especially since the JsonWebSiganture2020 suite exists: https://w3c-ccg.github.io/lds-jws2020/

> Should there be an issue/present option for "additional JWT claims" that the caller may want to add outside the VC/VP?

Again, feels like scope creep.

I'd prefer that those wanting JWT support write the VC JWT test suite first (or demonstrate interop of some kind before requiring this API to support it).

David-Chadwick:
JWT interworking may be relatively easy to achieve. The VCs that we create can be validated by the jwt.io debugger, which supports a much larger audience than just the W3C VC implementors. Currently our VP syntax can be validated but not the signature as the jwt.io debugger does not support DIDs.
So I would be happy to start to perform interworking tests with other JWT implementors if they ping me.

OR13:
We have some library support for JWT baseed VCs and JWS generally (since they are a building block for many popular VC proofs like `Ed25519Signature2018` and `EcdsaSecp256k1Siganture2019`.

I feel a bit torn on this issue.

## Arguments for Supporting JWT

- VC-JWT is described in the spec
- JOSE / JWA tooling is more widely available and understood, since its older
- Without better interoperability tests, its easier for big guys like Microsoft to ignore the W3C VC Data Model and call whatever JWT thing they make a VC, and there is no proof that its not interoperable.

## Arguments against Supporting JWT

- VC-JWT doesn't add a lot to vanilla JOSE
- VC-JWT failed to ensure interoperability by allowing lots of optional fields
- VC-JWT date format includes milliseconds and is literally incompatible with W3C / XSD Date time.
- VC-JWT never defined  a resolution or dereferencing for key material or proof purposes so we would be guessing about how that maps to DIDs (though I suspect  over time there may now we rough consensus to do the thing that aligns with LD proofs)
- VC-JWT does not support bls12381 / bbs or Schnorr for secp256k1, and we need to update the registry to do so... 
- `vp` and `vc` are STILL not registered: https://w3c.github.io/vc-data-model/#iana-considerations, https://www.iana.org/assignments/jose/jose.xhtml

## Conclusion

Of all these issues, the most frustrating for me is the registration stuff.  JOSE / JWT are undeniably easier to implement, and understand... yet innovation with JWT is controlled by registry updates... At the same time, many JWT implementers have been actively hostile to the major improvement that VCs have over JWTs, which is semantics for the claims... This has led me to generally dislike JWT implementations, but I have to admit, that my views of JWT have been distorted by the implementers and at its core, I still think VC-JWT has a place in the world, especially with better guidance on required vs optional fields. 

Unfortunately, I don't know that there is consensus on that part, for example, in our implementations we don't verify VC-JWTs that fail to define their claims in their context... yet this (verifying the credential semantics) is not a requirement of the spec, yet `@context` is... and... 

https://spec.smarthealth.cards/credential-modeling/#mapping-into-the-w3c-vc-data-model

```
"@context": [
  "https://www.w3.org/2018/credentials/v1",
  {
    "@vocab": "https://smarthealth.cards#",
    "fhirBundle": {
      "@id": "https://smarthealth.cards#fhirBundle",
      "@type": "@json"
    }
  }
]
```

Telling people to add a normatively required field:

https://w3c.github.io/vc-data-model/#contexts

> Verifiable credentials and verifiable presentations MUST include a @context property.

I would 100% support VC-JWT, if we don't loose any semantic checks, but this kind of "squint at a normal JWT and call it a VC / VP"... is incredibly harmful and undermines everything we are working for here, w.r.t. interoperability and standards compliance.

I'm not sure other VC-JWT implementers would agree with me, and frankly, the spec should have solved for this issue...

So in the interest of making it possible to support VC-JWT, can we agree to the following:

1. "Verifiable credentials and verifiable presentations MUST include a @context property."
2. vc['@context'] MUST be signed over when constructing the payload for the JWS that is produced for a VC-JWT
3. vp['@context'] MUST be signed over when constructing the payload for the JWS that is produced for a VP-JWT

OR13 (changes requested):
There is a lack of consensus regarding VC-JWT normative requirements, we've observed significantly different / non interoperable formats in the wild.

I am not against supporting VC-JWT, but we will need to get clearer consensus on what that means (especially w.r.t. issues like signatures and verification), before we can commit to api / spec text on this subject.

In the interest of making this "change request" actionable, perhaps we can get the VC-JWT data model and signature / verification requirements documented explicitly so we can point to that when we say "we support VC-JWT".

Sakurann:
@OR13 vc vp are registered in JWT registry: https://www.iana.org/assignments/jwt/jwt.xhtml

> vp and vc are STILL not registered: https://w3c.github.io/vc-data-model/#iana-considerations, https://www.iana.org/assignments/jose/jose.xhtml

OR13:
@Sakurann Thanks! I was searching the wrong place.

Sakurann:
@OR13 what do you mean by
> VC-JWT failed to ensure interoperability by allowing lots of optional fields

`@context` is being included in VC-JWT too

David-Chadwick:
@OR13 VC-JWT failed to ensure interoperability by allowing lots of optional fields

Sorry I dont understand this comment. VCs have an infinite number of optional fields, so JWTs are inheriting this from VCs.

David-Chadwick:
> There is a lack of consensus regarding VC-JWT normative requirements, we've observed significantly different / non interoperable formats in the wild.
> 
The best way to address this is by the VC WG clarifying any ambiguities or omissions in the V2 Recommendation that is now being worked on.

David-Chadwick:
@OR13 yet innovation with JWT is controlled by registry updates..

This restriction has now been removed by adding the vc and vp data types to the IANA registry. This means that any new properties in VCs or VPs are automatically encapsulated in the vc and vp claims and therefore do not need to be registered in IANA. The normal @context registration procedure is used.

OR13:
@David-Chadwick I mean the registries control `alg` such as ES256, ES256K, etc...

@Sakurann and @David-Chadwick "the normal `@context` process is being used".... how?

certainly its not being following by Microsoft and Apple in SmartHealthCards....

injecting payload attributes would break the signature... signing over the payload without the context is not normatively compliant.

peacekeeper:
I am generally in favor of this. In the very first version of the API (=long obsolete), there were options to specify the credential format as well as proof format (the limitation to JSON-LD only happened later):

https://github.com/w3c-ccg/vc-http-api/blob/main/docs/versions/v0.0.0/vc-http-api.yml#L481

But I also understand concerns of others in this thread about added complexity. If this addition creates problems for implementers, maybe better to first discuss on a call.

David-Chadwick:
@OR13 certainly its not being following by Microsoft and Apple in SmartHealthCards..

whenever did Microsoft conform to standards? They have always twisted them in some way: Kerberos, LDAP, email and now VCs. This does not surprise me at all.

clehner:
> Are there other proofFormats in the works?

In the `proofFormat` option in `v0.0.0` which @peacekeeper pointed out, there was a third possible value, "blockcerts". But I don't know the status of that format. I would be interested to know why those options were removed (and `credentialFormatOptions` and ` proofFormatOptions`) - I'm unable to find discussion of it.

@David-Chadwick in #209 has mentioned OIDC's encapsulation format; I don't know if that would be considered a proof format for VC Data Model.

VC Data Model lists just the two [proof formats](https://www.w3.org/TR/vc-data-model/#proof-formats); it also references [VC Extension Registry](https://w3c-ccg.github.io/vc-extension-registry/), but does not explicitly mention proof formats as an extension point.

> Can we just express the proofFormat via the type?

I think that could work. A proof type(s) could be defined for use here - like the "synthetic" `JwtProof2020` in DIF's [did-jwt-vc](https://github.com/decentralized-identity/did-jwt-vc/) implementation - that could define mapping between JWT and VP/VC with Proof object, building on what is already in VC Data Model v1.0. An exception would be if a proof type supported both JWT format and LDP-only format, a format option might be needed to request one or the other.

In the interest of simplicity, I amend the commit to remove `proofFormat`. Since VC JWTs using JWS don't have a proof type property, to enable verification according to specific JWT-supporting proof type, I add "type" to VerifyOptions (similar to #197).

David-Chadwick:
@OR13 wrt to @context, I can only speak for our own implementation. We add @contexts to our VCs for the subject properties being defined by our issuers. So as far as I know our VCs are 100% compliant.

David-Chadwick:
@clehner "I don't know if that would be considered a proof format for VC Data Model."

No, the OIDC encapsulation is there to tell the relying party whether the returned VP is a JWT or LDProof'ed VP. That's all. You can see the JSON here
https://bitbucket.org/openid/connect/src/master/openid-connect-4-verifiable-presentations/openid-connect-4-verifiable-presentations-1_0.md

David-Chadwick:
@msporny We support any version that complies with the standard. Full stop.
And we clarify in v2 of the standard any ambiguities that still exist for people (Personally, because I helped write the Recommendation I did not find any of the current text confusing. But then I might have had implicit knowledge that others are missing and that needs to be documented)

Sakurann:
Microsoft's Verifiable Credentials Service adds @context to the VCs - you can check it if you try sample issuer/verifier in https://docs.microsoft.com/en-us/azure/active-directory/verifiable-credentials/get-started-verifiable-credentials

For Smart Health Cards, there is a mapping at
https://spec.smarthealth.cards/credential-modeling/#mapping-into-the-w3c-vc-data-model

TallTed:
@Sakurann -- Please codefence (by wrapping in backticks, either single or triple) any `@`entity you put in github markdown, e.g.,``` `@context` ```.  Otherwise, they get pinged about many discussions which are of no interest to them.  (It's worth editing [your recent comment](https://github.com/w3c-ccg/vc-http-api/pull/208#issuecomment-869881849).)

OR13:
@Sakurann 

> For Smart Health Cards, there is a mapping at
https://spec.smarthealth.cards/credential-modeling/#mapping-into-the-w3c-vc-data-model

This proves that smart health cards are not VC Data Model conformant... you don't need a mapping when you follow the normative requirements in a standard.

> Microsoft's Verifiable Credentials Service adds @context to the VCs - you can check it if you try sample issuer/verifier in https://docs.microsoft.com/en-us/azure/active-directory/verifiable-credentials/get-started-verifiable-credentials

I cannot confirm the behavior from the link, can you link directly to a VC-JWT in either compact or decoded form?

OR13:
Here are the test vectors we use for our VC-JWT implementation:

https://github.com/transmute-industries/vc.js/tree/master/packages/vc.js/src/__fixtures__/test-vectors/vc-jwt

clehner:
@OR13

> In the interest of making this "change request" actionable, perhaps we can get the VC-JWT data model and signature / verification requirements documented explicitly so we can point to that when we say "we support VC-JWT".

Where might be a good place for this?

msporny:
> Where might be a good place for this?

Set up a `vc-jwt-test-suite` repository as a CCG work item, fill it with vc-jwt tests, and get multiple organizations to support passing all of the tests.

David-Chadwick:
@msporny "ideally, this would just be a vc-signatures-test-suite which would cover LD Signatures as well"
Ideally it would NOT include LD Signatures as well, but would use standard JWT defined algorithms and signature procedures. JWT does not need canonicalisation for example. JWT does however support X.509 PKCs, another well used standard.

OR13:
I am in favor of closing this PR, when:

There exists a w3c ccg work item titled: `vc-jwt-test-suite`.

Until then , the PR should remain up, to gather feedback / comments.

@clehner would you mind making the proposal and linking it back here?

clehner:
@OR13 @msporny I can do that. I'll need a couple days to prepare.

bumblefudge:
Update: the C&C meeting (and tomorrow's CCG meeting as well) are intended as requirements-gathering and input-gathering sessions to help scope and inform the test suite work item. Both sessions are public and people with relevant experience to share and compare are invited to make either or both, observing the IPR policies of the respective venues.

clehner:
Work item proposal for JWT interop testing is opened at https://github.com/w3c-ccg/community/issues/198

msporny:
> Work item proposal for JWT interop testing is opened at [w3c-ccg/community#198](https://github.com/w3c-ccg/community/issues/198)

Given that we now have this work item proposal, can we close this PR?

clehner:
@msporny yes, closing.

sloops77:
@msporny @clehner 
I'm Andres Olave from the Velocity Network. After our call a few weeks ago, I am exploring adding support for VC-API for some use cases for the Velocity Network. However, we will only be able to have the network ecosystsem support the vc-jwt's in the short to medium term as we do not want to burden them with two credential formats. Is this something that can be revisited since https://github.com/w3c-ccg/community/issues/198 is closed

msporny:
@sloops77 wrote:
> I have been exploring adding support for VC-API for some use cases for the Velocity Network. However, we will only be able to have the network ecosystem support vc-jwt's (for the short to medium term) as we do not want to burden members with two credential formats. Is this something that can be revisited since [w3c-ccg/community#198](https://github.com/w3c-ccg/community/issues/198) is closed

The VC API is capable of supporting an arbitrary number of credential formats, so yes, we've just been waiting for someone that's interested in actually implementing this feature. The easiest place to start would be with these two endpoints:

`[/credentials/issue](https://w3c-ccg.github.io/vc-api/#issue-credential)` and `[/credentials/verify](https://w3c-ccg.github.io/vc-api/#verify-credential)`

The presumption is that the endpoint for `/credentials/issue` would be configured to issue VC-JWTs (you'd give it the same parameters, but the result would be a VC-JWT held in a field called `vcJwt`, and the endpoint for verifying would take a parameter called `vcJwt`. 

I'll note that if you go this route, it'll be highly experimental as everyone else that has demonstrated interop in JFF Plugfest #3 (17+ platforms) using the VC API has taken the `application/vc+ld+json` route with Data Integrity Proofs. There are people out there issuing VC-JWTs, but due to how the VC-JWT spec is changing, and how implementations have taken different paths there, demonstrating interop will be more difficult until there is better tooling/testing available.

In any case, if you are committed to doing VC-JWT (which I suggest you avoid until things settle down a bit more there), then I'd start with the above.

PR #224, opened 2021-07-30
Add group resolutions to specification to signal editorial intent.

msporny:
This is just an editorial PR that adds all the resolutions the group has made to the spec to signal changes that we intend to make to the specification.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-http-api/pull/224.html" title="Last updated on Aug 17, 2021, 10:20 PM UTC (992a96c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-http-api/224/41dbd3e...992a96c.html" title="Last updated on Aug 17, 2021, 10:20 PM UTC (992a96c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -189,8 +189,8 @@ TBD
         <p class="issue" title="Apply group resolution">
 Convert the following group resolution into specification text:<br><br>
 <a href="https://w3c-ccg.github.io/meetings/2021-07-20-vchttpapi/#resolution-1">
-RESOLUTION</a>: If a caller desires to build a VC incrementally of from a
-template, the caller can use another API to do so and then send the fully formed
+RESOLUTION</a>: If a caller desires to build a VC incrementally or from a
+template, the caller can use another API to do so, and then send the fully formed
 VC to the VC HTTP API for processing.
         </p>
 
--- a/index.html
+++ b/index.html
@@ -182,6 +182,19 @@ TBD
           </tr>
         </tbody>
       </table>
+
+      <section>
+        <h3>Out of Scope</h3>
+
+        <p class="issue" title="Apply group resolution">
+Convert the following group resolution into specification text:<br><br>
+<a href="https://w3c-ccg.github.io/meetings/2021-07-20-vchttpapi/#resolution-1">
+RESOLUTION</a>: If a caller desires to build a VC incrementally of from a
+template, the caller can use another API to do so and then send the fully formed
+VC to the VC HTTP API for processing.
+        </p>
+
+      </section>
     </section>
 
     <section class="informative">
@@ -297,6 +310,43 @@ A <dfn>conforming VC HTTP API server</dfn> is ...
       <h3>Authorization</h3>
       <p>
       </p>
+
+      <p class="issue" title="Apply group resolution">
+Convert the following group resolution into specification text:<br><br>
+<a href="https://w3c-ccg.github.io/meetings/2021-07-13-vchttpapi/#resolution-1">
+RESOLUTION</a>: The VC HTTP API work item group will separate GNAP from OAuth2
+until it is clear how much extra work GNAP would add within the scope of the
+specification.
+      </p>
+
+      <p class="issue" title="Apply group resolution">
+Convert the following group resolution into specification text:<br><br>
+<a href="https://w3c-ccg.github.io/meetings/2021-07-13-vchttpapi/#resolution-2">
+RESOLUTION</a>: One of the authorization mechanisms defined for the VC-HTTP-API
+MUST be OAuth 2 Bearer tokens.
+      </p>
+
+      <p class="issue" title="Apply group resolution">
+Convert the following group resolution into specification text:<br><br>
+<a href="https://w3c-ccg.github.io/meetings/2021-07-13-vchttpapi/#resolution-3">
+RESOLUTION</a>: How a VC HTTP API server validates an authorization token is out
+of scope.
+      </p>
+
+      <p class="issue" title="Apply group resolution">
+Convert the following group resolution into specification text:<br><br>
+<a href="https://w3c-ccg.github.io/meetings/2021-07-13-vchttpapi/#resolution-4">
+RESOLUTION</a>: One of the authorization protocols that will be defined for use
+in the VC-HTTP-API MUST be OAuth 2 Client Credentials.
+      </p>
+
+      <p class="issue" title="Apply group resolution">
+Convert the following group resolution into specification text:<br><br>
+<a href="https://w3c-ccg.github.io/meetings/2021-07-13-vchttpapi/#resolution-5">
+RESOLUTION</a>: One of the authorization mechanisms defined for the VC HTTP API
+MUST be GNAP key-bound access tokens.
+      </p>
+
     </section>
 
     <section>
@@ -329,6 +379,16 @@ A <dfn>conforming VC HTTP API server</dfn> is ...
     <h2>Security Considerations</h2>
     <p>
     </p>
+
+    <p class="issue" title="Apply group resolution">
+Convert the following group resolution into specification text:<br><br>
+<a href="https://w3c-ccg.github.io/meetings/2021-07-27-vchttpapi/#resolution-1">
+RESOLUTION</a>: The VC HTTP API Work Item group will not define a "phone home"
+mechanism at present as it may create a privacy danger when a verifier contacts
+an issuer directly about a particular credential. Vendors are free to implement
+such a system, but we are not going to define it in the standard at present.
+    </p>
+
   </section>
 
   <section class="appendix">


index.html
```diff
@@ -182,6 +182,19 @@ <h2>Design Goals</h2>
...
       </table>
+
+      <section>
+        <h3>Out of Scope</h3>
+
+        <p class="issue" title="Apply group resolution">
+Convert the following group resolution into specification text:<br><br>
+<a href="https://w3c-ccg.github.io/meetings/2021-07-20-vchttpapi/#resolution-1">
+RESOLUTION</a>: If a caller desires to build a VC incrementally of from a
+template, the caller can use another API to do so and then send the fully formed
```
TallTed, line 193:
```suggestion
RESOLUTION</a>: If a caller desires to build a VC incrementally or from a
template, the caller can use another API to do so, and then send the fully formed
```

OR13 (changes requested):
do not merge, does not reflect consensus.

msporny:
Do not merge this PR until we have clarity over the "what did majority mean" and/or "what's the process here" questions.

index.html
```diff
@@ -297,6 +310,43 @@ <h2>The VC HTTP API</h2>
...
+      </p>
+
+      <p class="issue" title="Apply group resolution">
+Convert the following group resolution into specification text:<br><br>
+<a href="https://w3c-ccg.github.io/meetings/2021-07-13-vchttpapi/#resolution-4">
+RESOLUTION</a>: One of the authorization protocols that will be defined for use
+in the VC-HTTP-API MUST be OAuth 2 Client Credentials.
+      </p>
+
+      <p class="issue" title="Apply group resolution">
```
jandrieu, line 343: This resolution was passed with insufficient consensus. Contrary to all other CCG meetings, someone made a decision that "majority" voting would be sufficient for resolution approval. I objected at the time of the resolution and was told the decision had been made to use majority vote and so that's what we were doing for that meeting. 

Since the entirely of my previous experience with the W3C has always enabled either email or regular calls as a forum for individuals to register objections to decisions made in task forces and smaller subgroups, the lack of any mechanism for the several people on the call who voted against this resolution constitutes a process failure at best, one that, frankly, looks suspiciously like someone set up the majority proposal to intentionally push through their preferred features.

To avoid the appearance railroading this decision, I'd like it removed from this PR. If desired, a new issue could be created to establish whether or not the CCG as a whole supports this resolution. Or this can be brought up in a call where we rely on more standard W3C notions of consensus development.

TL;DR: when the group is clearly bifurcated for and against a resolution, 50+% is not an appropriate decision making tool. Please remove.

jandrieu (changes requested):

index.html
jandrieu: FWIW, Community process guidelines are at https://w3c-ccg.github.io/workitem-process/ which sadly, mentions the "CG's documented consensus process" but doesn't provide a link.

It does make it clear that in the adoption stage and final report stage that "principled objections" are sufficient to halt the adoption of a work item.

msporny: To clarify: is the resolution that you're objecting to this one: "RESOLUTION: One of the authorization mechanisms defined for the VC HTTP API ￼MUST be GNAP key-bound access tokens."

jandrieu: Yes. If there were a particular mechanism proposed we could evaluate it on its merit. However, we don't know what mechanism would be required. Making GNAP a MUST seems inappropriate at this stage.

That doesn't mean GNAP won't eventually be supported, just that we don't, today, have consensus that we should bind the VC-HTTP-API to unknown features outside our direct control.

OR13:
I'd prefer to see a new PR opened with only resolutions that have not been objected to implemented.

msporny:
> I'd prefer to see a new PR opened with only resolutions that have not been objected to implemented.

We know what happens when that PR gets opened... it just get blocked because it's missing one of the resolutions. :)

Let's keep this one open, suggesting alternatives to the text that are in proposals sent to the mailing list, and once the CCG Chairs weigh in, "The Decider(s)" will instruct the Editors to implement an option.

index.html
```diff
@@ -297,6 +310,43 @@ <h2>The VC HTTP API</h2>
...
+Convert the following group resolution into specification text:<br><br>
+<a href="https://w3c-ccg.github.io/meetings/2021-07-13-vchttpapi/#resolution-5">
+RESOLUTION</a>: One of the authorization mechanisms defined for the VC HTTP API
+MUST be GNAP key-bound access tokens.
```
msporny, line 347:
```suggestion
The group has decided to delay the discussion of implementing GNAP key-bound access tokens as one of the authorization mechanisms for the VC HTTP API until January 2022. The group might decide to take the topic up before that date if there is agreement to do so.
```

@jandrieu, @agropper, @jricher -- would any of you object to the text above?

OR13:
> it just get blocked because it's missing one of the resolutions. :)

then open a PR per resolution.

msporny:
> > it just get blocked because it's missing one of the resolutions. :)
> 
> then open a PR per resolution.

There you go, using logic again. :P 

Yes, I think you're right, it'll focus the PRs on single issues instead of lumping them all together like this PR does. This was supposed to be an "easy" PR because it only contained RESOLUTIONs. Clearly, there is opposition to one of the resolutions, so I'll do what you say and put each of them in a separate PR so we can focus the discussion in another PR.

jricher:
Honestly the biggest problem is that these resolutions are using normative language for the editors do to the spec, and this is not the normative language that will be in the spec itself. It's confusing people, thinking that the MUST here means "MUST use GNAP" where it really is "we'll define GNAP access tokens". The latter of these really is the text "you MAY use GNAP access tokens, and when using GNAP access tokens access is requested using the following `access` parameter:".

jandrieu:
> Honestly the biggest problem is that these resolutions are using normative language to describe what the editors are going to do to the spec at some point, and this is not the same as the normative language that will be in the spec itself. It's confusing people, thinking that the MUST here means "MUST use GNAP" where it really is "we'll define GNAP access tokens". The latter of these really is the text "you MAY use GNAP access tokens, and when using GNAP access tokens access is requested using the following `access` parameter:".

I agree with Justin. I've never seen a resolution from a meeting placed directly into spectext like this. It's another aspect of this decision that diverges from the W3C processes I'm familiar with. 

Normally (at least in the two W3C WGs I've been active in) there are multiple points of checks and balance, 
1. Resolution proposed (with show of support by voting or such) by a subgroup
2. Full group review to formally adopt the resolution
3. Resolution shared by email with a 7-14 day window for people to present concerns
4. Spectext drafted & proposed
5. PR debated
6. If consensus, PR is merged

We had skipped step 1 thanks to a bespoke process decision for just that single call,  using a majority rule instead of actual consensus. We also skipped #2. This matter has not been sent to the full CG for consideration. #3 was also never given a chance. When #4 occurred, I specifically asked if this is the right place in the process to object to what was happening. Instead of engagement and support for developing an actual consensus, that objection was only begrudgingly considered. On the last call I made the case that we have a process violation here, precisely because there is no consensus from the CG on this resolution. There wasn't even consensus from the subgroup. 

This debate is step #5. And given the lack of consensus, I believe it is premature to merge this text into the spec as it was not developed using normal W3C process and has not established legitimate consensus. Step #6 should not occur precisely because, regardless of a majority resolution, there is no consensus within the group to do so.

Given the sequence of events, I can't support any of the so called "resolutions" from this meeting as legitimate representations of the CG's consensus. However, this particular resolution was the one that stood out as a split decision, even among the relatively small group, which itself is a small minority of the CG. It is simply untenable for informal subgroup leadership to arbitrarily change the process purely to massage the record and manufacture the results they want. That's the exact opposite of the collaborative, consensus-driven standards making that is the hallmark of the W3C.

There remains plenty of time to add GNAP specific features. We don't even have a working group, yet. We have years ahead of us. The push to REQUIRE that the VC-HTTP-API adds a dependency on an undefined, unfinished specification at this stage just makes no sense, and sets us up for failure. 

We can still add GNAP-supporting features as the API is built out. The lack of consensus on this resolution should not be taken as a consensus to exclude GNAP from the new specification. 

However, if we bind the future work at this time we are playing political favorites and handicapping future development, especially when there is significant opposition to the requirement and a complete (or near complete) lack of support from actual implementers on the call. My recollection is that none of the implementers expressed any plans to support GNAP at this time. 

So... I'd like to find a way to return to a consensus-driven process rather than waste our time dealing with manufactured consensus on issues the group is, in fact, divided on.

msporny:
> Normally (at least in the two W3C WGs I've been active in) there are multiple points of checks and balance, 

A Work Item Group is not a W3C WG... the bar isn't as high on purpose; we're incubating specifications, not getting final sign off on a global standard. Applying a full blown W3C Process to a work item is overkill, so if that's the suggestion on how we operate a CCG work item group, I'm going to start objecting very loudly to that notion.

> 1. Resolution proposed (with show of support by voting or such) by a subgroup
> 2. Full group review to formally adopt the resolution
> 3. Resolution shared by email with a 7-14 day window for people to present concerns
> 4. Spectext drafted & proposed
> 5. PR debated
> 6. If consensus, PR is merged

> We had skipped step 1 thanks to a bespoke process decision for just that single call, using a majority rule instead of actual consensus. 

No, we did not skip this step. There was fair notice given about what we were going to do and months of debate that led up to that point.

The group was then notified that we'd be TEMPORARILY switching to majority voting for the "Authorization Proposals":

https://lists.w3.org/Archives/Public/public-credentials/2021Jul/0067.html

It was also explicitly mentioned on the call before we went into this mode (with no objections raised when we did). While it was stated that it was going to be simple majority (>50%), it seems as if some were confused about this at the time:

https://w3c-ccg.github.io/meetings/2021-07-13-vchttpapi/#40

There was a concern raised /after/ the votes came in and a suggestion to "not do this again":

https://w3c-ccg.github.io/meetings/2021-07-13-vchttpapi/#147

> We also skipped #2. This matter has not been sent to the full CG for consideration. 

No, we did not skip that step. The matter went out to the CG for consideration the second the minutes were published, which happened on the same day:

https://lists.w3.org/Archives/Public/public-credentials/2021Jul/0094.html

> #3 was also never given a chance. 

No, we didn't skip this step either. The resolution went out to the CCG mailing list on the same day and we have received ONE objection (from you).

> When #4 occurred, I specifically asked if this is the right place in the process to object to what was happening. Instead of engagement and support for developing an actual consensus, that objection was only begrudgingly considered. 

It was begrudgingly considered because there were multiple opportunities to object to the process when 1) the resolution was proposed on the mailing list, 2) when the approach was outlined on the call, and 3) when the voting happened. There were three opportunities to object before we went down this path and none of those opportunities were taken. Now, on the 4th step, you're objecting (which is fine)... but what's not fine is saying that we didn't follow the exact set of steps that you are noting that we should follow (which, again, I believe is overkill for an incubating work item, but we did it nonetheless because we were failing to get to consensus).

> On the last call I made the case that we have a process violation here, precisely because there is no consensus from the CG on this resolution. There wasn't even consensus from the subgroup.

Where does it say that we need consensus from the main CG to put in an issue marker stating that we're planning to specify how to use GNAP as an option?

> It is simply untenable for informal subgroup leadership to arbitrarily change the process purely to massage the record and manufacture the results they want. 

@jandrieu, please reconsider the words that you are using. This comes across as an insinuation that the subgroup leadership, which includes me (and perhaps the Editors), are acting in bad faith.

To be crystal clear -- none of the subgroup leadership (I'm including the Editors an myself in that group) seems to be supportive of the GNAP resolution. I'm certainly against it personally and as a representative of Digital Bazaar. That said, I don't think it does any harm and if people want GNAP, they'll implement GNAP (or we won't get enough implementations and we'll take it out)... and we were clear about the rules we were making when we did the resolution.

> So... I'd like to find a way to return to a consensus-driven process rather than waste our time dealing with manufactured consensus on issues the group is, in fact, divided on.

I disagree strongly with this characterization. No one said these resolutions were consensus w/ unamity. You could argue that most of them were consensus w/ sustained objections... and there is one, which is the one that we're talking about, which has turned out to be controversial.

> Given the sequence of events, I can't support any of the so called "resolutions" from this meeting as legitimate representations of the CG's consensus.

Are you planning to object to all of the resolutions made on the call?

msporny:
> It's confusing people, thinking that the MUST here means "MUST use GNAP" where it really is "we'll define GNAP access tokens". The latter of these really is the text "you MAY use GNAP access tokens, and when using GNAP access tokens access is requested using the following `access` parameter:".

I don't think the majority of the people in the group are confused on this point. We can ask them directly when we decided to talk about this matter again in the group.

jandrieu:
> > Normally (at least in the two W3C WGs I've been active in) there are multiple points of checks and balance,
> 
> A CCG Work Item Group is not a W3C WG... the bar isn't as high on purpose; we're incubating specifications, not getting final sign off on a global standard. Applying a full blown W3C Process to a work item is overkill, so if that's the suggestion on how we operate a CCG work item group, I'm going to start objecting very loudly to that notion.

Please do. The current charter requires that work items are approved without any principled objections. I have a principled objection and, despite your clout in the group, it's inappropriate for you to adjust the process just so you can advance a work topic despite a lack of consensus.

> > 1. Resolution proposed (with show of support by voting or such) by a subgroup
> > 2. Full group review to formally adopt the resolution
> > 3. Resolution shared by email with a 7-14 day window for people to present concerns
> > 4. Spectext drafted & proposed
> > 5. PR debated
> > 6. If consensus, PR is merged
> 
> > We had skipped step 1 thanks to a bespoke process decision for just that single call, using a majority rule instead of actual consensus.
> 
> No, we did not skip this step. There was fair notice given about what we were going to do and months of debate that led up to that point.

I disagree that there was fair notice. Was there a call for a proposal to the CCG mailing list? Was there a discussion on a CCG call? Was there a 7-day+ review of the suggestion? Was there a vote on this specific decision? To my knowledge there was just notification and not engagement, as evidenced by the number of people who were confused.

The process aberration was proposed and adopted without significant consideration. Clearly, many who nodded along did not understand the impact of that decision. So, yes, I understand that you adjusted the group's policy on an ad-hoc basis. I just don't see how that was a legitimate process by any documented or historical standard.

> The group was then notified that we'd be TEMPORARILY switching to majority voting for the "Authorization Proposals":
> 
> https://lists.w3.org/Archives/Public/public-credentials/2021Jul/0067.html

Right. You *told* the group what the policy was going to be. And the fact that it is "TEMPORARY" only underscores the nature of the manipulation: it was clear that no one would support this on an ongoing basis. It was purely to enable pushing past a particularly contentious point.

Where was the discussion of that policy shift?

Why do you feel it was appropriate for you to make this call, when you aren't even willing to accept the title of chair?

I'm arguing that it was not appropriate. Full stop. That the decision to use an unusual process was the process violation.

Can you cite anywhere in the CCG or W3C process documents that "moderators" of calls for subgroups without chairs are empowered to adopt arbitrary non-consensus rules for resolutions?

This is the crux of the problem.

> It was also explicitly mentioned on the call before we went into this mode (with no objections raised when we did). While it was stated that it was going to be simple majority in the worst case (>50%), it seems as if some were confused about this at the time:

> > We also skipped #2. This matter has not been sent to the full CG for consideration.
> 
> No, we did not skip that step. The matter went out to the CG for consideration the second the minutes were published, which happened on the same day:
> 
> https://lists.w3.org/Archives/Public/public-credentials/2021Jul/0094.html

Publishing the minutes is nowhere near the level of notification necessary for informed consent, IMO. For CCG decisions during my tenure as co-chair, there were explicit emails sent to the group with a request for opposition and that pattern seemed in play with the VCWG and DIDWG, so why do you think notification is sufficient? No, we didn't do it with every detail every group resolved. But we did do it for every decision that impacted process.

Arguing that publishing the minutes is the same as asking for buy-in is simply asserting that the subgroup has no need to escalate any of its process to the group. That is a recipe for rogue groups.

> > #3 was also never given a chance.
> 
> No, we didn't skip this step either. The resolution went out to the CCG mailing list on the same day and we have received ONE objection (from you).

Process objections don't need consensus. They need to be investigated and resolved, even if its just one party complaining.

> > When #4 occurred, I specifically asked if this is the right place in the process to object to what was happening. Instead of engagement and support for developing an actual consensus, that objection was only begrudgingly considered.
> 
> It was begrudgingly considered because there were multiple opportunities to object to the process when 1) the resolution was proposed on the mailing list, 2) when the approach was outlined on the call, and 3) when the voting happened. There were three opportunities to object before we went down this path and none of those opportunities were taken. Now, on the 4th step, you're objecting (which is fine)... but what's not fine is saying that we didn't follow the exact set of steps that you are noting that we should follow (which, again, I believe is overkill for an incubating work item, but we did it nonetheless because we were failing to get to consensus).

Just because I did not perceive the process violation in its early steps does not preclude me from raising the issue AND HAVING IT CONSIDERED. Are you saying that bank robbers can't be chased down after they've left the building? That's ridiculous. What you claim are "opportunities to object" I would characterize as "attempts to notify the community of substantial process changes did not have the effect of actually engendering informed consent."

As such, my objection should have been welcomed as legitimate.

Obviously, we have a different read on it. You seem to think that simply publishing minutes meets your obligations to the larger CCG. 

I don't. It's too easy to bury decisions in that fashion. 

Adjustments to process should have been giving a much wider vetting, otherwise we end up with exactly this scenario: process tweaking that corrupts the ability of the group to morally act as a coherent unit.

> > On the last call I made the case that we have a process violation here, precisely because there is no consensus from the CG on this resolution. There wasn't even consensus from the subgroup.
> 
> Where does it say that we need consensus from the main CG to put in an issue marker stating that we're planning to specify how to use GNAP as an option?

It says the process for acceptance of work items is principled objections. As one of the chairs who helped draft that language, I regret that we did not add that any subgroups (or task forces) would operate under the same rules. It was understood, but shame on us for not putting that in writing.

The fact that you think a subgroup "moderator" is empowered to arbitrarily change process by mere notification sent to overworked and unpaid community members is the problem here.

> > It is simply untenable for informal subgroup leadership to arbitrarily change the process purely to massage the record and manufacture the results they want.
> 
> @jandrieu, please reconsider the words that you are using. This comes across as an insinuation that the subgroup leadership, which includes me (and perhaps the Editors), are acting in bad faith.

I'm considering my words very carefully, thank you. 

I believe your heart is in the right place. It's the execution that is the problem.

It is admitted and well documented that the point of changing the process was specifically to bypass the challenges you were having getting consensus. In other words, the entire point of your actions were to enable resolutions to be passed that lacked consensus.

I realize you just want to get the work done, but it isn't appropriate to do it this way.

This is not the first time that the CCG has had issues with you using your informal authority outside consensus processes. I was co-chair when we had to resolve similar attempts to "just get the work done". Haste does not justify bypassing process.
 
> To be crystal clear -- none of the subgroup leadership (I'm including the Editors an myself in that group) seems to be supportive of the GNAP resolution. I'm certainly against it personally and as a representative of Digital Bazaar. That said, I don't think it does any harm and if people want GNAP, they'll implement GNAP (or we won't get enough implementations and we'll take it out)... and we were clear about the temporary voting rules we were adopting in advance of putting the resolution out to the group for a vote.

I don't think it was clear at all, certainly not clear enough for informed consent. The implications of the bespoke process shift was NOT made clear. It is this kind of procedural legerdemain that is not okay.
 
> > So... I'd like to find a way to return to a consensus-driven process rather than waste our time dealing with manufactured consensus on issues the group is, in fact, divided on.
> 
> I disagree strongly with this characterization. No one said these resolutions were consensus w/ unamity. You could argue that most of them were consensus w/ sustained objections... and there is one, which is the one that we're talking about, which has turned out to be controversial.

This is a false interpretation. No one is talking about unanimity. That's a red herring that emerged as people expressed frustration at the pace of spec development during the DHS interop work. Consensus, especially within the W3C, is not unanimity. You know that. And you know I'm not arguing for that.

I have, in fact, been satisfied with the handling of loud minorities in VCWG and DIDWG in the past. There were numerous times where strong voices, singularly or occasionally with two people, did NOT prevail in their efforts to impact the specification. I myself had it explained to me, by you personally, in no uncertain terms, that if I was unable to bring more voices in support of my position (on the "type" property), that I would NOT carry the day. And that is totally appropriate. A singular voice--or even two--should not be able to hobble the entire effort. That made sense. If I couldn't get others to see things my way, that's my failing.

But that's not what happened. We had a split decision that somehow became a PR that, for some reason, MUST be adopted despite their being a clear failing of consensus on the spectext. Why isn't the PR an appropriate place for someone to object, even on purely technical grounds?
 
> > Given the sequence of events, I can't support any of the so called "resolutions" from this meeting as legitimate representations of the CG's consensus.
> 
> Are you planning to object to all of the resolutions made on the call?

No, I'm not. Which why I made that statement (you might want to reread the full paragraph). IMO, the entire legitimacy of any resolution on that call is questionable. However, the only one I'm objecting to was the one that seemed to divide the group. Most, if not all, of the others actually had strong consensus.

This particular question had ten people vote. Of those, only 5 voted for the resolution. 5 other individuals either voted zero or in the negative (only one expressed a "0" vote). The mean vote was a mere 0.155 in favor. On a scale from -1 to 1. That's not consensus.

If those were straight-line votes, yes, it would be a plurality of the votes, but possibly not even a majority (depending on what a zero vote really means). That practice has been used in the past as a way to get a quick snapshot about whether or not there is consensus, but this is the first time in my experience in this community that the differential of those votes was used to assert a resolution rather than simply acknowledge a lack of consensus. I believe lots of us were surprised by this strange electoral math.

What I really don't understand, Manu, is why you are so adamant about this particular spec text. As Justin pointed out, it's odd and confusing to put guidance like this *in spec text*. In my experience resolutions are noted, PRs are written, and further debate continues. And when a PR has consensus, it gets merged in.

This PR does not have consensus.

In this case, you have not only repeatedly shut down my attempts to voice my objections at the point where I was informed (by you) was the appropriate place, but several additional voices have also expressed their dissatisfaction with the process and, as you yourself mention, the resolution itself doesn't even have *your* support.

So, why are you trying to push a PR that doesn't have consensus?

REGARDLESS of any resolutions, that alone should be sufficient for any member of the group to oppose the inclusion of the language in actual spectext.

msporny:
> What I really don't understand, Manu, is why you are so adamant about this particular spec text. 
> So, why are you trying to push a PR that doesn't have consensus?

There is a tremendous amount of miscommunication in the last two messages and the latest response is demonstrating that the miscommunication gap is getting wider (not more narrow). 

This medium isn't helping us have a productive discussion.

At this point, I'm going to suggest that we get on a call (public and recorded, but not the VC HTTP API call, so that others might benefit from the discussion if they choose to) and go through your concerns. This may be a call that would benefit @vsnt, @mprorock, and @wyc as they deliberate what should be done with the CCG process.

mavarley:
> > > Given the sequence of events, I can't support any of the so called "resolutions" from this meeting as legitimate representations of the CG's consensus.

> > Are you planning to object to all of the resolutions made on the call?

> No, I'm not. Which why I made that statement (you might want to reread the full paragraph). IMO, the entire legitimacy of any resolution on that call is questionable. However, the only one I'm objecting to was the one that seemed to divide the group. Most, if not all, of the others actually had strong consensus.

This is where I get confused; if there is a principle objection to the nature/process of how these resolutions were reached (which I can understand, given the above) then it seems the corrective course of action is (if the objection is actionable) - to drop the entire set of resolutions from the call, and try again with the accepted process. If most of the resolutions already have consensus, they should go through without an issue; any resolutions which do not have consensus can be addressed.

This may be expensive and frustrating (time and energy), and if a resolution that was accepted (in good faith, even if the process was incorrect) and now finds itself on the outside looking in - _let's work on the resolution to bring it back in_. To be clear, for those GNAP supporters who feel their being picked on that their resolution is the only one being removed: let's work on the resolution to get it to consensus. 

I believe from Justin's @jricher comment above

>  The latter of these really is the text "you MAY use GNAP access tokens, and when using GNAP access tokens access is requested using the following access parameter:".

Is a path to consensus - but let's work the 'correct'/accepted process, wherever it lands.

Again, if the objection on the process is actionable, I believe we need to reset - but if most of the resolutions already have consensus, they should pass any process we come up with. I believe there is also a requirement by this group to help work on any resolution that finds itself "on the outside" to work on finding consensus on those items, (and they do not get treated as dead topics).

msporny:
> At this point, I'm going to suggest that we get on a call (public and recorded, but not the VC HTTP API call, so that others might benefit from the discussion if they choose to) and go through your concerns. This may be a call that would benefit @vsnt, @mprorock, and @wyc as they deliberate what should be done with the CCG process.

@jandrieu and I talked, we're good, no further action needed by the CCG Chairs.

msporny:
This PR is being rapidly overtaken by events:

https://lists.w3.org/Archives/Public/public-credentials/2021Aug/0253.html

I expect that we'll close the whole thing and raise individual PRs (@mavarley already has these prepared in branches) for the resolutions that survive.

jandrieu:
> @jandrieu and I talked, we're good, no further action needed by the CCG Chairs.

Yep. I'm good with the process proposed by the chairs. It was unfortunate to bump into this wrinkle/gap in the CCG processes, but I think we found a reasonable path through. Many thanks to @mprorock @heathervescent and @wyc for their prompt engagement and quick resolution.

We still need to model the right cultural attitude *and* boundaries for consensus, but I feel this is in the right direction. My expectations for W3C work has always been that a lone voice is not a suitable standard for non-consensus, but that nevertheless, every voice has its opportunity to participate in the discussion and to propose alternatives that would address their concerns. It is up to leadership--whether chairs or editors--to make the judgment call about whether or not their is "enough" consensus to advance from a topic, whether adopting or deferring.

I especially appreciate the chairs' clarification that even in subgroups and task forces the standard is consensus rather than a simple majority.

How we manage that consensus still requires attention, but I'm comfortable on that footing thanks to the decades of process and culture behind the W3C.

OR13:
I suggest closing this, since its been overtaken by other PRs and conversations.

msporny:
We discussed this on the 2021-09-21 call and there were multiple people supporting closing this PR with no objections. This PR will be closed shortly.

PR #226, opened 2021-08-24
Security Framework

jricher:
***
<a href="https://pr-preview.s3.amazonaws.com/jricher/vc-http-api/pull/226.html" title="Last updated on Aug 24, 2021, 8:16 PM UTC (2ba8a71)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-http-api/226/41dbd3e...jricher:2ba8a71.html" title="Last updated on Aug 24, 2021, 8:16 PM UTC (2ba8a71)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -296,7 +296,111 @@ A <dfn>conforming VC HTTP API server</dfn> is ...
     <section>
       <h3>Authorization</h3>
       <p>
+As an API built on top of HTTP, the VC HTTP API MAY be protected by any security technology
+that is applicable to such HTTP APIs. It is RECOMMENDED that deployments use standard delegation
+protocols to allow for interoperable access to the functions of the API without divulging end-user
+credentials or details unnecessarily to different parties. It is NOT RECOMMENDED that deployments
+use direct authentication technologies such as HTTP Basic Authentication.
       </p>
+      <p>
+This specification defines interoperable bindings to several standard delegation and authorization
+protocols. A conformant implementation deployed with one of these technologies MUST use these
+profiles and MAY further profile the connections as necessary. Each VC HTTP API endpoint defines
+a an authorization <code>type</code>, <code>actions</code>, <code>datatypes</code>, and other 
+security values relevant to the different actions that can be taken on an endpoint.
+These security values are combined and encoded in different ways for the associated security
+protocol, allowing a client to request and a policy to specify fine-grained access control
+for the VC HTTP API endpoints under protection.
+      </p>
+      <section>
+        <h4>OAuth 2.0</h4>
+        <p>
+When calling a VC HTTP API using OAuth 2.0, the request MUST be made using 
+<a href="https://tools.ietf.org/html/draft-ietf-oauth-rar">OAuth 2.0 Rich Authorization Requests</a>
+to describe the rights requested for the given API. The security parameters relevant for
+the rights requested for the endpoints in question are encoded as separate objects within
+the authorization details array.
+        </p>
+        <p>
+For example, a client requesting access to the issuing endpoint hosted at
+<code>https://example.com/issue</code> for the purposes of issuing credentials would
+combine the security parameters defined in the issuing endpoint as follows:
+        </p>
+        <p>
+          <code><pre>
+[
+  {
+    "type": "https://www.w3c.org/TR/vc-http-api",
+    "actions": [ "issue" ],
+    "locations": [ "https://example.com/issue" ],
+    "datatypes": [ "credential" ]
+  }
+]
+          </pre></code>
+        </p>
+        <p>
+If this is the only request, it is encoded into the OAuth 2.0 Rich Authorization Request as follows:
+        </p>
+        <p>
+          <code>
+authorization_details=%5B%7B%22type%22%3A%22https%3A%2F%2Fwww.w3c.org%2FTR%2Fvc-http-api%22%2C%22actions%22%3A%5B%22issue%22%5D%2C%22locations%22%3A%5B%22https%3A%2F%2Fexample.com%2Fissue%22%5D%2C%22datatypes%22%3A%5B%22credential%22%5D%7D%5D&client_id=...
+          </code>
+        </p>
+        <p>
+The OAuth 2.0 <code>scope</code> and <code>resource</code> parameters are not defined for
+the VC HTTP API, though these parameters MAY be used to combine requests for other APIs with
+the request for access to the VC HTTP API. Requests for different actions and endpoints MAY
+be combined into the authorization details request list as different objects.
+        </p>
+        <p>
+The access tokens granted from this process MAY be OAuth 2.0 Bearer tokens or any other
+valid OAuth 2.0 token type.
+        </p>
+      </section>
+      <section>
+        <h4>Grant Negotiation and Authorization Protocol (GNAP)</h4>
+        <p>
+When calling a VC HTTP API using <a href="https://www.ietf.org/archive/id/draft-ietf-gnap-core-protocol-06.html">GNAP</a>
+, the request MUST be made using the GNAP resource access rights data structures
+to describe the rights requested for the given API. The security parameters relevant for
+the rights requested for the endpoints in question are encoded as separate objects within
+the <code>access</code> array.
+        </p>
+        <p>
+For example, a client requesting access to the issuing endpoint hosted at
+<code>https://example.com/issue</code> for the purposes of issuing credentials would
+combine the security parameters defined in the issuing endpoint in the following
+grant request structure:
+        </p>
+        <p>
+          <code><pre>
+{
+  "access_token": {
+    "access": [
+      {
+        "type": "https://www.w3c.org/TR/vc-http-api",
+        "actions": [ "issue" ],
+        "locations": [ "https://example.com/issue" ],
+        "datatypes": [ "credential" ]
+      }
+    ]
+  },
+  "client": ...
+}
+          </pre></code>
+        </p>
+        <p>
+This example shows a single request for a single access token, but the same structures
+apply to multiple requests or multiple access tokens.
+        </p>
+        <p>
+GNAP resource reference values are not defined for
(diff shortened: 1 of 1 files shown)


jricher:
Note that this doesn't add additional fields to the endpoint definitions because of limits in ReDoc not allowing display or definition of extensions. I hacked ReDoc locally to support additional security properties, and it looks like this:

PastedGraphic-2

PastedGraphic-3

From a definition that looks like:

~~~
  securitySchemes:
    vcHttpApiAuth:
      description: Control access to the API using OAuth 2.0 RAR or GNAP
      type: rar
      rarType: 'https://www.w3.org/TR/vc-http-api/issue'
      locations:
        - '<The absolute URL of the endpoint being called>'
      datatypes:
        - credential
      actions:
        - issue
~~~

index.html
```diff
@@ -296,7 +296,111 @@ <h2>The VC HTTP API</h2>
...
+a an authorization <code>type</code>, <code>actions</code>, <code>datatypes</code>, and other 
+security values relevant to the different actions that can be taken on an endpoint.
+These security values are combined and encoded in different ways for the associated security
+protocol, allowing a client to request and a policy to specify fine-grained access control
+for the VC HTTP API endpoints under protection.
+      </p>
+      <section>
+        <h4>OAuth 2.0</h4>
+        <p>
+When calling a VC HTTP API using OAuth 2.0, the request MUST be made using
```
OR13, line 318:
```suggestion
When calling a VC HTTP API using OAuth 2.0, the request MAY be made using 
```

OR13 (changes requested):
Overall, this PR is good.

My only concerns are:

1. MUST use RAR -> MAY use RAR with OAuth2.0
2. No examples of traditional scope based OAuth, such as those supported by Auth0 / Okta.

1 needs to be addressed before merging, but 2 could be handled in a subsequent PR, if there is agreement that we providing such examples is on scope...

here is an example of what I mean:

https://github.com/OR13/GNARLY/blob/main/src/spec/v1.0.0/openapi.yaml#L48

```
components:
  securitySchemes:
    OAuth2:
      type: oauth2
      flows:
        implicit:
          authorizationUrl: https://or13.us.auth0.com/authorize
          tokenUrl: https://or13.us.auth0.com/oauth/token
          redirectUri: 'http://localhost:8080/api/docs/static/oauth2-redirect.html'
          scopes:
            'create:credentials': Grants permission to create credentials
            'derive:credentials': Grants permission to derive credentials
            'create:presentations': Grants permission to create presentations
            'verify:presentations': Grants permission to verify presentations
            'exchange:presentations': Grants permission to exchange presentations
```

mprorock:
> 2\. No examples of traditional scope based OAuth, such as those supported by Auth0 / Okta.

big +1 to getting scopes in there - this is the industry default right now and does not appear to be changing any time soon, so we should provide appropriate guidance on how to implement

mprorock (changes requested):
As noted by @OR13 :
1) MUST use RAR -> MAY use RAR with OAuth2.0
2) No examples of traditional scope based OAuth as is the industry standard

then I think this would be good

index.html
```diff
@@ -296,7 +296,111 @@ <h2>The VC HTTP API</h2>
...
+        <p>
+The access tokens granted from this process MAY be OAuth 2.0 Bearer tokens or any other
+valid OAuth 2.0 token type.
+        </p>
+      </section>
+      <section>
+        <h4>Grant Negotiation and Authorization Protocol (GNAP)</h4>
+        <p>
+When calling a VC HTTP API using <a href="https://www.ietf.org/archive/id/draft-ietf-gnap-core-protocol-06.html">GNAP</a>
+, the request MUST be made using the GNAP resource access rights data structures
```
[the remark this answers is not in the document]
jandrieu: Since the GNAP specification is not final, IMO, we should not reference it in this way.

If GNAP has identified a specific mechanism, however, we should just include it here directly, answering concretely "what is a GNAP resource access rights data structure".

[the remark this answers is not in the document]
TallTed: @jandrieu - Perhaps better to add your comment distinctly from my change request, which is more about markup & appearance.

agropper:
I feel strongly that Justin's PR language should stay as MUST.

TallTed:
`SHOULD` might be better than `MAY` for general interop ... 

and at this stage of everything, `SHOULD take { RAR, GNap, other } into consideration before going down another road` might be better still, especially if such consideration leads to actionable feedback on *this* doc.

index.html
```diff
@@ -296,7 +296,111 @@ <h2>The VC HTTP API</h2>
...
+]
+          </pre></code>
+        </p>
+        <p>
+If this is the only request, it is encoded into the OAuth 2.0 Rich Authorization Request as follows:
+        </p>
+        <p>
+          <code>
+authorization_details=%5B%7B%22type%22%3A%22https%3A%2F%2Fwww.w3c.org%2FTR%2Fvc-http-api%22%2C%22actions%22%3A%5B%22issue%22%5D%2C%22locations%22%3A%5B%22https%3A%2F%2Fexample.com%2Fissue%22%5D%2C%22datatypes%22%3A%5B%22credential%22%5D%7D%5D&client_id=...
+          </code>
```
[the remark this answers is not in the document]
msporny: Yes, please. I wouldn't characterize it as a nitpick, I think there is a big question on whether or not we should define the scopes explicitly. If we're going to the trouble of laying out RAR, then I don't see why we wouldn't also lay out all the scope details.

We need separate sections on scopes, at the very least.

index.html
```diff
@@ -296,7 +296,111 @@ <h2>The VC HTTP API</h2>
...
+        </p>
+        <p>
+The OAuth 2.0 <code>scope</code> and <code>resource</code> parameters are not defined for
+the VC HTTP API, though these parameters MAY be used to combine requests for other APIs with
+the request for access to the VC HTTP API. Requests for different actions and endpoints MAY
+be combined into the authorization details request list as different objects.
+        </p>
+        <p>
+The access tokens granted from this process MAY be OAuth 2.0 Bearer tokens or any other
+valid OAuth 2.0 token type.
```
msporny, line 357: I hesitate to say we should just leave this open ended. What other valid OAuth 2.0 token types that we expect to be used with the API are there?

msporny (changes requested):
I appreciate that this PR is attempting to show a full picture of what including RAR+GNAP in the spec could look like. It's a lot to swallow all at once, especially since we've seen signalling from the implementers that they have no intention of implementing GNAP or RAR any time soon. I'd feel better about this PR if we had more implementer support, because until we do, it's (well crafted) fiction.

If it looks like the group wants to include all of this, we might want to break it up into pieces. I have attempted to do that with the portions of this PR (with significant rewording) in PRs #228, #229, #230, and #231. I did those PRs in a way that we could just add the RAR and GNAP sections in if the group wanted to merge portions of this PR.

index.html
OR13: I require this change to approve the PR.

index.html
jricher: Just for immediate examples there are MTLS bound tokens, DPoP bound tokens, PoP bound tokens ...

jricher:
I disagree strongly to adding in scope descriptions.

index.html
```diff
@@ -296,7 +296,111 @@ <h2>The VC HTTP API</h2>
     <section>
       <h3>Authorization</h3>
       <p>
+As an API built on top of HTTP, the VC HTTP API MAY be protected by any security technology
+that is applicable to such HTTP APIs. It is RECOMMENDED that deployments use standard delegation
+protocols to allow for interoperable access to the functions of the API without divulging end-user
+credentials or details unnecessarily to different parties. It is NOT RECOMMENDED that deployments
+use direct authentication technologies such as HTTP Basic Authentication.
```
TallTed, line 303:
```suggestion
protocols to allow for interoperable access to the functions of the API without unnecessarily
divulging end-user credentials or other details to different parties. It is NOT RECOMMENDED
that deployments use direct authentication technologies such as HTTP Basic Authentication.
```

index.html
```diff
@@ -296,7 +296,111 @@ <h2>The VC HTTP API</h2>
...
+for the VC HTTP API endpoints under protection.
+      </p>
+      <section>
+        <h4>OAuth 2.0</h4>
+        <p>
+When calling a VC HTTP API using OAuth 2.0, the request MUST be made using 
+<a href="https://tools.ietf.org/html/draft-ietf-oauth-rar">OAuth 2.0 Rich Authorization Requests</a>
+to describe the rights requested for the given API. The security parameters relevant for
+the rights requested for the endpoints in question are encoded as separate objects within
+the authorization details array.
```
TallTed, line 322: I'm not sure I've used the right markup for the new paragraph here.... Also, I don't know whether it would be better to point to DID Registries, VC Registries, or somewhere else.
```suggestion
When making a VC API call using OAuth 2.0, the request SHOULD describe the rights
requested using <a href="https://tools.ietf.org/html/draft-ietf-oauth-rar">OAuth 2.0
Rich Authorization Requests</a>. The security parameters relevant for the rights
requested for the endpoints in question are encoded as separate objects within the
authorization details array. 
        </p>
        <p class="advisory">
Details regarding use of other authorization mechanisms may be specified as future
extensions to the VC API, managed through the _____ Registries.
```

agropper:
I'm commenting based on the assumption that GNAP and OAuth2 will be co-equal SHOULDs for VC API authorization, meaning that API conformance testing will require both protocols to pass. (I know this is not settled but I'm hoping to keep this issue clear and moving forward.)

In the case of co-equal SHOULDs we then have the choice of specifying RAR in both cases or not. From my position, limiting RAR to only the GNAP authorization path would save no work for the implementers but could provide a measure of backward compatibility in some deployments at some cost in interoperability.

msporny:
This PR has not seen any movement forward in over a year. Closing.

PR #231, opened 2021-09-12
Add section on OAuth 2.0 and bearer tokens.

msporny:
This PR specifies that OAuth 2.0 and OAuth 2.0 bearer tokens can be used as an authorization protocol and access token format.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-http-api/pull/231.html" title="Last updated on Sep 26, 2021, 12:03 AM UTC (901d266)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-http-api/231/ed6a27e...901d266.html" title="Last updated on Sep 26, 2021, 12:03 AM UTC (901d266)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -315,7 +315,7 @@ or legacy authorization technologies.
         <p>
 If Open Authorization is utilized for authorization, version 2.0 of the
 protocol (OAuth 2.0) [[RFC6749]] MUST be used. The access tokens utilized by
-clients MAY be OAuth 2.0 Bearer Tokens [[RFC6750]]  or any other valid OAuth 2.0
+clients MAY be OAuth 2.0 Bearer Tokens [[RFC6750]] or any other valid OAuth 2.0
 token type.
         </p>
       </section>
--- a/index.html
+++ b/index.html
@@ -310,6 +310,16 @@ technologies can be used. Implementers are cautioned against using non-standard
 or legacy authorization technologies.
       </p>
 
+      <section>
+        <h4>Open Authorization (OAuth)</h4>
+        <p>
+If Open Authorization is utilized for authorization, version 2.0 of the
+protocol (OAuth 2.0) [[RFC6749]] MUST be used. The access tokens utilized by
+clients MAY be OAuth 2.0 Bearer Tokens [[RFC6750]]  or any other valid OAuth 2.0
+token type.
+        </p>
+      </section>
+
     </section>
 
     <section>
--- a/index.html
+++ b/index.html
@@ -311,12 +311,12 @@ or legacy authorization technologies.
       </p>
 
       <section>
-        <h4>Open Authorization (OAuth)</h4>
+        <h4>OAuth 2.0</h4>
         <p>
-If Open Authorization is utilized for authorization, version 2.0 of the
-protocol (OAuth 2.0) [[RFC6749]] MUST be used. The access tokens utilized by
-clients MAY be OAuth 2.0 Bearer Tokens [[RFC6750]] or any other valid OAuth 2.0
-token type.
+If the OAuth 2.0 Authorization Framework [[RFC6749]] is utilized for authorization,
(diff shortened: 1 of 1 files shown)


OR13 (approved):

mprorock (approved):

mkhraisha (approved):

index.html
```diff
@@ -310,6 +310,16 @@ <h3>Authorization</h3>
 or legacy authorization technologies.
       </p>
 
+      <section>
+        <h4>Open Authorization (OAuth)</h4>
+        <p>
+If Open Authorization is utilized for authorization, version 2.0 of the
+protocol (OAuth 2.0) [[RFC6749]] MUST be used. The access tokens utilized by
+clients MAY be OAuth 2.0 Bearer Tokens [[RFC6750]]  or any other valid OAuth 2.0
```
TallTed, line 318:
```suggestion
clients MAY be OAuth 2.0 Bearer Tokens [[RFC6750]] or any other valid OAuth 2.0
```

TallTed:
I wonder if it's worth leaving the door open to later versions of OAuth which may exist before these docs are updated?

index.html
```diff
@@ -310,6 +310,16 @@ <h3>Authorization</h3>
 or legacy authorization technologies.
       </p>
 
+      <section>
+        <h4>Open Authorization (OAuth)</h4>
+        <p>
+If Open Authorization is utilized for authorization, version 2.0 of the
+protocol (OAuth 2.0) [[RFC6749]] MUST be used. The access tokens utilized by
+clients MAY be OAuth 2.0 Bearer Tokens [[RFC6750]] or any other valid OAuth 2.0
+token type.
```
jricher, line 319:
```suggestion
        <h4>OAuth 2.0</h4>
        <p>
If the OAuth 2.0 Authorization Framework [[RFC6749]] is utilized for authorization,
the access tokens utilized by clients MAY be OAuth 2.0 Bearer Tokens [[RFC6750]]
or any other valid OAuth 2.0 token type. Any valid OAuth 2.0 grant type MAY be used
to request the access tokens.
```

jricher (changes requested):
Existing language does not make sense for requirements, suggested changes.

index.html
```diff
@@ -315,7 +315,7 @@ <h4>Open Authorization (OAuth)</h4>
         <p>
 If Open Authorization is utilized for authorization, version 2.0 of the
```
TallTed, line 316:
```suggestion
If Open Authorization (OAuth) is used for authorization, version 2.0 of the
```

index.html
msporny: > It makes no sense to explicitly disallow OAuth 1.0 if someone's crazy enough to use it.

The reason not to use it is embedded in your response. "If someone's crazy enough to use it" -- we don't want crazy usages of the API, thus, just ban OAuth 1.0. There is no one that we know of in their right mind that would use OAuth 1.0, right?

> The name of the protocol is not "Open Authorization".

You should tell that to the Wikipedia page:

https://en.wikipedia.org/wiki/OAuth

... and lots of articles on the Web:

https://www.google.com/search?q=%22open+authorization%22

That said, I'm fine w/ the change to not spell it out, whatever it means.

I'm concerned that we don't have any normative statements from this language that will make their way into a test suite. @OR13, @mprorock, @jricher, thoughts on that?

index.html
msporny: Using @jricher's suggestion over yours @TallTed.

msporny:
I have concerns that the current language around OAuth 2.0 will result in zero tests being written in the test suite regarding OAuth 2.0 authorization. I have raised https://github.com/w3c-ccg/vc-http-api/issues/234 to track that concern.

Change requests processed, merging.

index.html
TallTed: OK. I don't have a RESOLVE button, or I'd click it...

PR #233, opened 2021-09-20
Add ReSpec auto-generator from OAS files.

msporny:
This PR adds an ability to ReSpec to generate sections of the specification directly from the OAS files. This ensures that the root of truth continues to be the OAS files, but that those are also translated into human-readable normative ReSpec descriptions.

This is the first iteration of this experimental feature. There is still a lot of formatting that needs to be done with the JSON Schema sections to ReSpec text, but that is fairly straight-forward to do and will just take a bit of time to get formatted into the shape that we want.

You can preview what all of this looks like here: https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-http-api/pull/233.html#the-vc-http-api

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-http-api/pull/233.html" title="Last updated on Sep 26, 2021, 12:20 AM UTC (597657d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-http-api/233/ed6a27e...597657d.html" title="Last updated on Sep 26, 2021, 12:20 AM UTC (597657d)">Diff</a>

### DIFF
--- a/common.js
+++ b/common.js
@@ -314,3 +314,69 @@ function unComment(doc, content) {
     .replace(/-\s*- >/g, '-->')
     .replace(/-\s*-\s*&gt;/g, '--&gt;');
 }
+
+/********************* OAS Experiment *************************/
+function getEndpoint({apis, path}) {
+  let endpoint = {
+    post: {
+      summary: 'Error: API Endpoint not defined - ' + path
+    }
+  };
+
+  for(const api of apis) {
+    if(api.paths[path] !== undefined) {
+      endpoint = api.paths[path];
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


OR13 (approved):
very cool, and very useful.

We've been doing something very similar in https://github.com/w3c-ccg/traceability-vocab, and recently resolved to update to using OAS instead of JSON Schema, because we wanted to avoid potential issues.... this is another good reason for that switch.

OR13:
open for 9 days without objection, merging.

Issue #236, opened 2021-10-26
Beware of broken revocation list urls


OR13:
https://w3c-ccg.github.io/vc-http-api/fixtures/revocationList.json -> causes all credentials issued against this endpoint to fail to verify.

All test data needs to be regenerated to use: https://w3c-ccg.github.io/vc-api/fixtures/revocationList.json

although probably it should use a w3id.org base instead to fix this properly.

OR13:
@clehner fixed this, assigning him and closing the issue.

PR #242, opened 2021-10-29
Regenerate revocation list credential for new URL

clehner:
Address #236

A corresponding PR to update the verifiable credentials using the revocation list is here: https://github.com/w3c-ccg/vc-api-test-suite/pull/10.

This uses a new key I generated locally, as I have not access to the existing key. Is this a correct approach?

### DIFF
--- a/fixtures/revocationList.json
+++ b/fixtures/revocationList.json
@@ -3,13 +3,13 @@
       "https://www.w3.org/2018/credentials/v1",
       "https://w3id.org/vc-revocation-list-2020/v1"
     ],
-    "id": "https://w3c-ccg.github.io/vc-http-api/fixtures/revocationList.json",
+    "id": "https://w3c-ccg.github.io/vc-api/fixtures/revocationList.json",
     "type": [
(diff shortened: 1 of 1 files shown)


OR13 (approved):

fixtures/revocationList.json
```diff
@@ -18,9 +18,9 @@
     },
     "proof": {
       "type": "Ed25519Signature2018",
```
msporny, line 20: Wondering what folks feelings are for switching to Ed25519Signature2020? We're all speed ahead at DB for that change because it switches to using multibase, which gets good compression in CBOR-LD, and because it makes a clean break from jws, which is what JsonWebSignature2020 is for.

msporny (approved):
LGTM, would be interested if we can target Ed25519Signature2020 for this round of interop?

msporny:
This PR is ready for merge.

msporny (approved):
Editorial, fixes bug in test suite, no objections, merging.

PR #250, opened 2021-11-12
Add "proofValue" to LinkedDataProof.

peacekeeper:
Addresses https://github.com/w3c-ccg/vc-api/issues/194.

### DIFF
--- a/components/LinkedDataProof.yml
+++ b/components/LinkedDataProof.yml
@@ -30,7 +30,10 @@ components:
           description: The purpose of the proof to be used with verificationMethod.
         jws:
           type: string
-          description: Detached JSON Web Signature
+          description: Detached JSON Web Signature.
+        proofValue:
+          type: string
+          description: Value of the Linked Data proof.
       example:
         {
           "type": "Ed25519Signature2018",


peacekeeper:
^ @vdods

components/LinkedDataProof.yml
```diff
@@ -30,7 +30,10 @@ components:
           description: The purpose of the proof to be used with verificationMethod.
         jws:
           type: string
-          description: Detached JSON Web Signature
+          description: Detached JSON Web Signature.
+        proofValue:
+          type: string
+          description: Value of the Linked Data proof.
```
msporny, line 36: Hrm, we might want cryptosuite-specific limitations here:

https://github.com/digitalbazaar/ed25519-signature-2020-context/blob/master/contexts/ed25519-signature-2020-v1.jsonld

... but that results in an endless schema upgrade cycle. What we're doing right now isn't as tight as some security professionals would want (but you'd expect the cryptosuite to catch the really terrible things). So, we're probably going to have to make a decision soon about this -- is this schema fairly loose, or do we want the schema to be exceedingly restrictive. My suggestion is the latter, and if we do that, we're going to have to figure out how to not make this a burden on the API (and to enable easy extensibility for these cryptosuite schemas).

msporny (approved):
I'm fine w/ this going in for now, but we're definitely going to have to revisit it as we march towards a standard.

components/LinkedDataProof.yml
peacekeeper: I agree it's not ideal, but I think adding this fixes a bug in the current schema. I'm not enough of an OpenAPI wizard to know if it's possible to allow/disallow certain properties depending on the value of another property ("type"). Maybe the "oneOf" keyword could be used here somehow.

Ideally, the cryptosuite specs, JSON-LD contexts, and OpenAPI schemas should all match.

mkhraisha:
I think we may need to split the LinkedDataProof object depending on if it is in the context of a Presentation or a VC, I'm not 100% sure but I can't think of a reason for a VC to have challenge/domain fields for example

components/LinkedDataProof.yml
msporny: > Ideally, the cryptosuite specs, JSON-LD contexts, and OpenAPI schemas should all match.

Yes, agreed... hope we can find some way to automate this. One way would be to leverage some mythical LDI Cryptosuite Registry in the future and pull in the schemas from there. I believe the traceability folks have already something like this working in their test suite?

msporny:
Normative, at least one review by an Editor, no objections, it's been 22 days, merging.

Issue #245, opened 2021-11-09
PROPOSAL: Interactive Issue and Re-Issue VC API flow


dmitrizagidulin:
There is a number of use cases for VC issuing that require one or more interactive request-response rounds, where an Issuer requires certain pre-requisites before actually issuing requested VCs.

### Use Case Examples

For example, consider a typical Re-Issue flow.

1. Client: "Issuer, I would like an expired credential to be re-issued."
2. Issuer: "Great, please present me a copy of the expired credential, a fresh instance of DIDAuth, and supporting VCs X, Y, and Z. You may submit them to the following endpoint."
3. Client: (POSTs to the designated endpoint) "Here are your required pre-requisites, in a VP."
4. Issuer: "Here is you re-issued credential (wrapped in a VP, as usual)."

Similarly, consider this situation, that's frequently encountered during "onramp" deployments, where VC issuing needs to interop with an existing identity management system. A typical example of this is the issuing of an educational credential (such as for a completion of a course).

1. Client: "Issuer, I would like to receive a credential for the completion of a course."
2. Issuer: "Great, please provide the following pre-requisites: a) an instance of DIDAuth (with the DID that you want the credential issued to)., b) an instance of an OpenID Connect ID Token or Access Token from the university's OIDC provider (so that we can verify in our databases which student is requesting the credential, if they're eligible, etc)., and c) a VC receipt / proof of payment. Please provide them as a VP, to the following endpoint."
3. Client: "I have performed the OpenID Connect authorization flow, and received the required access token. Here are the pre-requisites, including the access token, in a VP signed by a DID, serving as an example of DID Auth."
4. Issuer: "Here is your proof of course completion credential."

These examples only have one round of back-and-forth negotiation between, but hopefully it's clear that multiple iterations of this process can also occur, before a VC gets issued.

### Existing Techniques and Components

This proposal takes into consideration several existing tech ecosystems, for inspiration / prior art.

* CHAPI + VP Request
* WACI Presentation Exchange
* SIOP v2 and OIDC for VPs
* GNAP

#### W3C CCG Credential Handler API (CHAPI)

[...]

PR #255, opened 2022-01-16
Define Workflow APIs

msporny:
This PR defines the Workflow APIs non-normatively alluded to in PR #254 by:

* Moving PR #254 into an API section called Workflows.
* Adding two additional endpoints that implement workflows.
* Identifying that they are generalizations of existing endpoints in the VC API.
* Adding all the Open API specification YAML files necessary to define the endpoints and JSON Schemas.

The diagram below (requested by @mprorock during the last call and included in the spec/PR) shows the basic protocol for generalized workflows as proposed in this PR:

image

The generalized workflow above is realized in the Verifiable Credential Refresh 2021 spec like so (the Verifiable Credential Refresh 2021 subflow below IS the generalized VC API workflow proposal above):

image

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/pull/255.html" title="Last updated on Jan 16, 2022, 4:39 PM UTC (7b4e007)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/255/dff796d...7b4e007.html" title="Last updated on Jan 16, 2022, 4:39 PM UTC (7b4e007)">Diff</a>


mprorock (approved):
This is very helpful, especially the response to the request section.  I would still like to see if we can get this in line with the trace side of things as it is very close, but i think the gaps appear to be more in terminology at this point.

peacekeeper (commented):
I think this is useful, but not sure if it has been thought through and discussed sufficiently, especially in relation to other things such as DIDComm, WACI PEx, etc..

Anyway, since the text says that it's an experimental features, it would probably be okay to merge.

dmitrizagidulin:
This PR also addresses issue https://github.com/w3c-ccg/vc-api/issues/245

dmitrizagidulin:
@peacekeeper 

> I think this is useful, but not sure if it has been thought through and discussed sufficiently, especially in relation to other things such as DIDComm, WACI PEx, etc..

I think there's a lot of discussion on #245, with regards to those other protocols, and this PR is basically an implementation on that.

mprorock:
#255 provides some good examples here

components/VerifiablePresentationRequest.yml
```diff
@@ -0,0 +1,81 @@
...
+    name: W3C Software and Document License
+    url: http://www.w3.org/Consortium/Legal/copyright-software.
+  contact:
+    name: GitHub Source Code
+    url: https://github.com/w3c-ccg/vc-api
+paths:
+components:
+  schemas:
+    VerifiablePresentationRequest:
+      type: object
```
OR13, line 16: any required fields here? or are they all optional?

components/WorkflowStartResponse.yml
```diff
@@ -0,0 +1,48 @@
...
+    url: https://github.com/w3c-ccg/vc-api
+paths:
+components:
+  schemas:
+    WorkflowStartResponse:
+      type: object
+      description: A response to a request to start a workflow.
+      properties:
+        verifiablePresentationRequest:
+          $ref: "./VerifiablePresentationRequest.yml#/components/schemas/VerifiablePresentationRequest"
```
OR13, line 20: it is possible to get better refs, with some effort.

index.html
```diff
@@ -493,8 +493,98 @@ <h4>Update Status</h4>
...
+
+    <section>
+      <h3>Workflows</h3>
+      <p>
+The following APIs are defined for executing Verifiable Credential workflows.
+      </p>
+
+      <table class="simple api-summary-table"
+        data-api-path="
+          /workflows/{name}/start
```
OR13, line 585: dislike using name parameter here, per our previous resolutions on naming conventions: https://restfulapi.net/resource-naming/

I would prefer a controller style endpoint for this:

```
POST /workflows/initiate
POST /workflows/submit
```

^ Notice that these are nearly identical to the existing holder exchange endpoints.

index.html
```diff
@@ -540,21 +636,21 @@ <h4>Automation via Workflows</h4>
...
         </p>
 
         <pre class="example"
           title="Step 1 (request to issuer): Initiate degree refresh workflow">
-  GET /workflows/refresh-degree HTTP/1.1
-  Host: example.edu
-  Content-Type: application/json
-  Accept: application/json, */*
-  Accept-Encoding: gzip, deflate
+POST /workflows/refresh-degree/start HTTP/1.1
```
OR13, line 644: really strongly dislike this example.

index.html
```diff
@@ -580,7 +676,7 @@ <h4>Automation via Workflows</h4>
       "interact": {
         "service": [{
           "type": "UnmediatedPresentationService2021",
-          "serviceEndpoint": "https://example.edu/active-flows/123456"
+          "serviceEndpoint": "https://example.edu/workflows/123456/presentations"
```
OR13, line 679: I don't hate the way this looks, but I worry about collisions on workflow `id` and its use in other systems.

Prefer for this endpoint and its parameters to be defined fully in this repo.

index.html
```diff
@@ -650,7 +746,7 @@ <h4>Automation via Workflows</h4>
     },
     "refreshService": {
       "type": "AutoRefresh2021",
-      "url": "https://example.edu/workflows/refresh-degree",
+      "url": "https://example.edu/workflows/refresh-degree/start",
```
OR13, line 749: how much of `{name}` can be carried in the post body?

index.html
```diff
@@ -726,85 +822,43 @@ <h4>Automation via Workflows</h4>
...
+allowing arbitrary data to be posted to the endpoint. It is suggested that we
+can merge the API endpoint defined in this section with the <a
+href="#presentation-availability">Presentation Availability</a> endpoint via
+the generalization provided in this section. The term "workflow" is a working
+title and is not intended to be the final feature name.
         </p>
 
         <div class="api-detail"
-          data-api-endpoint="post /credentials/derive"></div>
+          data-api-endpoint="post /workflows/{name}/start"></div>
```
OR13, line 840: embedding name like this will limit its potential to URL safe characters.... prefer to see it as a post body argument.

index.html
```diff
@@ -726,85 +822,43 @@ <h4>Automation via Workflows</h4>
...
-      </section>
-
-      <section>
-        <h4>Submit Presentation</h4>
-        <p>
-        </p>
-
-        <div class="api-detail"
-          data-api-endpoint="post /presentations/submissions"></div>
+          data-api-endpoint="post  /workflows/{id}/presentations"></div>
```
OR13, line 861: TODO: define the type of `id`.... UUID? URN? ... what about if `id` is a URL? lots of problems could be avoided by switching to a post body argument.

workflow.yml
```diff
@@ -0,0 +1,74 @@
...
+      description:
+        Starts a workflow with the given name. Optionally, additional information can be provided to the server in the request body that can help it fine tune what information to request from the client in order to continue with the workflow.
+      requestBody:
+        description: Optional and arbitrary POST data that the client might provide to the server to help it fine tune what information to request from the client in order to continue with the workflow.
+        content:
+          application/json:
+            schema:
+              oneOf:
+                - type: object
+                - type: string
```
OR13, line 25: I string a mistake here? lets not offer more types than we need... increases implementation burden... especially in languages that respect types.

workflow.yml
```diff
@@ -0,0 +1,74 @@
...
+      responses:
+        "201":
+          description: A new workflow was successfully created for the client.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/WorkflowStartResponse"
```
OR13, line 32: this is a proper ref.

workflow.yml
```diff
@@ -0,0 +1,74 @@
...
+        "400":
+          description: The workflow request was malformed.
+        "500":
+          description: Internal Error
+        "501":
+          description: Not implemented
+  /workflows/{id}/presentations:
+    post:
+      summary: Present information to an active workflow.
+      operationId: storePresentation
```
OR13, line 42: abuse of existing operationId... see https://github.com/w3c-ccg/vc-api/blob/main/holder.yml#L88


perhaps more evidence that these APIs already exist in some form?

OR13, line 42:
```suggestion
      operationId: storeWorkflowPresentation
```

workflow.yml
```diff
@@ -0,0 +1,74 @@
...
+        "400":
+          description: Presentation is malformed
+        "501":
+          description: Not Implemented
+        "500":
+          description: Internal Server Error
+components:
+  schemas:
+    WorkflowStartResponse:
+      $ref: "./components/WorkflowStartResponse.yml#/components/schemas/WorkflowStartResponse"
```
OR13, line 72: you can acctually seperate schemas and responses with folders, see https://github.com/dgarcia360/openapi-boilerplate/tree/master/src

OR13 (changes requested):
only blocking change request is the collision on `operationId`... the rest can be fixed in subsequent PRs... our aim should be to end with either "workflows" or "holder exchange apis" not both... but we can and probably should tollerate an intermediate state where we have both.

index.html
mprorock: this should definitely be defined, likely as a UUID

mprorock (changes requested):
This is generally in the right direction but we need to get it more in line with proper REST conventions

index.html
OR13: if we are going to include `id` in the route... its domain becomes limited to URL safe IDs... aka not other URLs..

When you are the operator of the resource server, and you assign the ID, this is a safe assumption... but I have seen this create endless pain that is easily avoided by proper use of the `controller` naming convention.

also consider the current "status" endpoints... they pass the `credentialId` via post body... mostly because it can be a URL.

index.html
OR13: yes, it MUST be defined if it is assumed to be in the route... it may be assumed to be a string, if it's in the post body.

components/VerifiablePresentationRequest.yml
msporny: At present, all optional. There could be an argument to make at least "query" required. Raised an issue to track this concern: https://github.com/w3c-ccg/vp-request-spec/issues/16

components/WorkflowStartResponse.yml
msporny: Fixed in e9dc263eb99297529f56ac4aee950a2211fdf9ba.

index.html
msporny: A couple of points: 

1) It sounds like @mprorock is challenging the "controller style endpoint" pattern that the group provisionally agreed to. If that is the case, I agree with @mprorock -- I've never liked the "controller style endpoint" pattern and we should see if the group has new opinions on this now that we have some implementation experience under our belt.

2) I agree with @mprorock's comment on the `POST /workflows/{name|ID}` or `PUT /workflows/{name|ID}` pattern.

3) I wrote this up in this way because when I asked the question to the group, @OR13 wanted it to be defined in this way rather than it be a "pattern" that can be used. That is, I'd rather /not/ define ONE endpoint to start all workflows or accept all presentations. Developers should be free to choose any endpoint to act in this way. I'm not going to argue that here, though. I just needed to get this down in a way that @OR13 felt like we could have a discussion over it. I think focusing on the name/structure of the URL is probably too deep in the weeds at this point of the analysis.

4) I agree that there is a debate to be had about whether or not `{id}` belongs in the URL or if it belongs in a POST body. One argument to allow it in the URL is for when browsers are invoked by native apps, where it is not possible to perform a POST. 

In short, there is a lot to discuss here, so I've raised an issue to track this concern: https://github.com/w3c-ccg/vc-api/issues/256

index.html
msporny: I'm sorry. :(

index.html
msporny: It is fully defined here (in this PR): https://github.com/w3c-ccg/vc-api/pull/255/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R861

index.html
msporny: There is a debate to be had over where it belongs. There are good arguments to put it in either place. I've raised this issue to track the debate: https://github.com/w3c-ccg/vc-api/issues/256

index.html
msporny: Yes, understood. Being tracked in this issue: https://github.com/w3c-ccg/vc-api/issues/256

This will be a topic of discussion on the next VC API call.

index.html
msporny: Agree that it needs to be defined more strictly. There is a tracking issue for where we put it here: https://github.com/w3c-ccg/vc-api/issues/256

workflow.yml
msporny: Hrm, for `application/json` it is a mistake. Fixed in 1574f791143531155320cce458ecbae3bb5fc9a5.

What I was going for is trying to convey that any arbitrary post data is probably fine here. For example, it should be fine if someone wanted to post `application/octet-stream` or `application/x-www-form-urlencoded` if their workflows supported that. I almost added it into this PR, but didn't. We don't want to make posting content other than `application/json` illegal. That said, we can deal with that when someone requests it.

workflow.yml
msporny: Fixed the one above to look more like this one.

workflow.yml
msporny: Good catch, wasn't meaning to do that. Fixed in 95535e63b44331b4af5fe33f34ea39e37c709544.

workflow.yml
msporny: Sounds like something we should do. Will do this in a separate PR.

msporny:
> This is generally in the right direction but we need to get it more in line with proper REST conventions

Raised issue https://github.com/w3c-ccg/vc-api/issues/257 to track this concern.

PR #258, opened 2022-02-04
Refactor presentation endpoints into exchange endpoints

msporny:
This PR is based on ACTIONS requested to be taken on the last VC API call (see summary at top of transcription): https://w3c-ccg.github.io/meetings/2022-02-01-vcapi/

The goals here are to:

- [x] Refactor the existing traceability holder APIs (in a backwards-compatible way) using input from workflows section and 'exchanges' discussion on VC API call.
- [ ] Merge non-normative text from workflows into new refactored sections (with appropriate wording changes, e.g., "workflow" -> "exchange").
- [ ] Remove workflow sections once everything has been merged together. 

I've done the first of three items and will halt here for feedback from the Traceability work item group to ensure that this meets their expectations from the call two days ago.

The target merge window is in 7 days, after the VC API call next week.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/pull/258.html" title="Last updated on Feb 9, 2022, 9:45 PM UTC (fc195cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/258/6b8c865...fc195cd.html" title="Last updated on Feb 9, 2022, 9:45 PM UTC (fc195cd)">Diff</a>

### DIFF
--- a/components/ErrorResponse.yml
+++ b/components/ErrorResponse.yml
@@ -1,0 +1,33 @@
+openapi: 3.0.0
+info:
+  version: "0.0.3-unstable"
+  title: VC HTTP API
+  description: This is an Experimental Open API Specification for the [VC Data Model](https://www.w3.org/TR/vc-data-model/).
+  license:
+    name: W3C Software and Document License
+    url: http://www.w3.org/Consortium/Legal/copyright-software.
+  contact:
+    name: GitHub Source Code
+    url: https://github.com/w3c-ccg/vc-api
+paths:
+components:
+  schemas:
+    ErrorResponse:
+      type: object
+      description: A response that denotes that an error has occurred.
+      properties:
+        id:
+          type: "string"
+          pattern: "[a-z0-9\\-]{8,}"
+        message:
+          type: "string"
+          minLength: 10
+          maxLength: 100
+        details:
+          type: "object"
+      required: ["id", "message"]
+      example:
+        {
+          "id": "invalid-json",
+          "message": "The provided JSON data was malformed."
+        }
--- a/holder.yml
+++ b/holder.yml
@@ -101,7 +101,7 @@ paths:
         "500":
           description: Internal server error.
   /exchanges/{exchange-id}/{transaction-uuid}:
-    post:
+    put:
       summary: Receives information related to an existing exchange.
       operationId: receiveExchangeData
       description:
--- a/holder.yml
+++ b/holder.yml
@@ -106,13 +106,8 @@ paths:
       operationId: receiveExchangeData
       description:
         A client can use this endpoint to continue the exchange of information
-        associated with an initiated exchange. A Verifiable Presentation is
-        sent to this endpoint. If the server has received all of the information
-        it needs, a 200 OK is returned with either an empty response or a
-        Verifiable Presentation containing requested credentials by the client.
-        If the server requires more information, it returns a Verifiable
-        Presentation Request. A request that the server cannot understand
-        results in an error.
+        associated with an initiated exchange by sending a Verifiable Presentation 
+        with information requested by the server to this endpoint.
       parameters:
         - $ref: "./components/parameters/path/ExchangeId.yml"
         - $ref: "./components/parameters/path/TransactionUuid.yml"
--- a/holder.yml
+++ b/holder.yml
@@ -67,13 +67,7 @@ paths:
         the request, it returns a Verifiable Presentation Request. A request
         that the server cannot understand results in an error.
       parameters:
-        - in: path
-          name: exchange-id
-          required: true
-          schema:
-            type: string
-            minimum: 3
-            pattern: "[a-z0-9\\-]{3,}"
+        - $ref: "./components/parameters/path/ExchangeId.yml"
       requestBody:
         description:
           Information related to the type of exchange the client would like
@@ -120,20 +114,8 @@ paths:
         Presentation Request. A request that the server cannot understand
         results in an error.
       parameters:
-        - in: path
-          name: exchange-id
-          description:
-            A human readable exchange type identifier such as "diplomas".
-          required: true
-          schema:
-            type: string
-            pattern: "[a-z0-9\\-]{3,}"
-        - in: path
-          name: transaction-uuid
-          required: true
-          schema:
-            type: string
-            pattern: "([0-9a-f]{8}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{12}){0,1}"
+        - $ref: "./components/parameters/path/ExchangeId.yml"
+        - $ref: "./components/parameters/path/TransactionUuid.yml"
       requestBody:
         description:
           A Verifiable Presentation.
--- a/holder.yml
+++ b/holder.yml
@@ -56,60 +56,108 @@ paths:
           description: invalid input!
         "500":
           description: error!
-  /presentations/available:
+  /exchanges/{exchange-id}:
     post:
-      summary: Notifies a holder of an available presentation.
-      operationId: notifyPresentationAvailable
+      summary: Initiates an exchange of information.
+      operationId: initiateExchange
       description:
-        This API is the first part of a holder to holder credential exchange over http.
-        Notifies this server-holder that the client-holder (caller) is prepared to deliver a presentation compliant with the provided Verifiable Presentation Request query. This is the first part of a presentation exchange over this HTTP API. This server-holder MUST return the required 'domain' and 'challenge' parameters for the Presentation. This server-holder MAY reply with an alternate Verifiable Presentation Request query if it requires something different than the one proposed by the client-holder. The client-holder MUST return a Presentation in accordance with the response parameters.
+        A client can use this endpoint to initiate an exchange of a particular
+        type. The client can include HTTP POST information related to the
+        details of exchange it would like to initiate. If the server understands
+        the request, it returns a Verifiable Presentation Request. A request
+        that the server cannot understand results in an error.
+      parameters:
+        - in: path
+          name: exchange-id
+          required: true
+          schema:
+            type: string
+            minimum: 3
+            pattern: "[a-z0-9\\-]{3,}"
       requestBody:
-        description: Details the client provides to the server to help it decide if the presentation should be made.
+        description:
+          Information related to the type of exchange the client would like
+          to start.
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/NotifyPresentationAvailableRequest"
+              anyOf:
+                -
+                  {
+                    "type": "object",
+                    "description": "Data necessary to initiate the exchange."
+                  }
+                -
+                  $ref: "#/components/schemas/NotifyPresentationAvailableRequest"
       responses:
         "200":
-          description: Proceed with presentation
+          description: Proceed with exchange.
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/NotifyPresentationAvailableResponse"
+                $ref: "#/components/schemas/VerifiablePresentationRequest"
         "400":
-          description: Request for presentation is malformed
+          description: Request is malformed.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
         "501":
-          description: Not implemented
+          description: Service not implemented.
         "500":
-          description: internal error
-  /presentations/submissions:
+          description: Internal server error.
+  /exchanges/{exchange-id}/{transaction-uuid}:
     post:
-      summary: Provide a presentation for a holder to store.
-      operationId: storePresentation
-      description: "Delivery endpoint for a client-holder to provide a presentation in compliance with the Verifiable Presentation Request provided in the Notification call. This server-holder MUST match the 'domain' and 'challenge' to a 'domain' and 'challenge' previously sent in a Notification Response, and not already received. The server-holder MUST verify the Presentation. The server MUST store the Presentation if verification passes."
+      summary: Receives information related to an existing exchange.
+      operationId: receiveExchangeData
+      description:
+        A client can use this endpoint to continue the exchange of information
+        associated with an initiated exchange. A Verifiable Presentation is
+        sent to this endpoint. If the server has received all of the information
+        it needs, a 200 OK is returned with either an empty response or a
+        Verifiable Presentation containing requested credentials by the client.
+        If the server requires more information, it returns a Verifiable
+        Presentation Request. A request that the server cannot understand
+        results in an error.
+      parameters:
+        - in: path
+          name: exchange-id
+          description:
+            A human readable exchange type identifier such as "diplomas".
+          required: true
+          schema:
+            type: string
+            pattern: "[a-z0-9\\-]{3,}"
+        - in: path
+          name: transaction-uuid
+          required: true
+          schema:
+            type: string
+            pattern: "([0-9a-f]{8}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{12}){0,1}"
       requestBody:
-        description: Details the client provides to the server to help it decide if the presentation should be made.
+        description:
+          A Verifiable Presentation.
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/StorePresentationRequest"
+              $ref: "#/components/schemas/VerifiablePresentation"
       responses:
-        "202":
-          description: Presentation accepted
+        "200":
+          description: Received data was accepted.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/VerifiablePresentationRequest"
         "400":
-          description: Presentation is malformed
-        "401":
-          description: Presentation did not contain a proof
-        "402":
-          description: Payment required
-        "403":
-          description: Presentation verification failed
-        "425":
-          description: Server is unwilling to risk processing a request that might be replayed
-        "501":
-          description: Not implemented
+          description: Received data is malformed.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
         "500":
-          description: internal error
+          description: Internal server error.
+        "501":
+          description: Service not implemented.
 components:
   schemas:
     DeriveCredentialRequest:
@@ -213,5 +261,11 @@ components:
           "domain": "jobs.example.com",
           "challenge": "3182bdea-63d9-11ea-b6de-3b7c1404d57f",
         }
+    ErrorResponse:
+      $ref: "./components/ErrorResponse.yml#/components/schemas/ErrorResponse"
     StorePresentationRequest:
       $ref: "./components/VerifiablePresentation.yml#/components/schemas/VerifiablePresentation"
+    VerifiablePresentation:
+      $ref: "./components/VerifiablePresentation.yml#/components/schemas/VerifiablePresentation"
+    VerifiablePresentationRequest:
+      $ref: "./components/VerifiablePresentationRequest.yml#/components/schemas/VerifiablePresentationRequest"
--- a/index.html
+++ b/index.html
@@ -569,7 +569,7 @@ The following APIs are defined for presenting a Verifiable Credential:
         </p>
 
         <div class="api-detail"
-          data-api-endpoint="post /exchanges/{exchange-id}/{transaction-uuid}"></div>
+          data-api-endpoint="put /exchanges/{exchange-id}/{transaction-uuid}"></div>
       </section>
 
     </section>
--- a/index.html
+++ b/index.html
@@ -555,21 +555,21 @@ The following APIs are defined for presenting a Verifiable Credential:
       </section>
 
       <section>
-        <h4>Presentation Availability</h4>
+        <h4>Initiate Exchange</h4>
         <p>
         </p>
 
         <div class="api-detail"
-          data-api-endpoint="post /presentations/available"></div>
+          data-api-endpoint="post /exchanges/{exchange-id}"></div>
       </section>
 
       <section>
-        <h4>Submit Presentation</h4>
+        <h4>Continue Exchange</h4>
         <p>
         </p>
 
         <div class="api-detail"
-          data-api-endpoint="post /presentations/submissions"></div>
+          data-api-endpoint="post /exchanges/{exchange-id}/{transaction-uuid}"></div>
       </section>
 
     </section>
--- a/components/parameters/path/ExchangeId.yml
+++ b/components/parameters/path/ExchangeId.yml
@@ -1,0 +1,8 @@
+name: exchange-id
+description: A potentially human-readable identifier for an exchange.
+in: path
+required: true
+schema:
+  type: string
+  minimum: 3
+  pattern: "[a-z0-9][a-z0-9\\-]{2,}"
--- a/components/parameters/path/TransactionUuid.yml
+++ b/components/parameters/path/TransactionUuid.yml
@@ -1,0 +1,6 @@
+in: path
+name: transaction-uuid
+required: true
+schema:
+  type: string
+  pattern: "([0-9a-f]{8}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{12}){0,1}"


holder.yml
```diff
@@ -56,60 +56,108 @@ paths:
...
             schema:
-              $ref: "#/components/schemas/NotifyPresentationAvailableRequest"
+              anyOf:
+                -
+                  {
+                    "type": "object",
+                    "description": "Data necessary to initiate the exchange."
+                  }
+                -
+                  $ref: "#/components/schemas/NotifyPresentationAvailableRequest"
```
msporny, line 91: I kept this here for the time being, we could probably keep it here -- but I do want to have a chat w/ the Traceability folks to understand why they're approaching the problem in this way. There might be a simpler way.

holder.yml
```diff
@@ -56,60 +56,108 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
         "501":
-          description: Not implemented
+          description: Service not implemented.
         "500":
-          description: internal error
-  /presentations/submissions:
+          description: Internal server error.
+  /exchanges/{exchange-id}/{transaction-uuid}:
```
msporny, line 109: @msporny @mkhraisha -- I'm not sure if this is what you guys were talking about wrt. two IDs to track a particular exchange or not?

holder.yml
```diff
@@ -56,60 +56,108 @@ paths:
...
+        the request, it returns a Verifiable Presentation Request. A request
+        that the server cannot understand results in an error.
+      parameters:
+        - in: path
+          name: exchange-id
+          required: true
+          schema:
+            type: string
+            minimum: 3
+            pattern: "[a-z0-9\\-]{3,}"
```
msporny, line 76: This is a very western-centric regex. I expect people to object to the fact we're not allowing non-western characters in here (in time).

index.html
```diff
@@ -555,21 +555,21 @@ <h4>Prove Presentation</h4>
       </section>
 
       <section>
-        <h4>Presentation Availability</h4>
+        <h4>Initiate Exchange</h4>
```
msporny, line 558: @jandrieu I waffled back and forth on "Initiate Presentation Exchange" and what I ended up using above. Any ideas? Is this what you were going for?

index.html
```diff
@@ -555,21 +555,21 @@ <h4>Prove Presentation</h4>
...
         </p>
 
         <div class="api-detail"
-          data-api-endpoint="post /presentations/available"></div>
+          data-api-endpoint="post /exchanges/{exchange-id}"></div>
       </section>
 
       <section>
-        <h4>Submit Presentation</h4>
+        <h4>Continue Exchange</h4>
```
msporny, line 567: @jandrieu same here -- "continue exchange" seems strange... maybe? I don't know. Any suggestions?

components/ErrorResponse.yml
```diff
@@ -0,0 +1,33 @@
...
+  license:
+    name: W3C Software and Document License
+    url: http://www.w3.org/Consortium/Legal/copyright-software.
+  contact:
+    name: GitHub Source Code
+    url: https://github.com/w3c-ccg/vc-api
+paths:
+components:
+  schemas:
+    ErrorResponse:
```
OR13, line 15: finally! real and potentially useful errors! <3

holder.yml
```diff
@@ -56,60 +56,108 @@ paths:
           description: invalid input!
         "500":
           description: error!
-  /presentations/available:
+  /exchanges/{exchange-id}:
```
OR13, line 59: please define exchange id as a proper parameter, here is an example:

https://github.com/dgarcia360/openapi-boilerplate/blob/master/src/parameters/path/petId.yaml

please include an example as well.

holder.yml
```diff
@@ -56,60 +56,108 @@ paths:
...
       description:
-        This API is the first part of a holder to holder credential exchange over http.
-        Notifies this server-holder that the client-holder (caller) is prepared to deliver a presentation compliant with the provided Verifiable Presentation Request query. This is the first part of a presentation exchange over this HTTP API. This server-holder MUST return the required 'domain' and 'challenge' parameters for the Presentation. This server-holder MAY reply with an alternate Verifiable Presentation Request query if it requires something different than the one proposed by the client-holder. The client-holder MUST return a Presentation in accordance with the response parameters.
+        A client can use this endpoint to initiate an exchange of a particular
+        type. The client can include HTTP POST information related to the
+        details of exchange it would like to initiate. If the server understands
+        the request, it returns a Verifiable Presentation Request. A request
+        that the server cannot understand results in an error.
+      parameters:
+        - in: path
```
OR13, line 70: you can define these one at the top and reuse them see:

https://github.com/dgarcia360/openapi-boilerplate/blob/master/src/parameters/path/petId.yaml

holder.yml
OR13: Why would we encourage so much diversity in characters for an id that all parties need to agree too?

See https://github.com/json-schema-org/json-schema-spec/issues/542

I think we should consider if this `id` also shows up inside VPs and VCs and if so, is in a proper IRI or not.

I prefer for it to be `urn:uuid:`.

holder.yml
```diff
@@ -56,60 +56,108 @@ paths:
...
+          Information related to the type of exchange the client would like
+          to start.
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/NotifyPresentationAvailableRequest"
+              anyOf:
+                -
+                  {
+                    "type": "object",
```
OR13, line 87: let's not add this... since you just destroyed the ability to know the shape of the data... of course if you have a new shape you want to define I would be happy to see it defined and added but this harmful for interop as is.

holder.yml
OR13: there is always a simpler way, but it must be defined in a schema to be accepted :)

holder.yml
```diff
@@ -56,60 +56,108 @@ paths:
...
+                  $ref: "#/components/schemas/NotifyPresentationAvailableRequest"
       responses:
         "200":
-          description: Proceed with presentation
+          description: Proceed with exchange.
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/NotifyPresentationAvailableResponse"
+                $ref: "#/components/schemas/VerifiablePresentationRequest"
```
OR13, line 98: nice!

holder.yml
OR13: not a fan of this ... I am pretty sure `{transaction-uuid}` is a post body argument in our cases. 

these are values that are inside VPs or VCs for the purposes of relating them (intentional correlation fields).

holder.yml
```diff
@@ -56,60 +56,108 @@ paths:
...
+          required: true
+          schema:
+            type: string
+            pattern: "[a-z0-9\\-]{3,}"
+        - in: path
+          name: transaction-uuid
+          required: true
+          schema:
+            type: string
+            pattern: "([0-9a-f]{8}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{12}){0,1}"
```
OR13, line 136: hmm might be better to make this an IRI same comment as before, it may show up in other JSON-LD `id` fields.

OR13 (approved):

OR13 (changes requested):
Blocking:
1. remove the "any object" is a valid request part.
2. remove the `transaction-uuid` part.

Nonblocking:

We should handle all `id` values consistently... this is not a place to support diversity.

I suggest we make them all valid IRIs so that they can be safely embedded in JSON-LD.

I propose `urn:uuid:`.

holder.yml
msporny: Done in 748d5be9eb54dacde1d5b5017d2b9d458e93164c.

Gave up trying to correctly reference them using JSON Schemas completely broken referencing system after 30 minutes and referred to them directly. We have a broken pattern at the bottom of many of these files w/ relative references that we're going to want to clean up.

holder.yml
msporny: > Why would we encourage so much diversity in characters for an id that all parties need to agree too?

Not all parties need to agree to this identifier. It is (or can be) issuer-specific. 

> I prefer for it to be urn:uuid:

I don't understand this desire, so we should discuss on the next VC API call. In any case, my suggestion is that we deal w/ the range of exchange-id later.

holder.yml
msporny: > let's not add this... since you just destroyed the ability to know the shape of the data... of course if you have a new shape you want to define I would be happy to see it defined and added but this harmful for interop as is.

Accepting an arbitrary object with arbitrary fields is exactly what we want here. We discussed this on the last VC API call and there wasn't push back on it. You need to show up on the next VC API call and make your point (or clearly articulate it here). Please refer to the discussion that you missed here: https://w3c-ccg.github.io/meetings/2022-02-01-vcapi/#topic-2

holder.yml
msporny: Let's discuss on next VC API call -- it's not clear to me why the Traceability folks chose to use a VPR here.

holder.yml
msporny: > not a fan of this ... I am pretty sure {transaction-uuid} is a post body argument in our cases.

This is not what I heard @mkhraisha and @mprorock request on the call. It was not entirely clear to me, but these were the ACTIONS that were taken on the last VC API call and what I implemented. This was the "We need to follow REST principles" argument that @mprorock was making throughout the call, which I agree with, so we really need to sort this out as it was this approach that enabled us to merge the presentation/workflow endpoints in the first place.

holder.yml
msporny: Need a use case to understand what you're saying. I'm sure it's relevant, and yes, we need to be careful about things like this showing up in JSON-LD `id` fields... that said, it's a UUID in a URL, which can fit in a JSON-LD `id` field.

components/ErrorResponse.yml
msporny: If people like this, I'll apply it to all 400 error codes. This standard object also enables us to define standard error code responses more specifically in the spec.

msporny:
> Blocking:
> 
> 1. remove the "any object" is a valid request part.

This was not an objection by the other two traceability cohort members, so I suggest you speak with them to see why they didn't object to this on the call. As explained in https://github.com/w3c-ccg/vc-api/pull/258#discussion_r799852493, DB requires this feature for its customers and product line. The reason it's an empty object is because our products enable arbitrary objects to be posted to start a flow (and we have server-side logic that does stuff with that arbitrary input -- for a particular exchange).

Let me know where the Traceability folks want to go on this. DB will object to being blocked on both `workflows` and merging the APIs into a unified API if the only reason is because Transmute doesn't want to allow arbitrary post data to start exchanges. If you don't want to support that, you don't have to -- return an error code.

> 2. remove the `transaction-uuid` part.

This was specifically requested by two of the members in the Traceability cohort. I suggest you discuss this with them first and then come back with a group suggestion. DB requires `transaction-uuid` for it's credential refresh workflows with our customers.

> Nonblocking:
> 
> We should handle all `id` values consistently... this is not a place to support diversity.

If we want to take a unified approach, DB will object to using only UUIDs everywhere. We will suggest being able to select between multibase character sets limited to 128-bit values OR UUIDs.

> I suggest we make them all valid IRIs so that they can be safely embedded in JSON-LD.

I could agree to "the entire URL should be a valid IRI", but I'll note that includes UTF-8 characters and some fairly hairy encoding/decoding rules. The rules at present ensure that the entire URL can be safely embedded in JSON-LD. Are you suggesting that we expand the scope of valid value range for all VC API endpoint URLs to include UTF-8 characters?

> I propose `urn:uuid:`.

Confused by this, so we should discuss.

mprorock:
> We will suggest being able to select between multibase character sets limited to 128-bit values OR UUIDs.

I am good with this

mprorock:
> The reason it's an empty object is because our products enable arbitrary objects to be posted to start a flow (and we have server-side logic that does stuff with that arbitrary input -- for a particular exchange).

I am totally fine if the server says, "I accept JSON" and an empty JSON object is sent.  It is up to the server to know what to do with it

mprorock (approved):
This is not yet perfect, but it moves significantly in the right direction.  I say we get it in and move things forward, opening issues if we need to address specific items.  If we we in a WG I might be more inclined to get it "just right"

holder.yml
OR13: > Accepting an arbitrary object with arbitrary fields is exactly what DB needs here (we have customers that require exchange types to be set up, where they take arbitrary JSON input for a defined exchange type, and then use that to drive VPR logic).

I doubt very much that your customers like reading API docs that say "just give us any JSON you can imagine"...

However, I would bet they would be happy to see a schema defined that accepting some additional properties.

You can set `additionalProperties: true` and still provide a useful API definition for the request format.

holder.yml
OR13: We chose to support VPR in the hopes that eventually browser and HTTP APIs would support the same request / respond query data model.

holder.yml
OR13: The current endpoints support `POST /presentations/available` and `POST /presentations/submissions`... its not clear to me that adding another URL parameter does anything helpful wrt presentation exchange, given we don't need it currently.

Controller style endpoints are defined in the style guide we agreed to use.

https://restfulapi.net/resource-naming/

for example:

```
http://api.example.com/cart-management/users/{id}/cart/checkout
```

Since this PR is attempting to merge the existing endpoints with a new proposed one, I suggest only adding 1 `id` at a time... and given the `id` here is `exchange-id`.. I don't understand the relation to the other IDs or the existing use cases for grouping presentations by unique identifier.

holder.yml
```diff
@@ -56,60 +56,90 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
         "501":
-          description: Not implemented
+          description: Service not implemented.
         "500":
-          description: internal error
-  /presentations/submissions:
+          description: Internal server error.
+  /exchanges/{exchange-id}/{transaction-uuid}:
```
mprorock, line 103: So some clarification:
`/exchanges/{exchange-id}/{transaction-uuid}` is got for `PUT` because you already know the UUID of the object you are referrencing - on `POST` you are creating and object, in this case of a certain type, e.g. `/exchanges/{exchange-id}/` makes sense.

In the case of REST we should have the IDs for objects in both URI and in `body`

@msporny does that make sense

mprorock (changes requested):
see note around diff between PUT and POST

holder.yml
mprorock: This article on the topic may help clarify what I mean [https://restfulapi.net/rest-put-vs-post/](https://restfulapi.net/rest-put-vs-post/)

OR13: following normal collection style naming convention, I would assume:

```
POST /exchanges -> create {id: 123}
PUT or PATCH /exchanges/{id} -> update an existing resource with the given id.
```

I don't know what `transaction-uuid` is... and I suggest we not create it.

holder.yml
OR13: A. `d2dfc100-9384-49e4-b547-5e7d84ec5bf5` -> not an IRI

B. `urn:uuid:d2dfc100-9384-49e4-b547-5e7d84ec5bf5` -> IRI

C. `https://app.example/d2dfc100-9384-49e4-b547-5e7d84ec5bf5` -> IRI

D. `https://app.example/urn:uuid:d2dfc100-9384-49e4-b547-5e7d84ec5bf5` -> IRI

My understanding of JSON-LD is that only B-D would be acceptable for a `credentialSubject.id` or `issuer.id`... does that match you understanding?

holder.yml
```diff
@@ -56,60 +56,90 @@ paths:
...
-      summary: Provide a presentation for a holder to store.
-      operationId: storePresentation
-      description: "Delivery endpoint for a client-holder to provide a presentation in compliance with the Verifiable Presentation Request provided in the Notification call. This server-holder MUST match the 'domain' and 'challenge' to a 'domain' and 'challenge' previously sent in a Notification Response, and not already received. The server-holder MUST verify the Presentation. The server MUST store the Presentation if verification passes."
+      summary: Receives information related to an existing exchange.
+      operationId: receiveExchangeData
+      description:
+        A client can use this endpoint to continue the exchange of information
+        associated with an initiated exchange. A Verifiable Presentation is
+        sent to this endpoint. If the server has received all of the information
+        it needs, a 200 OK is returned with either an empty response or a
```
OR13, line 111: a bit of a nit, but your description is redundant to the description for the supported statuses.

You can keep the description focused on defining the `operationId` and use the status to convey the different reponses.

holder.yml
```diff
@@ -56,60 +56,90 @@ paths:
...
-          description: Presentation did not contain a proof
-        "402":
-          description: Payment required
-        "403":
-          description: Presentation verification failed
-        "425":
-          description: Server is unwilling to risk processing a request that might be replayed
-        "501":
-          description: Not implemented
+          description: Received data is malformed.
```
OR13, line 134: I don't see how this could ever be the case given the api definition says that any data is legal X )

Implementers should be throwing 400 as soon as they have validated the request data according to a schema, per OWASP best practices...

https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html

holder.yml
msporny: > I doubt very much that your customers like reading API docs that say "just give us any JSON you can imagine"...

Clearly, that is not the case.

The case is this: The customers are the ones that define what is valid to send to a particular exchange endpoint... the /customer/ supplies the JSON Schema and then what to do with the input JSON.

So, to clarify for the endpoint: `/exchanges/age-verification-training`, the client system posts `{"retailerId": "abc-store-145", "clerkId": "423"}` to the endpoint, and the issuer server logic goes 1) does that match the JSON Schema, configured by the administrator, for that endpoint and 2) runs code that takes `retailerId` and `clerkId` as input, to produce a VPR that is specific to that retailer and clerk (request a specific employment credential).

Do you understand the use case? We can't limit the JSON Schema for our use case because the input set is unbounded for our servers (but the Traceability folks are free to limit the input set to what makes sense to them).

> However, I would bet they would be happy to see a schema defined that accepted some additional properties.

No, they really don't want us to define any JSON Schema. They want to do it.

> You can set `additionalProperties: true` and still provide a useful API definition for the request format.

Not really, see above.

holder.yml
msporny: VPR is supported in the response format for this call (so there would be no drop in support for VPR). What I'm wondering is why /POST/ a VPR. What you get in return is a VPR. So, with the traceability stuff (IIUC) you POST a VPR and then get a VPR in response.

holder.yml
msporny: > @msporny does that make sense

Yes, I think what you want is for `POST /exchanges/{exchange-id}/{transaction-uuid}` to be changed to `PUT  /exchanges/{exchange-id}/{transaction-uuid}` ... if so, I'm good with that.

However, it seems like @mprorock and @OR13 disagree? Still hard for me to tell.

To be clear, Digital Bazaar would like to be able to identify different types of exchanges at different endpoints (load balancing via URL instead of having to look into each message -- useful at scale). We would also like to be able to refer to individual instances of each exchange (again, load balancing via URL instead of having to look into each message -- also useful at scale). Remember that the age verification system does >40 million individual interactions per day, and the number of HTTP calls are multiples of that... so, not having to look into the payload body is of significant importance at scale.

holder.yml
msporny: > @msporny does that make sense

Yes, I think what you want is for `POST /exchanges/{exchange-id}/{transaction-uuid}` to be changed to `PUT  /exchanges/{exchange-id}/{transaction-uuid}` ... if so, I'm good with that.

However, it seems like @mprorock and @OR13 disagree? Still hard for me to tell.

To be clear, Digital Bazaar would like to be able to identify different types of exchanges at different endpoints (load balancing via URL instead of having to look into each message -- useful at scale). We would also like to be able to refer to individual instances of each exchange (again, load balancing via URL instead of having to look into each message -- also useful at scale). Remember that the age verification system does >40 million individual interactions per day, and the number of HTTP calls are multiples of that... so, not having to look into the payload body is of significant importance at scale.

Moving these things into the payload body doesn't allow us to use some of the faster load balancing frameworks out there that only have to look at the first few lines of the HTTP Headers instead of having to consume the entire message.

holder.yml
```diff
@@ -56,60 +56,90 @@ paths:
...
+      operationId: receiveExchangeData
+      description:
+        A client can use this endpoint to continue the exchange of information
+        associated with an initiated exchange. A Verifiable Presentation is
+        sent to this endpoint. If the server has received all of the information
+        it needs, a 200 OK is returned with either an empty response or a
+        Verifiable Presentation containing requested credentials by the client.
+        If the server requires more information, it returns a Verifiable
+        Presentation Request. A request that the server cannot understand
+        results in an error.
```
msporny, line 115:
```suggestion
        A client can use this endpoint to continue the exchange of information
        associated with an initiated exchange by sending a Verifiable Presentation 
        with information requested by the server to this endpoint.
```

holder.yml
msporny: @OR13 wrote:
> a bit of a nit, but your description is redundant to the description for the supported statuses.

Does this work for you? https://github.com/w3c-ccg/vc-api/pull/258/files#r801206879

holder.yml
msporny: @OR13 wrote:
> I don't see how this could ever be the case

Please see this for a more details explanation with a use case: https://github.com/w3c-ccg/vc-api/pull/258/files#r801201273

holder.yml
```diff
@@ -56,60 +56,90 @@ paths:
...
+                $ref: "#/components/schemas/ErrorResponse"
         "501":
-          description: Not implemented
+          description: Service not implemented.
         "500":
-          description: internal error
-  /presentations/submissions:
+          description: Internal server error.
+  /exchanges/{exchange-id}/{transaction-uuid}:
     post:
```
msporny, line 104:
```suggestion
    put:
```

Is this what you're looking for, @mprorock?

mprorock: close.  let me take a stab in the code - might make it clearer

mprorock: This PR has gotten to be a bit of a mess with comments and suggestions, but rescanning through to make sure i didn't miss any key points.  focusing on `holder.yml` for these suggestions

the create a new one of type looks workable, though as @OR13 we may want to explicitly define some types for the `POST` body, I understand the use case for @msporny though, and would like to see if we can ideally satisfy a desire for stronger typing and a more flexible use case

```
/exchanges/{exchange-id}:
    post:
      summary: Creates an exchange as specified by type identified in exchange-id.
      operationId: createExchangeData
```

then onto the `PUT` etc.

```
/exchanges/{exchange-id}/{transaction-uuid}:
    put:
      summary: Updates information related to an existing exchange.
      operationId: updateExchangeData
```

There likely should be defined corresponding getters as well, likely allowing query by example if appropriate, e.g.

```
/exchanges/{exchange-id}/{transaction-uuid}:
    get:
      summary: Gets information related to an existing exchange.
      operationId: getExchangeData 
```

and 

```
/exchanges/{exchange-id}:
    get:
      summary: Gets information related to a exchange(s) identified by the exchange id.
      operationId: updateExchangeData
      parameters:
         - in: path
           name: query
           schema:
             type: any
           required: false    
           description: JSON query by example for filtering exchanges matching the id based on property values
```

and an explicit way to delete and exchange (note, in practice this may be a soft delete for audit purposes)

```
/exchanges/{exchange-id}/{transaction-uuid}:
    delete:
      summary: Deletes an exchange as specified by transaction-uuid.
      operationId: deleteExchangeData 
```

msporny: > ```
> /exchanges/{exchange-id}:
>     post:
>       summary: Creates an exchange as specified by type identified in exchange-id.
>       operationId: createExchangeData
> ```

I'm good with this, let's do this in this PR.

> then onto the `PUT` etc.
> 
> ```
> /exchanges/{exchange-id}/{transaction-uuid}:
>     put:
>       summary: Updates information related to an existing exchange.
>       operationId: updateExchangeData
> ```

I think I'm good with this, let's do this in this PR. I have to go back to our engineering team to see if they see any issues here.

> There likely should be defined corresponding getters as well, likely allowing query by example if appropriate, e.g.
> 
> ```
> /exchanges/{exchange-id}/{transaction-uuid}:
>     get:
>       summary: Gets information related to an existing exchange.
>       operationId: getExchangeData 
> ```

This one is a bit strange for DB's use cases, but we can probably live with it. The first concern that popped into my mind is one of authentication to do the GET. VP can be used to authn (and is in some of DB's use cases)... but if you can't POST/PUT that, then you need some other form of authn (presumably OAuth2, HTTP Signatures, zcaps, etc.).

I'd like to pursue this in another PR (ideally, raised by @mprorock), but not philosophical objections at this point in time.

> and (edit to note: i put the query in as a path parameter mongo style, but this can be handled other ways as well)

Hrm, what do you mean by this? I think you mean "don't use query parameters, put the query parameter into the path"?

> ```
> /exchanges/{exchange-id}:
>     get:
>       summary: Gets information related to a exchange(s) identified by the exchange id.
>       operationId: updateExchangeData
>       parameters:
>          - in: path
>            name: query
>            schema:
>              type: any
>            required: false    
>            description: JSON query by example for filtering exchanges matching the id based on property values
> ```

Confused about this one, need an example to understand more at depth. Suggest we do this one in a separate PR.

> and an explicit way to delete and exchange (note, in practice this may be a soft delete for audit purposes)
> 
> ```
> /exchanges/{exchange-id}/{transaction-uuid}:
>     delete:
>       summary: Deletes an exchange as specified by transaction-uuid.
>       operationId: deleteExchangeData 
> ```

No philosophical objections to this one. Let's do this in another PR.

OR13: @mprorock in 1 sentence what does `{transaction-uuid}` mean to you?

mprorock: the {transaction-uuid} actual unique identifier that represents the item in question - in this case a specific transaction as a part of a particular exchange

mprorock:
@msporny want to trim out the contentious bits so we can merge the good, agreed on stuff that is in here, and then I can PR separately around stateful stuff like getters/delete, etc ?

msporny:
> @msporny want to trim out the contentious bits so we can merge the good, agreed on stuff that is in here, and then I can PR separately around stateful stuff like getters/delete, etc ?

What are the contentious bits? I need a list that both @OR13, @mprorock, @mkhraisha, and I can agree to. :) -- in the very worst case, we'll spend VC API time today discussing what that is.

holder.yml
mkhraisha: > This PR has gotten to be a bit of a mess with comments and suggestions, but rescanning through to make sure i didn't miss any key points. focusing on `holder.yml` for these suggestions
> 
> the create a new one of type looks workable, though as @OR13 we may want to explicitly define some types for the `POST` body, I understand the use case for @msporny though, and would like to see if we can ideally satisfy a desire for stronger typing and a more flexible use case
> 
> ```
> /exchanges/{exchange-id}:
>     post:
>       summary: Creates an exchange as specified by type identified in exchange-id.
>       operationId: createExchangeData
> ```
> 
> then onto the `PUT` etc.
> 
> ```
> /exchanges/{exchange-id}/{transaction-uuid}:
>     put:
>       summary: Updates information related to an existing exchange.
>       operationId: updateExchangeData
> ```
> 
> There likely should be defined corresponding getters as well, likely allowing query by example if appropriate, e.g.
> 
> ```
> /exchanges/{exchange-id}/{transaction-uuid}:
>     get:
>       summary: Gets information related to an existing exchange.
>       operationId: getExchangeData 
> ```
> 
> and (edit to note: i put the query in as a path parameter mongo style, but this can be handled other ways as well)
> 
> ```
> /exchanges/{exchange-id}:
>     get:
>       summary: Gets information related to a exchange(s) identified by the exchange id.
>       operationId: updateExchangeData
>       parameters:
>          - in: path
>            name: query
>            schema:
>              type: any
>            required: false    
>            description: JSON query by example for filtering exchanges matching the id based on property values
> ```
> 
> and an explicit way to delete and exchange (note, in practice this may be a soft delete for audit purposes)
> 
> ```
> /exchanges/{exchange-id}/{transaction-uuid}:
>     delete:
>       summary: Deletes an exchange as specified by transaction-uuid.
>       operationId: deleteExchangeData 
> ```

This is exactly what i had envisioned

index.html
mkhraisha: Initiate exchange sounds pretty good to me

index.html
```diff
@@ -555,21 +555,21 @@ <h4>Prove Presentation</h4>
...
         <p>
         </p>
 
         <div class="api-detail"
-          data-api-endpoint="post /presentations/submissions"></div>
+          data-api-endpoint="post /exchanges/{exchange-id}/{transaction-uuid}"></div>
```
mkhraisha, line 572: I believe this needs to be updated to the put route

holder.yml
mkhraisha: @OR13 `exchange id` = type of exchange for example TransferOfOilCredential
`transaction-uuid` = instance of transferOfOilCredential Exchange

OR13: ahh to exchange id => definition, transaction uuid => instance... and both are current communicated via post body arguments... and this PR copies them into URL parameters... I assume we would error if they differed?

OR13:
Probably better to merge and have @mprorock  or @mkhraisha  do a follow up PR.

mprorock (approved):
approval contingent on follow up PRs to this

index.html
msporny: Done in a172d8118a70ae6d42e8eadd333727b151d6b4a3.

holder.yml
msporny: > does that match you understanding?

Yes, that matches my understanding and I now understand what you are saying. 

Based on the discussion during the telecon yesterday, I believe the group decided that we need to be very careful about this point. That is, if a server does not have control over what an `id` or `name` can be, we should not allow it to be a part of a path parameter (to avoid ugly URL encoding). I believe we agreed to that as a general rule and will need to determine how this fact affects how future path parameters are defined and used.

PR #262, opened 2022-02-11
Enable transaction identifiers to be multibase encoded 128-bit numbers.

msporny:
This PR expands the range of transaction identifiers to allow for multibase (base58btc and base64url) encoded 128-bit numbers.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/pull/262.html" title="Last updated on Feb 16, 2022, 11:05 PM UTC (d667a0c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/262/7b43c72...d667a0c.html" title="Last updated on Feb 16, 2022, 11:05 PM UTC (d667a0c)">Diff</a>

### DIFF
--- a/components/parameters/path/TransactionId.yml
+++ b/components/parameters/path/TransactionId.yml
@@ -1,0 +1,10 @@
+in: path
+name: transaction-id
+required: true
+schema:
+  type: string
+  pattern: "(z[1-9A-HJ-NP-Za-km-z]{21,22})|(u[a-zA-Z0-9_-]{22,23})|([0-9a-f]{8}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{12})"
+example:
+  - f37e5114-8b4a-11ec-b32e-fbd62b9502c1
+  - z4Z1T4HdLdfXCFAhdhzdCqm
+  - u7m8_ybOArXaEWeADqWCsIw
--- a/components/parameters/path/TransactionId.yml
+++ b/components/parameters/path/TransactionId.yml
@@ -2,9 +2,14 @@ in: path
 name: transaction-id
 required: true
 schema:
-  type: string
-  pattern: "(z[1-9A-HJ-NP-Za-km-z]{21,22})|(u[a-zA-Z0-9_-]{22,23})|([0-9a-f]{8}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{12})"
+  anyOf:
+    - type: string
+      pattern: "[0-9a-f]{8}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{12}"
+    - type: string
+      pattern: "z[1-9A-HJ-NP-Za-km-z]{21,22}"
+    - type: string
+      pattern: "u[a-zA-Z0-9_-]{22,23}"
 example:
   - f37e5114-8b4a-11ec-b32e-fbd62b9502c1
   - z4Z1T4HdLdfXCFAhdhzdCqm
-  - u7m8_ybOArXaEWeADqWCsIw
+  - u7m8_ybOArX-EWeADqWCsIw
--- a/holder.yml
+++ b/holder.yml
@@ -100,17 +100,17 @@ paths:
           description: Service not implemented.
         "500":
           description: Internal server error.
-  /exchanges/{exchange-id}/{transaction-uuid}:
+  /exchanges/{exchange-id}/{transaction-id}:
     put:
       summary: Receives information related to an existing exchange.
       operationId: receiveExchangeData
       description:
         A client can use this endpoint to continue the exchange of information
-        associated with an initiated exchange by sending a Verifiable Presentation 
+        associated with an initiated exchange by sending a Verifiable Presentation
         with information requested by the server to this endpoint.
       parameters:
         - $ref: "./components/parameters/path/ExchangeId.yml"
-        - $ref: "./components/parameters/path/TransactionUuid.yml"
+        - $ref: "./components/parameters/path/TransactionId.yml"
       requestBody:
         description:
           A Verifiable Presentation.
--- a/components/parameters/path/TransactionUuid.yml
+++ b/components/parameters/path/TransactionUuid.yml
@@ -1,6 +1,0 @@
-in: path
-name: transaction-uuid
-required: true
-schema:
-  type: string
-  pattern: "([0-9a-f]{8}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{12}){0,1}"
--- a/index.html
+++ b/index.html
@@ -533,7 +533,7 @@ The following APIs are defined for presenting a Verifiable Credential:
       <table class="simple api-summary-table"
         data-api-path="
           /credentials/derive /presentations/prove
-          /exchanges/{exchange-id} /exchanges/{exchange-id}/{transaction-uuid}"
+          /exchanges/{exchange-id} /exchanges/{exchange-id}/{transaction-id}"
         ></table>
 
       <section>
@@ -569,7 +569,7 @@ The following APIs are defined for presenting a Verifiable Credential:
         </p>
 
         <div class="api-detail"
-          data-api-endpoint="put /exchanges/{exchange-id}/{transaction-uuid}"></div>
+          data-api-endpoint="put /exchanges/{exchange-id}/{transaction-id}"></div>
       </section>
 
       <section>


holder.yml
```diff
@@ -100,17 +100,17 @@ paths:
           description: Service not implemented.
         "500":
           description: Internal server error.
-  /exchanges/{exchange-id}/{transaction-uuid}:
+  /exchanges/{exchange-id}/{transaction-id}:
```
OR13, line 103: I still don't understand or plan to use transaction id.

OR13 (approved):

mavarley:
I like the idea of not strictly requiring a uuid - but I'm wondering why we are being so prescriptive on an identifier; this seems like s server-side session management concern and not something that needs to be so strictly defined...

@brianorwhatever had a similar question on the call 08-02-2022. 

To @OR13 's point, providing guidance to ensure that identifiers are not other URLs (as this can lead to all sorts of other problems) can't we just specify a url safe regex?

components/parameters/path/TransactionId.yml
```diff
@@ -0,0 +1,10 @@
+in: path
+name: transaction-id
+required: true
+schema:
+  type: string
+  pattern: "(z[1-9A-HJ-NP-Za-km-z]{21,22})|(u[a-zA-Z0-9_-]{22,23})|([0-9a-f]{8}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{12})"
```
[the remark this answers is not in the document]
msporny: Agreed. I'll do that in my next pass.

holder.yml
msporny: > I still don't ... plan to use transaction id.

This is fine.

> I still don't ... understand transaction id.

This is concerning, my hope is that we can at least get you to understanding it so that you use it if it ever becomes useful to you.

Digital Bazaar finds transaction id useful to refer to any sort of stateful binding on the server side. For example, when the client is like: "I want to do an exchange of a specific type with you.", and the server responds with "Ok, here's a VPR containing a challenge", the server *might* need to store that challenge and information specific to that particular transaction server-side. It can store this indexed against a transaction-id (or something derived from the transaction-id... transaction-id + nonce, for example).

Does that make sense?

msporny:
@mavarley wrote:
> I like the idea of not strictly requiring a uuid - but I'm wondering why we are being so prescriptive on an identifier; this seems like s server-side session management concern and not something that needs to be so strictly defined...

@OR13 wanted us to be specific in a way that is testable. I believe everyone wants transaction IDs, and having per system transaction IDs that are larger than 128-bits doesn't make sense (in an information theoretical sense). So, the problem then became "How many different ways do we want to encode a transaction ID?". 

We can always start very strict and then relax the requirements to allow a greater range.

> To @OR13 's point, providing guidance to ensure that identifiers are not other URLs (as this can lead to all sorts of other problems) can't we just specify a url safe regex?

Yes, we could relax the requirements for transaction-id but no one has been able to provide a compelling argument for that other than "I want to be able to use something different." -- the question then becomes, "Ok, what else do you want to be able to do and what is your compelling reason for doing that?"

holder.yml
msporny: > @msporny so `transaction-id` is like `domain` in the VC Data Model?

Hrm, nope... the client sets domain, not the server. If the server sets domain, it can attack the client by replaying challenge/domain combinations that it's doing behind the scenes w/ other websites. That is, if we let the server set the domain, it's an attack vector for any client doing DID Auth.

OR13: client server language is problematic here... I think you mean that the `verifier` sets the `domain` and `challenge` for a presentation... and my question was, are either of those terms related to these new identifiers... seems like the answer is not, and you would expect a verifier to receive many presentations with a unique domain and challenge values to the same `transaction-id`.

OR13: in trace interop profile we use the term `definition` for `{exchange-id}` and `instance` for `{transaction-id}`... I wonder if its too late to change the names here to align with that...

```
{exchange-id} -> {definition-id}
```

```
{transaction-id} -> {instance-id}
```

msporny: > I think you mean that the verifier sets the domain and challenge for a presentation (and the holder signs over them)

No, I definitely don't mean that. That could become a security vulnerability. verifier app/service sets the challenge, holder app/service sets the domain and signs over both values.

> are either of those terms related to these new identifiers... seems like the answer is no

The answer isn't "no", it's: They might be associated in a 1-to-1 fashion, or they might be associated in an n-to-1 fashion.

> you would expect a verifier to receive many presentation exchanges with a unique domain and challenge values to the same transaction-id

Digital Bazaar doesn't have that use case currently, but we do admit that it's possible that some verifiers might structure their management of transaction-ids in this way. For example, if a single transaction requires N holders to submit VPs to a single transaction-id-identified endpoint.

OR13: > aud MUST represent (i.e., identify) the intended audience of the [verifiable presentation](https://www.w3.org/TR/vc-data-model/#dfn-verifiable-presentations) (i.e., the [verifier](https://www.w3.org/TR/vc-data-model/#dfn-verifier) intended by the presenting [holder](https://www.w3.org/TR/vc-data-model/#dfn-holders) to receive and verify the [verifiable presentation](https://www.w3.org/TR/vc-data-model/#dfn-verifiable-presentations)).

We should probably update the definition here: https://w3c-ccg.github.io/security-vocab/#domain

If a verifier suggests a domain and the holder signs a presentation over a different value, what happens?

peacekeeper:
> I'm wondering why we are being so prescriptive on an identifier; this seems like s server-side session management concern

I agree with @mavarley, I had the same thought. The identifier should be allowed to be any string and completely opaque to the client.

We have a very similar concept called `jobId` in a specification for "DID Registration" at DIF. In practice we use UUIDs there but we don't prescribe any specific format in the [specification](https://identity.foundation/did-registration/#jobid) or the [OpenAPI definition](https://github.com/decentralized-identity/universal-registrar/blob/main/swagger/api.yml#L197).

OR13:
https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html

please make `<script>alert(1);</script>` a "spec legal id", I can't wait to read about this on twitter.

peacekeeper:
One advantage of leaving the format undefined is that implementations can potentially encode state in the transactionId itself. This "state inside the transactionId" would be opaque to the client but can reduce the resources needed for state tracking on the server side.

That's what happened historically with OAuth2 access tokens. They are considered to be opaque to a client, and early deployments of OAuth2 used random strings as access tokens which were essentially lookup keys for databases on the server side. But then implementers decided that they could also use JWTs as access tokens which eliminates or reduces the need for a state database on the server side.

Just something to consider, I don't feel strongly about it. I can also see advantages of prescribing a specific format, if this is strictly meant to be an "ID", similar to a thread ID in DIDComm.

OR13:
> This "state inside the transactionId" would be opaque to the client but can reduce the resources needed for state tracking on the server side.

You can do this and still restrict legal characters.

I can also imagine encoding state like this as a clever way to track folks...

```
transactionId = AES( SSN and other PII, badActorKey ).toBase64();
```

mavarley:
> please make c:\$MFT\123 and <script>alert(1);</script> "spec legal id", I can't wait to read about this on twitter.

> I can also imagine encoding state like this as a clever way to track folks...

These are exactly the type of requirements/anti-requirements/best practices I was hoping to draw out ;) 

To @peacekeeper 's point, opaque tokens have been very useful in the OAuth world, allowing for a variety of solution across distributed systems - signed tokens can be verified w/o expensive callbacks to introspection endpoints etc... But `transaction-id` is not meant to be a security token (true?) and misuse of a base64url string can cause more problems then not... but I think it is important to get these concepts out in the open rather than saying "uuid".

For example, if encoding PII is a real concern for this API, doesn't that mean we should _not_ allow multi-base encoded 128-bit numbers? it could be an SSN!

For the record, I think the above concern is a bit too far; bad actors can be bad actors regardless of `uuid`. But url safe encoded strings of a max length might be the right compromise? Or we provide some verbiage around why uuid and multibase encoding (maybe with some rules?) are the right choices here and move along? I feel like we need to add a "why", and am willing to help structure that text, but I'm still not 100% clear why. Thanks.

mprorock (approved):

peacekeeper (approved):

components/parameters/path/TransactionId.yml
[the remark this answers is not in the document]
msporny: Done in d667a0cf91a9bcf3e891db2c1c1105dc27a143ab.

holder.yml
msporny: > AFAIK, there is no mutual authentication happening here

You could argue that mutual auth /is/ happening *if* the client is paying attention to TLS (which is almost certainly is doing these days)... and the server is ensuring the DID Auth response is associated with a specific DID.

> and domain and challenge don't actually protect against anything other than replay attacks on the verifiers web server. 

If the client sets the `domain`, then the client is protecting itself from replays across the entirety of the Internet (if the domain value maps to DNS) and as long as the server is protecting itself from replays by paying attention to the domain value.

> (assuming that challenges are stateful and consumed, which also implies a DDoS vector)

Yep, it is a DDoS vector if the server doesn't follow normal DDoS-protection hygiene (limit requests by IP, HTTP Header, challenge-under-attack, or any other form of preliminary client identification).

PR #271, opened 2022-03-13
Initial GET and DELETE Operations

mprorock:
The bare minimum get and delete operations for credentials and presentations - putting this in for feedback and once merged we can extend the pattern as required.  in particular - i think we will want allow a PUT to update and / or replace items by ID

NOTE: Depends on pr #270 and should not be merged until that is resolved


msporny:
> NOTE: Depends on pr #270 and should not be merged until that is resolved

Marking 'do not merge' temporarily so no one accidentally merges this before #270.

mprorock:
@msporny i am inclined to set "id" for getters to be UUID or a URI - does this make sense to you?

msporny:
> @msporny i am inclined to set "id" for getters to be UUID or a URI - does this make sense to you?

I'm fine to start there. I expect that we'll quickly move towards something that looks more like 'transaction-id' in time. Every time we try to pick UUID, it turns into "UUID and multi-base 128 bit value", which then turns into "but I want to encode JWT-like objects in the ID, so give me 1024 URL-safe characters to play around with".

I get why we'd want URIs for GET... shudder a bit at having to URIEncode those. No strong feelings, but again, feels like "id"s are just going to turn into "1024 URL-safe characters" over time. Just my $0.02 for now.

mprorock:
> just going to turn into "1024 URL-safe characters" over time. Just my $0.02 for now.

yeah - i suspect you are right for top level API and then i also expect profiles to tighten that down to their specific requirements - e.g. i would bet the traceability goes with just UUID for a variety of reasons

OR13 (approved):

peacekeeper:
I think this is useful, but GET and DELETE operations should probably be optional. There may be implementations that can issue VCs but then don't remember them.

msporny:
This needs to be refactored on top of main. I suggest cherry picking the four commits instead of rebasing -- I cleaned up the history to remove the add/removal of package-lock.json as well as the auto-generated files in the commit history.

OR13:
There are conflicts

msporny:
> Conflicts should now be resolved

Still showing conflits with rebase and merge. I can squash-merge if you'd like, but would prefer to keep your history.

OR13:
Screen Shot 2022-03-29 at 8 58 08 AM

The other code owners have not reviewed.

holder.yml
```diff
@@ -10,10 +10,106 @@ info:
...
+      tags:
+       - Credentials
+      summary: Gets a credential or verifiable credential by ID
+      operationId: getCredential
+      parameters:
+        - in: path
+          name: id
+          schema:
+            type: string
+            format: uuid
```
msporny, line 24: Hrm, disagree w/ this only being able to be a UUID -- but we can deal w/ that in another PR. Just noting my objection to this one design choice.

holder.yml
```diff
@@ -10,10 +10,106 @@ info:
...
+                  - $ref: "./components/Credential.yml#/components/schemas/Credential"
+                  - $ref: "./components/VerifiableCredential.yml#/components/schemas/VerifiableCredential"
+        "400":
+          description: Bad Request
+        "401":
+          description: Not Authorized
+        "410":
+          description: Gone! There is no data here
+        "418":
+          description: I'm a teapot - MUST not be returned outside of pre-arranged scenarios between both parties
```
msporny, line 41: Har har :) -- presuming we're going to eventually remove this... because if we don't, we'll have to test for it.

holder.yml
```diff
@@ -10,10 +10,106 @@ info:
...
+      operationId: deleteCredential
+      parameters:
+        - in: path
+          name: id
+          schema:
+            type: string
+            format: uuid
+      responses:
+        "202":
+          description: Credential deleted - this is a 202 by default as soft deletes and processing time are assumed
```
msporny, line 59: This should probably say "Credential marked for deletion". Looks like the proposal here is that this call is sometimes implemented as eventually consistent.

msporny (approved):
Ok to merge as a work in progress. I've noted a few things that are concerning with the PR, but good enough for now.

msporny:
This is what I see: 

image

holder.yml
OR13: @mprorock we can take this change in our profile and leave this repo with `type` string where `<script> alert(1);</script>` is a spec legal identifier.

OR13, line 24:
```suggestion
```

OR13: Lets make sure to allow xss / command injection in our identifiers.

holder.yml
```diff
@@ -10,10 +10,106 @@ info:
...
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: "./components/Credential.yml#/components/schemas/Credential"
+                  - $ref: "./components/VerifiableCredential.yml#/components/schemas/VerifiableCredential"
+        "400":
+          description: Bad Request
+        "401":
+          description: Not Authorized
```
OR13, line 37: are we going to define authorization? and test it?

holder.yml
```diff
@@ -10,10 +10,106 @@ info:
...
+      parameters:
+        - in: query
+          name: type
+          schema:
+            type: array
+            items:
+              type: string
+              oneOf:
+                - "credentials"
+                - "verifiablecredentials"
```
OR13, line 85:
```suggestion
                - "verifiable-credentials"
```

holder.yml
msporny: > leave this repo with `type` string where `<script> alert(1);</script>` is a spec legal identifier.

Or, you know, do something more sane and re-use our existing transactionId/exchangeId regexes, which don't allow the attack you mention above:

https://github.com/w3c-ccg/vc-api/blob/main/components/parameters/path/TransactionId.yml

holder.yml
msporny: > are we going to define authorization? and test it?

The test suites will need to have some authz mechanism in them because the production software that all implementers are implementing are all (AFAIK) requiring authz. If we have interop on the test suites, the spec will have to reflect reality and document how authz is being performed in an interoperable fashion.

holder.yml
msporny: Why kebab-case instead of camelCase? We're using camelCase everywhere else, no?

msporny (changes requested):
Hrm, just spotted the two massive auto-generated files. We have Github Actions to auto-generate files, we should not commit autogenerated files.

holder.yml
OR13: its a query string param,  and its pluralized... so i figured it would be better to not have it look like `verifiableCredential` which shows up in VPs, but I am not attached to anything other than fixing the lowercase concatenation issue.

OR13 (changes requested):
remove auto generated files

holder.yml
mprorock: +1 manu - added an `ObjectId` that matches that transaction id

mprorock:
> Hrm, just spotted the two massive auto-generated files. We have Github Actions to auto-generate files, we should not commit autogenerated files.

that should be fixed now - thanks for catching that

OR13 (approved):

msporny:
> ahh thats because you are using rebase... you should stop using rebase since it makes it harder for non technical git users to contribute ; )

When there are non-technical Github users involved, I use squash. 

This is not one of those cases :), @mprorock knows what he's doing. I consider it bad form to squash merge contributor's commits if it looks like they've gone to the trouble of creating a clean history. In those cases, I ask them if they're ok with it before doing it.

holder.yml
```diff
@@ -10,10 +10,98 @@ info:
...
+      summary: Gets list of credentials or verifiable credentials
+      operationId: getCredentials
+      parameters:
+        - in: query
+          name: type
+          schema:
+            type: array
+            items:
+              type: string
+              oneOf:
```
msporny, line 75: These options seem a bit arbitrary... fine for now, but we need to define exactly what should be happening when each option is provided... for example, is this really checking the `type` field, or is it doing something else -- my guess is the latter... while the former might be more useful?

msporny (approved):
LGTM. @mprorock, do I have your permission to squash-merge this PR?

mprorock:
> LGTM. @mprorock, do I have your permission to squash-merge this PR? Or, would you like to rebase so I can rebase-merge? Note: If you rebase, please just make it so that the auto-generated files don't show up in history at all... I had to go in and surgically remove auto-gen'd files from history the last time.

personally I am a squash and merge fan

PR #315, opened 2022-10-20
Add 404 to exchanges/exchange-id/transaction-id responses.

aljones15:
Addresses: https://github.com/w3c-ccg/vc-api/issues/305

This a DRAFT because the description needs bikeshedding.

### DIFF
--- a/holder.yml
+++ b/holder.yml
@@ -376,7 +376,7 @@ paths:
         "500":
           description: Internal server error.
   /exchanges/{exchange-id}/{transaction-id}:
-    put:
+    post:
       summary: Receives information related to an existing exchange.
       tags:
        - Exchanges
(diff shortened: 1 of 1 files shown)


holder.yml
```diff
@@ -408,6 +408,9 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorResponse"
+        "404":
+          description: Either the exchange-id, transaction-id, or something related
+                       to the request was not found.
```
dlongley, line 413:
```suggestion
          description: The associated exchange or transaction was not found.
```

dlongley: Also, the method needs to change from `put` to `post` above.

dlongley (approved):

msporny (approved):

PR #316, opened 2022-10-20
Add OAuth2 to securitySchemes for holder, issuer, & verifier.

aljones15:
This borrows the `OAuth2` stuff from traceability and abstracts it into the `VC-API`.

This adds `OAuth2` to all of the paths in issuer.

This is a DRAFT PR for a variety of reasons:

- [x] the auth details don't show up in the `index.html` and most of the `api/*.html` pages
- [x] We probably want to add `OAuth2` to other paths/ routes, but I'm not sure if it goes on all of them
- [ ] We might want to add a `Zcap` auith component too

### DIFF
--- a/components/oauth2.yml
+++ b/components/oauth2.yml
@@ -1,0 +1,19 @@
+openapi: 3.0.0
+info:
+  version: "0.0.3-unstable"
+  title: VC API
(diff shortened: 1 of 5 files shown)


dlongley:
Closing this PR because it doesn't represent the consensus of the group.

PR #323, opened 2022-10-31
Add verifiablePresentationRequest property to exchange initiate.

aljones15:
This matches the response schema for `exchanges initiate` so that it matches the examples.
The change made is that the `exchanges initiate` response now has a property: `verifiablePresentationRequest`.
This is a DRAFT PR as discussion might need to be made on if the examples are wrong or the schema is wrong.

### DIFF
--- a/holder.yml
+++ b/holder.yml
@@ -526,7 +526,10 @@ components:
     StorePresentationRequest:
       $ref: "./components/VerifiablePresentation.yml#/components/schemas/VerifiablePresentation"
     VerifiablePresentationRequest:
-      $ref: "./components/VerifiablePresentationRequest.yml#/components/schemas/VerifiablePresentationRequest"
+      type: object
+      properties:
+        verifiablePresentationRequest:
+          $ref: "./components/VerifiablePresentationRequest.yml#/components/schemas/VerifiablePresentationRequest"
     VerifiablePresentationResponse:
       type: object
       properties:


holder.yml
```diff
@@ -526,7 +526,10 @@ components:
     StorePresentationRequest:
       $ref: "./components/VerifiablePresentation.yml#/components/schemas/VerifiablePresentation"
     VerifiablePresentationRequest:
-      $ref: "./components/VerifiablePresentationRequest.yml#/components/schemas/VerifiablePresentationRequest"
+      type: object
+      properties:
+        verifiablePresentationRequest:
+          $ref: "./components/VerifiablePresentationRequest.yml#/components/schemas/VerifiablePresentationRequest"
```
dlongley, line 532: Is this circular? It seems like the schema named `VerifiablePresentationRequest` should just be a VPR -- and that there should be another named schema that has `verifiablePresentationRequest` as a property that references it. I might just not understand how this language works -- if so, please ignore.

aljones15: The pattern with the other schemas is this: https://github.com/w3c-ccg/vc-api/blob/f12d687b6bfecbfc432985e66e03177798c2b55a/holder.yml#L524-L534

So it does look like the property should be moved inside of the schema in `components`

aljones15: Addressed here: https://github.com/w3c-ccg/vc-api/pull/323/commits/0a8bb1f33f3b005ead8f7dec7bcb17c77f7f6b4b

msporny (approved):

holder.yml
dlongley: A VPR doesn't have a `verifiablePresentationRequest` inside of it -- a request body to certain VC-API endpoints does. This is what makes me think this isn't modeled quite right.

IMO, it's that request body that needs a name and that should have a single property of `verifiablePresentationRequest` with a schema value that references `VerifiablePresentationRequest`. That's also how it works with `VerifiableCredential` elsewhere: https://github.com/w3c-ccg/vc-api/blob/e48c5c1721efe82157d9f174513e8018ecb14a14/issuer.yml#L70-L72

The group should feel free to merge over my objections, though, if I just continue to misunderstand how this works.

mkhraisha: I agree with your point Dave, it doest make sense to have a `verifiablePresentationRequest` object in the VPR

msporny (changes requested):
Yes, agreed, this PR is wrong. We shouldn't have a `verifiablePresentationRequest` in `components/VerifiablePresentationRequest.yml`

msporny:
@aljones15 ETA on fixing up this PR?

dlongley (approved):

PR #326, opened 2022-12-30
Reference error schema on error endpoints

ottonomy:
A useful feature at this point would be to show the error responses and examples in the OpenAPI file.

Went to add an error example to the first endpoint I noticed needed one and encountered the additional oddity: looking at `GET /credentials/{id}` it is strange to have a 400 but not a 404 response. 400 usually indicates an error in the request (specifically, in the body), but there is no body in the request for this GET request, so the `{id}` parameter in the path is the only thing that the client could get wrong somehow. Probably a 404 instead of a 400 is a more relevant response.

How does this look at adding a schema reference to the ErrorResponse and example response id/message that would be applicable to the type of error received? 

Edit... improving PR, my editor formatted for consistency on save, yielding many additional changed lines beyond minimal. Unintentional changes fixed in second commit.

### DIFF
--- a/components/ErrorResponse.yml
+++ b/components/ErrorResponse.yml
@@ -27,8 +27,8 @@ components:
           maxLength: 100
         details:
           type: "object"
-          description: Further machine-readable details about the error.
-      required: ["id", "message"]
+          description: An object with error details.
+      required: ["name", "message"]
       example:
         {
           "name": "Invalid JSON",
--- a/components/ErrorResponse.yml
+++ b/components/ErrorResponse.yml
@@ -16,18 +16,21 @@ components:
       type: object
       description: A response that denotes that an error has occurred.
       properties:
-        id:
+        name:
           type: "string"
+          description: A brief descriptive name for the error.
           pattern: "[a-z0-9\\-]{8,}"
         message:
           type: "string"
+          description: Human-readable detail about the error.
           minLength: 10
           maxLength: 100
         details:
           type: "object"
+          description: Further machine-readable details about the error.
       required: ["id", "message"]
       example:
         {
-          "id": "invalid-json",
+          "name": "Invalid JSON",
           "message": "The provided JSON data was malformed."
         }
--- a/holder.yml
+++ b/holder.yml
@@ -13,7 +13,7 @@ paths:
   /credentials/{id}:
     get:
       tags:
-        - Credentials
+       - Credentials
       summary: Gets a credential or verifiable credential by ID
       operationId: getCredential
       parameters:
@@ -27,6 +27,8 @@ paths:
                 oneOf:
                   - $ref: "./components/Credential.yml#/components/schemas/Credential"
                   - $ref: "./components/VerifiableCredential.yml#/components/schemas/VerifiableCredential"
+        "400":
+          description: Bad Request
         "401":
           description: Not Authorized
         "404":
@@ -48,7 +50,7 @@ paths:
           description: Not Implemented
     delete:
       tags:
-        - Credentials
+       - Credentials
       summary: Deletes a credential or verifiable credential by ID
       operationId: deleteCredential
       parameters:
@@ -69,7 +71,7 @@ paths:
   /credentials:
     get:
       tags:
-        - Credentials
+       - Credentials
       summary: Gets list of credentials or verifiable credentials
       operationId: getCredentials
       parameters:
@@ -105,7 +107,7 @@ paths:
   /credentials/derive:
     post:
       tags:
-        - Credentials
+       - Credentials
       summary: Derives a credential and returns it in the response body.
       operationId: deriveCredential
       description: Derives a credential and returns it in the response body.
@@ -132,7 +134,7 @@ paths:
   /presentations/{id}:
     get:
       tags:
-        - Presentations
+       - Presentations
       summary: Gets a presentation or verifiable presentation by ID
       operationId: getPresentation
       parameters:
@@ -158,7 +160,7 @@ paths:
           description: Not Implemented
     delete:
       tags:
-        - Presentations
+       - Presentations
       summary: Deletes a presentation or verifiable presentation by ID
       operationId: deletePresentation
       parameters:
@@ -179,7 +181,7 @@ paths:
   /presentations:
     get:
       tags:
-        - Presentations
+       - Presentations
       summary: Gets list of presentations or verifiable presentations
       operationId: getPresentations
       parameters:
@@ -217,7 +219,7 @@ paths:
     post:
       summary: Proves a presentation and returns it in the response body.
       tags:
-        - Presentations
+       - Presentations
       operationId: provePresentation
       description: Proves a presentation and returns it in the response body.
       requestBody:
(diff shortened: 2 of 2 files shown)


aljones15:
+1 to standardized errors, but I do think it might be better to figure out a standardized set of error codes or ids.
DID Resolution uses these codes for instance: https://w3c-ccg.github.io/did-resolution/#invaliddid

So from the test perspective I think the group needs to decide if errors should have an `id` or `code` or both or neither. As code is fairly standard for errors in multiple languages we should probably stick with code, but I'm not a string participant in this specification.

msporny:
@ottonomy wrote:
> I admit that this is a bit verbose here

Yes, it's a side effect of the respec plugin I wrote to convert OAS files into W3C ReSpec language. It needs work and I suggest we ignore the ugliness for now. The important bit is the use of the standard error object, which I'm +1 to. I agree with @ottonomy that we've picked some strange return values in certain sections of the specification and that those need to be normalized/fixed.

msporny (approved):
If we adopt this PR, we should probably just blanket use the ErrorResponse everywhere (with examples). We need to be consistent and it's probably a good idea to depend on more than just a response code when things go wrong.

I'd like to get a few more implementer's thoughts on this before proceeding with the merge... and if we merge, we should follow it up with updates to all errors.

msporny:
The group discussed this on 2023-01-10. @dlongley noted that he was ambivalent about whether or not everything has an error response. If there are values that need to be sent back to be acted upon vs. all information being communicated via status -- that seems to be the most important thing. Ideally, we don't invent our own thing for error codes or the error data model -- there is some RFC that Mark Nottingham might have put together for HTTP Errors, don't know if anyone uses it, don't know if it fits our use cases, might not match up with how errors are returned from some implementations today -- some implementations use `name` field to match JavaScript errors, might be problematic for a W3C spec -- a number of considerations. @jandrieu asked "what is the framework for extensibility in the responses?" -- there might be open ended things to pass back. @msporny noted that current error response has `id` which is mandatory, `description` which contains explanatory text, and `details` which can contain arbitrary infomation from the implementation. @dlongley noted that `id` is problematic and maybe we should use `name` instead. @jandrieu agreed. 

Do we want to always return an error object on an error? @TallTed noted mandating minimal requirements but allowing for larger responses -- as a troubleshooter, want detailed errors, could be configurable/switchable. That is implementation specific. Leave the door open for implementations to provide sufficient information to troubleshoot information that's erroring. @dlongley noted that we shouldn't ALWAYS return an error code until we it's clear that we should. We need to find cases where HTTP status code is insufficient and mandate detailed errors through that mechanism, but only when necessary. An implementation is allowed to return an ErrorResponse if it chooses to do so, and those errors might not be the same across implementations UNLESS it is specified to be the same by the specification. @PatStLouis noted that covering all status codes is a must, two different errors could return the same error code, would be good to know the reason for it ... for cases where implementer wants to return more detail, good to have a pre-defined schema, but leave content to implementer.

The group arrived at consensus on the following:

1. Errors will, at a minimum, be conveyed using HTTP error response codes.
2. Implementers MAY return error responses using the ErrorResponse schema, or their own, (for any HTTP error).
3. The specification MAY (at a later date) define specific ErrorResponses in specific situations.

@dlongley NOTE: We might want to just say error response is JSON and not use a suggested schema and that can be bike shedded in the PR.

@PatStLouis asked if we wanted to have a negative test and analyze response to match the test, might be easier to do if there is a pre-defined error schema. @dlongley noted that if we are going to do that, we might mandate that you have to have something in the spec -- so, either test suite is going to look for HTTP status codes, or we find cases where we know where we want the errors to be and mandate those. We shouldn't do something in between that and it becomes "de-facto".

ottonomy:
Thanks for discussing this. I'll try to join upcoming calls to continue working on it as I consider how to implement some use cases (for complex authorization scenarios on the issue endpoint -- the spec is right to mostly stay out of what authorized scopes may be granted or checked, but I suspect that there are lots of use cases for issuer coordinator applications that are not implicitly broadly trusted to trigger signing of any credential on an issuer's behalf. In some cases, client side differentiation of what specifically has not been authorized about a particular award request is of interest to me).

I like, from what I've heard discussed:

- The ability to respond at a minimum with the status code only (no response body). In most cases, no need for us to overcomplicate things by requiring more than this.
- A standardized error schema.
- We might take on as a work item the enumeration of some error codes that might appear in a `name` or `code` or similar parameter of the error schema. When a server wants to include additional details, these codes may go make the specific scenario more clear to a client.
- If we test error responses (expecting a certain error response), most test use cases might be satisfied by checking the response's HTTP status code (though if a body is returned it should at least conform to the schema). Some tests might benefit from checking the HTTP status and a specific error code in the body. But tests should follow from specific mandates in the spec.

msporny:
The group discussed this on 2023-01-24, there was concern about not having use cases yet. @ottonomy noted a few use cases that were useful. @dlongley noted that the `id` field in the error schema was problematic and wanted to just go with named errors. It was raised that we should get @mprorock, @mkhraisha, and @OR13's input on this from the traceability side on their desire/need for error schemas. 

Next steps: Here is an API, when it throws an error, we need X information back from it, and here is the information that we need back to determine if approach works across API calls. Other comment is "light on authorization", we do have something to say which endpoints are authz'd and which ones are not... depending on authz used, there might be standardized error codes already, so might not be that we would define something here other than use another spec.

@ottonomy noted that there are use cases that should be denied based on authz vs. authn. There is a clear different status level vs. auth header vs. auth token included was invalid, 401 vs permissions based authz failure. You cannot award a VC that is "blue", you can only issue "red" credentials. The spec is correct to be quiet about authz errors, not scope of spec, but interesting to have software that can enforce authz rules on endpoints, so want to be able to express error messages in a predictable way. @dmitrizagidulin noted that some authz methods used like OAuth2 dictate their own error format, so might be useful to follow error format or have it isomorphic -- OAuth2 defines "error/error_description/error_uri/name/status_code" as mentioned.

@ottonomy noted that he can put some time into this over the next couple of weeks, could take a stab at adding use cases into the document. Could also examine optional response body for OAS.

OR13:
@nissimsan any thoughts on VC API errors?

nissimsan:
In trace-interop we also decided to [loosen up the error response schema](https://github.com/w3c-ccg/traceability-interop/pull/319). The idea of an error code also surfaced, but we [got rid of that as well](https://github.com/w3c-ccg/traceability-interop/pull/413). This discussion seems get to the same conclusion: avoid unnecessary schema complications. 

Noting that our error-discussions have mainly been around security (`401` and `403`) and `POST /credentials/issue`'s potentially bad issuer (`422`, see https://w3c-ccg.github.io/traceability-interop/openapi/#post-/credentials/issue) - I don't remember us having talked much specifically about `GET /credentials/{id}`.

ottonomy:
I made progress on a couple action items:

### 1: attempt to describe how a content body may be optional.

This appears to be something that OpenAPI does not handle well. It assumes that an endpoint will be defined to have `responses.content` or that it will be defined to not have that, not that it could be defined with an optional response body.

I committed an approach for consideration, but I don't consider it great. Below, these cases appear:
* 200: there is response content in `application/json` media type
* 400, 401: there is no content
* 404 (the one I've been tweaking as an example): there is an `application/json` media type with a JSON response in the `ErrorResponse` schema, or "other" (`*/*` with no schema, and an example of an empty string). 

```yaml
responses:
        "200":
          description: Credential retrieved
          content:
            application/json:
              schema:
                oneOf:
                  - $ref: "./components/Credential.yml#/components/schemas/Credential"
                  - $ref: "./components/VerifiableCredential.yml#/components/schemas/VerifiableCredential"
        "400":
          description: Bad Request
        "401":
          description: Not Authorized
        "404":
          description: Not Found
          content:
            '*/*':
              example: ''
            application/json:
              schema:
                $ref: "./components/ErrorResponse.yml#/components/schemas/ErrorResponse"
              example:
                name: Not found
                message: The requested credential with this ID was not found.
```

### Action item 2: update the ErrorResponse schema to avoid using `id` (or `code`), which would have made people think it was coming from a controlled vocabulary. Renaming this short label property to `name`, which people would not assume would be restricted to a controlled vocabulary.

```yaml
  schemas:
    ErrorResponse:
      type: object
      description: A response that denotes that an error has occurred.
      properties:
        name:
          type: "string"
          description: A brief descriptive name for the error.
          pattern: "[a-z0-9\\-]{8,}"
        message:
          type: "string"
          description: Human-readable detail about the error.
          minLength: 10
          maxLength: 100
        details:
          type: "object"
          description: An object with error details.
      required: ["name", "message"]
      example:
        {
          "name": "Invalid JSON",
          "message": "The provided JSON data was malformed."
        }

```

ottonomy:
On the call today, we resolved to close this PR and replace it with 2 actionable PRs, for each of the above action items I researched:
* I will make a PR to remove the specification of response content for unsuccessful status codes, replacing that with spec text describing that the response should conform to the `ErrorResponse` schema if content is returned. 
* I will make a PR to clarify the ErrorResponse schema roughly matching the above example, except that the `pattern` should be removed from `name` and minLength, maxLength should be removed from `message`, relying on clear textual description that `name` should be short, `message` can be longer.

msporny:
Closing per @ottonomy's https://github.com/w3c-ccg/vc-api/pull/326#issuecomment-1421391501.

PR #334, opened 2023-01-23
Fix linting errors in OAS files.

msporny:
This PR fixes linting errors in the OAS files. OAS linting requires that every endpoint define security properties and at least one server endpoint. Reviewers should pay specific attention to this PR because it does add authorization mechanisms available on each endpoint, including noAuth, didAuth, oAuth2, and zCap (since the current set of implementers support at least one of those mechanisms, if not multiple). Setting this PR to draft so we can discuss.

The intent of this PR is NOT to establish the correct authz mechanisms for each endpoint, that will be done in a future PR. The goal here is to get "close enough" and then fine tune the authz mechanisms in a future PR. Do not take the authz mechanisms provided as anything more than a work in progress at this point in time.

### DIFF
--- a/components/SecuritySchemes.yml
+++ b/components/SecuritySchemes.yml
@@ -1,0 +1,39 @@
+openapi: 3.0.0
+info:
+  version: "0.0.3-unstable"
+  title: VC API
+  description: This is an Experimental Open API Specification for the [VC Data Model](https://www.w3.org/TR/vc-data-model/).
+  license:
+    name: W3C Software and Document License
+    url: http://www.w3.org/Consortium/Legal/copyright-software.
+  contact:
+    name: GitHub Source Code
+    url: https://github.com/w3c-ccg/vc-api
+paths:
+components:
+  securitySchemes:
+    noAuth:
+      type: http
+      scheme: bearer
+      bearerFormat: VerifiablePresentation
+
+    zCap:
+      type: http
+      scheme: bearer
+      bearerFormat: VerifiablePresentation
+
+
+    didAuth:
+      type: http
+      scheme: bearer
+      bearerFormat: VerifiablePresentation
+
+    oAuth2:
+      type: oauth2
+      flows:
+        authorizationCode:
+          authorizationUrl: /oauth2/authorize
+          tokenUrl: /oauth2/token
+          scopes:
+            read: Grants read access
+            write: Grants write access
--- a/components/SecuritySchemes.yml
+++ b/components/SecuritySchemes.yml
@@ -15,13 +15,12 @@ components:
     noAuth:
       type: http
       scheme: bearer
-      bearerFormat: VerifiablePresentation
+      bearerFormat: emptyString
 
     zCap:
       type: http
-      scheme: bearer
-      bearerFormat: VerifiablePresentation
-
+      scheme: signature
+      bearerFormat: AuthorizationCapability
 
     didAuth:
       type: http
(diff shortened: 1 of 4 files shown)


components/SecuritySchemes.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    oAuth2:
+      type: oauth2
+      flows:
+        authorizationCode:
+          authorizationUrl: /oauth2/authorize
+          tokenUrl: /oauth2/token
+          scopes:
+            read: Grants read access
+            write: Grants write access
```
msporny, line 38:
```suggestion
          authorizationUrl: /TBD
          tokenUrl: /TBD
          scopes:
            TBD: There are currently no scopes defined.
```

components/SecuritySchemes.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    didAuth:
+      type: http
+      scheme: signature
+      bearerFormat: VerifiablePresentation
+
+    oAuth2:
+      type: oauth2
+      flows:
+        authorizationCode:
```
msporny, line 33: You don't have to use authorizationCode -- need to figure out how to say "this is up to implementers".

components/SecuritySchemes.yml
```diff
@@ -0,0 +1,38 @@
...
+  license:
+    name: W3C Software and Document License
+    url: http://www.w3.org/Consortium/Legal/copyright-software.
+  contact:
+    name: GitHub Source Code
+    url: https://github.com/w3c-ccg/vc-api
+paths:
+components:
+  securitySchemes:
+    noAuth:
```
msporny, line 15:
```suggestion
    firewall:
```

ottonomy: As noted on call, we understand the scope of this PR is to fix lint errors (we never want to compromise about dev tools not being broken), but we will update in future PRs to correctly make sure we don't unintentionally rule out use cases for "no authorization (header) required", "I enforce access to this endpoint with a firewall" or "some other secret auth method".

dlongley (changes requested):
We should rename `oauth2` to `oauth2-vcapi` so that we can only define the parts we know we can define at this time. There were some requirements when using `oauth2` that weren't really ready to set, so if we invent a name for now as a placeholder we can just do the parts we have consensus on / know what we're doing.

components/SecuritySchemes.yml
msporny, line 38:
```suggestion
    oAuth2:
      type: oauth2-vcapi
```

msporny: This has been removed in commit f2c17c0907ee42e5a819dfb858432688500ee20b.

components/SecuritySchemes.yml
msporny: This has been removed in commit f2c17c0907ee42e5a819dfb858432688500ee20b and is up to the instance to define.

components/SecuritySchemes.yml
msporny: This has been fixed in f2c17c0907ee42e5a819dfb858432688500ee20b.

msporny:
> We should rename `oauth2` to `oauth2-vcapi` so that we can only define the parts we know we can define at this time. 

I was able to update it to just `oAuth2` and make the scheme `bearer` without defining all the oauth-y details in commit f2c17c0907ee42e5a819dfb858432688500ee20b.

dlongley (approved):

msporny:
Discussed on the 2023-03-21 call.

PR #354, opened 2023-12-12
Update workflows to Node v20.x; use latest actions

BigBlueHat:


BigBlueHat:
Note how the linting checks pass now, but are failing on PRs still using the older Node v16.x setup. That's the fix. 😁

msporny (approved):

PR #355, opened 2023-12-12
Fix OpenAPI validity & publishing.

BigBlueHat:
- Remove unused NotifyPresentationAvailableResponse.
- Fix openapi bundling.


BigBlueHat:
FWIW, `NotifyPresentationAvailableResponse` is only present in the OpenAPI files and not in the current spec. So, removing it should be fine. However, I'd love a 👍 from someone before merging this...just in case. 😃

msporny (approved):

dlongley (approved):

Issue #194, opened 2021-06-08
BbsBls sig test fixtures (valid by various W3C docs) not permitted by current OpenAPI spec


vdods:
LinkedDataProof specifies only a proof accepting `jws` as the signature https://github.com/w3c-ccg/vc-http-api/blob/main/docs/components/LinkedDataProof.yml :
```
    LinkedDataProof:
      type: object
      description: A JSON-LD Linked Data proof.
      properties:
        type:
          type: string
          description: Linked Data Signature Suite used to produce proof.
        created:
          type: string
          description: Date the proof was created.
        verificationMethod:
          type: string
          description: Verification Method used to verify proof.
        proofPurpose:
          type: string
          description: The purpose of the proof to be used with verificationMethod.
        jws:
          type: string
          description: Detached JSON Web Signature
```

but the various test fixtures allow for other schemas, e.g. case-18.json https://github.com/w3c-ccg/vc-http-api/blob/main/packages/vc-http-api-test-server/__fixtures__/verifiableCredentials/case-18.json :
```
  "proof": {
    "type": "BbsBlsSignature2020",
    "created": "2021-02-23T19:36:07Z",
    "proofPurpose": "assertionMethod",
    "proofValue": "qSjCNJzoDV3hv3gBPoUNN9m5lj8saDBBxC0iDHuFTXXz4PbbUhecmn/L3rPoGuySNatqC4I8VE22xQy0RAowIxoZCC+B2mZQIAb+/JGlXeAlWgEQc71WipfvsfqSn+KmR/rN1FREOy3rtSltyQ92rA==",
    "verificationMethod": "did:key:zUC724vuGvHpnCGFG1qqpXb81SiBLu3KLSqVzenwEZNPoY35i2Bscb8DLaVwHvRFs6F2NkNNXRcPWvqnPDUd9ukdjLkjZd3u9zzL4wDZDUpkPAatLDGLEYVo8kkAzuAKJQMr7N2#zUC724vuGvHpnCGFG1qqpXb81SiBLu3KLSqVzenwEZNPoY35i2Bscb8DLaVwHvRFs6F2NkNNXRcPWvqnPDUd9ukdjLkjZd3u9zzL4wDZDUpkPAatLDGLEYVo8kkAzuAKJQMr7N2"
  }
```

[...]

PR #358, opened 2024-01-07
Rename LinkedDataProof to DataIntegrityProof.

msporny:
This PR attempts to address issue #194 by renaming the legacy "LinkedDataProof" mechanism to the new "DataIntegrityProof" mechanism being standardized by W3C's VCWG.

### DIFF
--- a/components/DataIntegrityProof.yml
+++ b/components/DataIntegrityProof.yml
@@ -12,13 +12,16 @@ info:
 paths:
 components:
   schemas:
-    LinkedDataProof:
+    DataIntegrityProof:
       type: object
-      description: A JSON-LD Linked Data proof.
+      description: A Data Integrity Proof as defined by the W3C VC Data Integrity specification.
       properties:
         type:
           type: string
-          description: Linked Data Signature Suite used to produce proof.
+          description: Data Integrity proof type.
+        cryptosuite:
+          type: string
+          description: The name of the cryptographic suite.
         created:
           type: string
           description: Date the proof was created.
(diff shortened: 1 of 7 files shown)


components/DeriveCredentialOptions.yml
```diff
@@ -19,7 +19,7 @@ components:
       properties:
         nonce:
           type: string
-          description: An encoded nonce provided by the holder of the credential to be included into the LinkedDataProof.
+          description: An encoded nonce provided by the holder of the credential to be included into the DataIntegrityProof.
```
TallTed, line 22:
```suggestion
          description: An encoded nonce provided by the holder of the credential, to be included in the DataIntegrityProof.
```

TallTed (changes requested):
trivial, for clarity

components/DataIntegrityProof.yml
```diff
@@ -12,13 +12,16 @@ info:
...
-    LinkedDataProof:
+    DataIntegrityProof:
       type: object
-      description: A JSON-LD Linked Data proof.
+      description: A Data Integrity Proof as defined by the W3C VC Data Integrity specification.
       properties:
         type:
           type: string
-          description: Linked Data Signature Suite used to produce proof.
+          description: Data Integrity proof type.
```
jrhender, line 21: "Proof" is capitalized in the object description so it might make sense to do the same here as well or to have it lower case for both
```suggestion
          description: Data Integrity Proof type.
```

jrhender (approved):
I added a minor suggestion and I think Ted's suggestion is good, but I think the PR is good.

brianorwhatever (approved):

dlongley (approved):

components/DataIntegrityProof.yml
TallTed: They should be consistent, but I'm not sure whether titlecase `Proof` or lowercase `proof` is better.

PR #361, opened 2024-01-30
Add support for credentialId

wes-smith:
credentialId allows identification of credentials without credential.id set.

### DIFF
--- a/components/Credential.yml
+++ b/components/Credential.yml
@@ -23,7 +23,7 @@ components:
             type: string
         "id":
           type: string
-          description: The ID of the credential.
+          description: The ID of the credential. This property MAY be empty. The issuer SHOULD NOT auto-generate the id property, as this can lead to unrecoverable partitioning errors.
         "type":
           type: array
           description: The JSON-LD type of the credential.
--- a/components/IssueCredentialOptions.yml
+++ b/components/IssueCredentialOptions.yml
@@ -33,6 +33,9 @@ components:
             type:
               type: string
               description: The type of credential status to issue the credential with
+        credentialId:
+          type: string
+          description: An identifier that can be used to identify the credential. credentialId SHOULD NOT be set if the id property is set. If credential.id is set, that value will be duplicated for credentialId. If neither credentialId nor id is set, there will be no way to uniquely identify a credential or deal with duplicate errors. This property should not be used as a replacement for credential.id, rather as a means of identifying a credential without the id property set.
       example:
         {
           "created": "2020-04-02T18:48:36Z",
--- a/components/IssueCredentialSuccess.yml
+++ b/components/IssueCredentialSuccess.yml
@@ -1,0 +1,34 @@
+openapi: 3.0.0
+info:
+  version: "0.0.3-unstable"
+  title: VC API
+  description: This is an Experimental Open API Specification for the [VC Data Model](https://www.w3.org/TR/vc-data-model/).
+  license:
+    name: W3C Software and Document License
+    url: http://www.w3.org/Consortium/Legal/copyright-software.
+  contact:
+    name: GitHub Source Code
+    url: https://github.com/w3c-ccg/vc-api
+paths:
+components:
+  schemas:
+    IssueCredentialSuccess:
+      type: object
+      description: The format returned from a successful issue request.
+      properties:
+        VerifiableCredential:
+          type: object
+          description: A JSON-LD Verifiable Credential with a proof.
+          allOf:
+            - $ref: "./Credential.yml#/components/schemas/Credential"
+            - type: object
+              properties:
+                proof:
+                  $ref: "./DataIntegrityProof.yml#/components/schemas/DataIntegrityProof"
+        IssuanceMetadata:
+          type: object
+          description: Metadata about the issued credential.
+          properties:
+            credentialId:
+              type: string
+              description: An identifier for the credential. If credentialId is not set by the user in the issue request, credential.id will be duplicated here.
--- a/holder.yml
+++ b/holder.yml
@@ -21,7 +21,7 @@ paths:
        - networkAuth: []
        - oAuth2: []
        - zCap: []
-      summary: Gets a credential or verifiable credential by ID
+      summary: Gets a credential or verifiable credential by ID. To get a credential that does not have credential.id set but has an associated credentialId value, pass credentialId instead.
       operationId: getCredential
       parameters:
         - $ref: "./components/parameters/path/ObjectId.yml"
@@ -55,7 +55,7 @@ paths:
        - networkAuth: []
        - oAuth2: []
        - zCap: []
-      summary: Deletes a credential or verifiable credential by ID
+      summary: Deletes a credential or verifiable credential by ID. To delete a credential that does not have credential.id set but has an associated credentialId value, pass credentialId instead.
       operationId: deleteCredential
       parameters:
         - $ref: "./components/parameters/path/ObjectId.yml"
--- a/issuer.yml
+++ b/issuer.yml
@@ -81,8 +81,8 @@ components:
     IssueCredentialResponse:
       type: object
       properties:
-        verifiableCredential:
-          $ref: "./components/VerifiableCredential.yml#/components/schemas/VerifiableCredential"
+        IssueCredentialSuccess:
+          $ref: "./components/IssueCredentialSuccess.yml#/components/schemas/IssueCredentialSuccess"
     UpdateCredentialStatus:
       type: object
       description: Request for updating the status of an issued credential.


mkhraisha:
If `credentialId` is set but `credential.id` is empty is the intention to tell the issuer to assign `credential.id` the value of `crdentialId`?

components/IssueCredentialSuccess.yml
```diff
@@ -0,0 +1,34 @@
...
+    name: GitHub Source Code
+    url: https://github.com/w3c-ccg/vc-api
+paths:
+components:
+  schemas:
+    IssueCredentialSuccess:
+      type: object
+      description: The format returned from a successful issue request.
+      properties:
+        VerifiableCredential:
```
msporny, line 19:
```suggestion
        verifiableCredential:
```

components/IssueCredentialSuccess.yml
```diff
@@ -0,0 +1,34 @@
...
+              properties:
+                proof:
+                  $ref: "./DataIntegrityProof.yml#/components/schemas/DataIntegrityProof"
+        IssuanceMetadata:
+          type: object
+          description: Metadata about the issued credential.
+          properties:
+            credentialId:
+              type: string
+              description: An identifier for the credential. If credentialId is not set by the user in the issue request, credential.id will be duplicated here.
```
msporny, line 34:
```suggestion
        issuanceMetadata:
          type: object
          description: Metadata about the issued credential.
          properties:
            credentialId:
              type: string
              description: An identifier for the credential. If credentialId is not set by the user in the issue request, credential.id will be duplicated here.
```

components/IssueCredentialSuccess.yml
```diff
@@ -0,0 +1,34 @@
...
+        VerifiableCredential:
+          type: object
+          description: A JSON-LD Verifiable Credential with a proof.
+          allOf:
+            - $ref: "./Credential.yml#/components/schemas/Credential"
+            - type: object
+              properties:
+                proof:
+                  $ref: "./DataIntegrityProof.yml#/components/schemas/DataIntegrityProof"
+        IssuanceMetadata:
```
msporny, line 28:
```suggestion
        issuanceMetadata:
```

components/Credential.yml
```diff
@@ -23,7 +23,7 @@ components:
             type: string
         "id":
           type: string
-          description: The ID of the credential.
+          description: The ID of the credential. This property MAY be empty. The issuer SHOULD NOT auto-generate the id property, as this can lead to unrecoverable partitioning errors.
```
msporny, line 26:
```suggestion
          description: The ID of the credential. This property MAY be empty. The issuer SHOULD NOT auto-generate the id property since the client wanted to issue a VC without the id.
```

components/IssueCredentialOptions.yml
```diff
@@ -33,6 +33,9 @@ components:
             type:
               type: string
               description: The type of credential status to issue the credential with
+        credentialId:
+          type: string
+          description: An identifier that can be used to identify the credential. credentialId SHOULD NOT be set if the id property is set. If credential.id is set, that value will be duplicated for credentialId. If neither credentialId nor id is set, there will be no way to uniquely identify a credential or deal with duplicate errors. This property should not be used as a replacement for credential.id, rather as a means of identifying a credential without the id property set.
```
msporny, line 38:
```suggestion
          description: An identifier that can be used to identify the credential. credentialId SHOULD NOT be set if the id property is set. If credential.id is set, that value will be duplicated for credentialId. If neither credentialId nor id is set, there will be no way to uniquely identify a credential or deal with duplicate errors. This property should not be used as a replacement for credential.id, rather as a means of identifying a credential without the id property set. The issuer SHOULD NOT auto-generate a credentialId because doing so could create a partitioning error if the result is never received by the client.
```

dlongley:
@mkhraisha,

> If `credentialId` is set but `credential.id` is empty is the intention to tell the issuer to assign `credential.id` the value of `crdentialId`?

No, `credential.id` should remain empty as the client requests -- since clients need to be able to generate VCs that do not include `credential.id` (which are valid VCs -- and useful things like unlinkable presentations). The `credentialId` option allows clients to ensure that they have a reference to the credential they intended to issue even when `credential.id` is not set.

mkhraisha:
> @mkhraisha,
> 
> 
> 
> > If `credentialId` is set but `credential.id` is empty is the intention to tell the issuer to assign `credential.id` the value of `crdentialId`?
> 
> 
> 
> No, `credential.id` should remain empty (unset) as the client requests -- since clients need to be able to generate VCs that do not include `credential.id` (which are valid VCs -- and useful for things like unlinkable selective disclosure). The `credentialId` option allows clients to ensure that they have a reference to the credential they intended to issue even when `credential.id` is not set.

With the initial wording it was a bit confusing but after reading Manu's suggested changes, this explanation makes sense.

components/IssueCredentialSuccess.yml
msporny, line 34:
```suggestion
```

The group discussed this and decided to not define the metadata field at this point in time.

components/Credential.yml
```diff
@@ -23,7 +23,7 @@ components:
             type: string
         "id":
           type: string
-          description: The ID of the credential. This property MAY be empty. The issuer SHOULD NOT auto-generate the id property, as this can lead to unrecoverable partitioning errors.
+          description: The ID of the credential. This property MAY be empty. The issuer SHOULD NOT auto-generate the id property since the client wanted to issue a VC without the id.
```
TallTed:
```suggestion
          description: The ID of the credential. This property MAY be empty. The issuer SHOULD NOT auto-generate the id property since the client wanted to issue a VC without an ID.
```

components/IssueCredentialOptions.yml
```diff
@@ -35,7 +35,7 @@ components:
...
+          description: A URI that can be used to identify the credential, that will be used in APIs to refer to the issued verifiable credential. If not provided, the value will be auto-populated from credential.id. If credentialId is not provided and credential.id is not provided, it is not possible to refer to this credential once issued or deal with duplicate errors. credentialId SHOULD NOT be set if the credential.id property is set. This property should not be used as a replacement for credential.id, rather as a means of identifying a credential without the id property set. The issuer SHOULD NOT auto-generate a credentialId if one is not supplied because doing so could create a partitioning error if the result is never received by the client.
```
TallTed, line 38:
```suggestion
          description: A URI that can be used to identify the credential, that will be used in APIs to refer to the issued verifiable credential. If not provided, the value will be auto-populated from credential.id. If neither credentialId nor credential.id is provided, it is not possible to refer to this credential once issued nor to deal with duplicate errors. credentialId SHOULD NOT be set if the credential.id property is set. This property should not be used as a replacement for credential.id; rather, it is a means of identifying a credential without the id property being set. The issuer SHOULD NOT auto-generate a credentialId if one is not supplied, because doing so could create a partitioning error if the result is never received by the client. 
```

TallTed (changes requested):

components/Credential.yml
wes-smith: Outdated - see newest version of Credential.yml.

components/IssueCredentialOptions.yml
```diff
@@ -35,7 +35,7 @@ components:
...
+          description: A URI that can be used to identify the credential, that will be used in APIs to refer to the issued verifiable credential. If not provided, the value will be auto-populated from credential.id. If neither credentialId nor credential.id is provided, it is not possible to refer to this credential once issued nor to deal with duplicate errors. credentialId SHOULD NOT be set if the credential.id property is set. This property should not be used as a replacement for credential.id; rather, it is a means of identifying a credential without the id property being set. The issuer SHOULD NOT auto-generate a credentialId if one is not supplied, because doing so could create a partitioning error if the result is never received by the client.
```
TallTed, line 38: I am struggling to understand the totality of this description. "If credentialId is not provided, its value will be auto-populated from credential.id" but "credentialId SHOULD NOT be set if the credential.id property is set". These seem contradictory. What am I missing?
```suggestion
          description: A URI that can be used to identify the credential, that can be used in APIs to refer to the issued verifiable credential. If credentialId is not provided, its value will be auto-populated from credential.id. If neither credentialId nor credential.id is provided, it will not be possible to refer to this credential once issued nor to deal with errors of duplication. credentialId SHOULD NOT be set if the credential.id property is set, nor should credentialId be used as a replacement for credential.id; rather, credentialId is a means of identifying a credential without the id property being set. The issuer SHOULD NOT auto-generate a credentialId if one is not supplied, because doing so could create a partitioning error if the result is never received by the client. 
```

PatStLouis:
Should we include some text about duplicate `credentialId` and maybe add a 409 error on the issuance endpoint?

TallTed:
Updated change request, as discussed.

msporny (approved):

msporny:

PR #366, opened 2024-02-06
Add returnCredential option to VerifyOptions

jrhender:
PR to address https://github.com/w3c-ccg/vc-api/issues/207

TODO:
- [x] Fix extra json that is showing up in `index.html`
- [x] Describe what "canonical form" means. clarified here: https://github.com/w3c-ccg/vc-api/issues/207#issuecomment-1954988441
- [x] Mention that the parameter is optional
- [x] Add return examples -> decided not to add `verifiedCredential` to example as it is an optional parameter
- [x] Add the feature to the `Verify Presentation` presentation end point -> I think it is better to do this in a separate PR

### DIFF
--- a/components/VerifyOptions.yml
+++ b/components/VerifyOptions.yml
@@ -26,7 +26,7 @@ components:
         returnCredential:
           type: boolean
           description:
-            Should the verified credential be returned in the response.
+            Should the verified credential be returned in the response?
             If true, then the verified credential should be returned in the form in which it was verified.
             If false or not provided, then the verified credential should not be returned.
       example:
--- a/components/VerifyOptions.yml
+++ b/components/VerifyOptions.yml
@@ -23,6 +23,9 @@ components:
         domain:
           type: string
           description: The intended domain of validity for the proof. For example website.example
+        returnCredential:
+          type: boolean
+          description: Should the verified credential be returned in its canonical form.
       example:
         {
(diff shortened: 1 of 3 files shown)


components/VerifyOptions.yml
```diff
@@ -23,11 +23,15 @@ components:
         domain:
           type: string
           description: The intended domain of validity for the proof. For example website.example
+        returnCredential:
+          type: boolean
```
jrhender, line 27: Maybe `boolean` is not a valid keyword and this is why the extra json is appearing

jrhender: Fixed by https://github.com/w3c-ccg/vc-api/pull/366/commits/5fa7f3ae6e840587edeb263aa95a0dcc0c3c6b09

components/VerifyOptions.yml
```diff
@@ -23,11 +23,18 @@ components:
...
+          description:
+            Should the verified credential be returned in the response.
```
TallTed, line 29:
```suggestion
            Should the verified credential be returned in the response?
```

TallTed (changes requested):
trivial, human facing, for clarity

msporny (approved):

msporny:
Normative, multiple reviews (on the call), changes requested and made, no objections, merging.

PR #367, opened 2024-02-09
Modify to support selective disclosure options

Wind4Greg:
This PR addresses issues https://github.com/w3c-ccg/vc-api/issues/364 and https://github.com/w3c-ccg/vc-api/issues/363 to support selective disclosure cryptosuites such as ECDSA-SD and BBS. It adds `mandatoryPointers` to the issue options and `selectivePointers` to the derive options. It also removes the `frame` from the derive inputs.


components/IssueCredentialOptions.yml
```diff
@@ -1,40 +1,43 @@
...
+      properties:
+        created:
+          type: string
+          description: The date and time of the proof (with a maximum accuracy in seconds). Default current system time.
+        challenge:
+          type: string
+          description: A challenge provided by the requesting party of the proof. For example 6e62f66e-67de-11eb-b490-ef3eeefa55f2
+        domain:
+          type: string
+          description: The intended domain of validity for the proof. For example website.example
```
dlongley, line 28: These are not options for issuance (they are for authentication proofs that holders generate), we could remove them in this PR or another later.

dlongley (approved):

TallTed (changes requested):
Additionally, variably quoted values (`""`-wrapped, `''`-wrapped, not quote-wrapped) should be made consistent.

components/IssueCredentialOptions.yml
msporny: I think these are re-used in presentation creation, which is why they're here. Agree that we probably want to differentiate the Credential issuance from presentation creation/issuance functions.

holder.yml
```diff
@@ -1,550 +1,547 @@
...
+                    type: number
+                    description: The total number of elements available.
+                  exchanges:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        id:
+                          type: string
+                          description: 'the path name of the exchange endpoint. May be a UUID.'
```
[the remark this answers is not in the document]
msporny: No idea, it doesn't need to be. Fixing by removing single quotes.

```suggestion
                          description: the path name of the exchange endpoint. May be a UUID.

```

msporny:
@Wind4Greg there are merge conflicts that need to be resolved. I also don't understand why it looks like the .yml files are being wholesale deleted and re-created (even though much of the content is the same)... wondering if there is something problematic going on here... or if Github's UI is just confused?

Wind4Greg:
Per the ["development instructions"](https://github.com/w3c-ccg/vc-api?tab=readme-ov-file#development) I ran the `npm run build` command after making my changes. The instructions say that this generates files in the `api/` directory. But don't know if this is the source of the problem.

components/IssueCredentialOptions.yml
```diff
@@ -1,44 +1,47 @@
...
+components:
+  schemas:
+    IssueCredentialOptions:
+      type: object
+      additionalProperties: false
+      description: Options for specifying how the DataIntegrityProof is created.
+      properties:
+        created:
+          type: string
+          description: The date and time of the proof (with a maximum accuracy in seconds). Default current system time.
```
TallTed, line 22:
```suggestion
          description: The date and time of the proof (with a maximum accuracy in seconds). Defaults to current system time.
```

components/IssueCredentialOptions.yml
```diff
@@ -1,44 +1,47 @@
...
+        challenge:
+          type: string
+          description: A challenge provided by the party requesting the proof. For example, 6e62f66e-67de-11eb-b490-ef3eeefa55f2
+        domain:
+          type: string
+          description: The intended domain of validity for the proof. For example, website.example
```
TallTed, line 28:
```suggestion
          description: A challenge provided by the party requesting the proof. For example, 6e62f66e-67de-11eb-b490-ef3eeefa55f2.
        domain:
          type: string
          description: The intended domain of validity for the proof. For example, website.example.
```

components/IssueCredentialOptions.yml
```diff
@@ -1,44 +1,47 @@
...
+
+        credentialStatus:
+          type: object
+          description: The method of credential status to include in the issued credential. If omitted, credential status will be included.
```
TallTed, line 32: This doesn't sense to me. "If [credentialStatus is] omitted, credential status will be included."
```suggestion
          description: The method of credential status to include in the issued credential. If omitted, credential status will be included.
```

components/IssueCredentialOptions.yml
```diff
@@ -1,44 +1,47 @@
...
+          properties:
+            type:
+              type: string
+              description: The type of credential status to issue the credential with
+        mandatoryPointers:
+          type: array
+          description: Used with selective disclosure schemes to specify mandatory reveal statements
```
TallTed, line 39:
```suggestion
              description: The type of credential status with which to issue the credential.
        mandatoryPointers:
          type: array
          description: Used with selective disclosure schemes to specify mandatory reveal statements.
```

components/IssueCredentialOptions.yml
```diff
@@ -1,44 +1,47 @@
...
+        credentialId:
+          type: string
+          description: A URI that can be used to identify the credential, that can be used in APIs to refer to the issued verifiable credential. If credentialId is not provided by the issuer coordinator, the issuer service will auto-populate its value from credential.id. If neither credentialId nor credential.id is provided, it will not be possible to refer to this credential once issued nor to deal with errors of duplication. credentialId SHOULD NOT be set by the issuer coordinator if the credential.id property is set, nor should credentialId be used as a replacement for credential.id; rather, credentialId is a means of identifying a credential without the id property being set. The issuer service SHOULD NOT auto-generate a credentialId if one is not supplied by the issuer coordinator, because doing so could create a partitioning error if the result is never received by the client.
```
TallTed, line 42:
```suggestion
          description: A URI that can be used to identify the credential, that can be used in APIs to refer to the issued verifiable credential. If credentialId is not provided by the issuer coordinator, the issuer service will auto-populate its value from credential.id. If neither credentialId nor credential.id is provided, it will not be possible to refer to this credential once issued nor to deal with errors of duplication. credentialId SHOULD NOT be set by the issuer coordinator if the credential.id property is set, nor should credentialId be used as a replacement for credential.id; rather, credentialId is a means of identifying a credential without the id property (i.e., credential.id) being set. The issuer service SHOULD NOT auto-generate a credentialId if one is not supplied by the issuer coordinator, because doing so could create a partitioning error if the result is never received by the client.
```

holder.yml
```diff
@@ -1,553 +1,553 @@
...
+        "500":
+          description: Internal Error
+        "501":
+          description: Not Implemented
+
+  /presentations/{id}:
+    get:
+      tags:
+       - Presentations
+      summary: Gets a presentation or verifiable presentation by ID
```
TallTed, line 152:
```suggestion
      summary: Gets a presentation or verifiable presentation by ID.
```

holder.yml
```diff
@@ -1,553 +1,553 @@
...
+          description: Gone! There is no data here
+        "500":
+          description: Internal Error
+        "501":
+          description: Not Implemented
+  /presentations:
+    get:
+      tags:
+       - Presentations
+      summary: Gets list of presentations or verifiable presentations
```
TallTed, line 211:
```suggestion
      summary: Gets a list of presentations and/or verifiable presentations.
```

holder.yml
```diff
@@ -1,553 +1,553 @@
...
+      security:
+       - networkAuth: []
+       - oAuth2: []
+       - zCap: []
+      operationId: discoverExchanges
+      description:
+        This endpoint returns an array of the exchange-ids (path endpoints) supported by this server,
+        and the associated protocol supported by each exchange endpoint. The list supports pagination.
+
+        Clients consuming this list wishing to use an exchange endpoint MUST recognize and support
```
TallTed, line 290:
```suggestion
        Clients consuming this list and wishing to use an exchange endpoint MUST recognize and support
```

holder.yml
```diff
@@ -1,553 +1,553 @@
...
+
+          required: false
+          schema:
+            type: string
+        - name: limit
+          in: query
+          description:
+            The maximum number of items to return in the response. If omitted, the service should
+
+            return all remaining items from the start index.
```
TallTed, line 308:
```suggestion
            The maximum number of items to return in the response. If omitted, the service should
            return all remaining items from the start index.
```

holder.yml
```diff
@@ -1,553 +1,553 @@
...
+          required: false
+          schema:
+            type: number
+      responses:
+        "200":
+          description: A map of the exchange-id endpoints to protocols those exchanges support.
```
TallTed, line 314:
```suggestion
          description: A map associating the exchange-id endpoints with the protocols those exchanges support.
```

holder.yml
```diff
@@ -1,553 +1,553 @@
...
+                          description: MUST be a string that references the protocol supported on that endpoint.
+                  index:
+                    type: object
+                    properties:
+                      self:
+                        type: string
+                        description:
+                          The index position of the start of the returned list. Examples could be a numerical value,
+
+                          a URL, or a value meaningful to the server.
```
TallTed, line 350:
```suggestion
                          The index position of the start of the returned list. Examples could be a numerical value,
                          a URL, or a value meaningful to the server.
```

holder.yml
```diff
@@ -1,553 +1,553 @@
...
+       - Exchanges
+      security:
+       - networkAuth: []
+       - oAuth2: []
+       - zCap: []
+      operationId: initiateExchange
+      description:
+        A client can use this endpoint to initiate an exchange of a particular
+        type. The client can include HTTP POST information related to the
+        details of exchange it would like to initiate. If the server understands
```
TallTed, line 398:
```suggestion
        details of the exchange it would like to initiate. If the server understands
```

TallTed (changes requested):

components/IssueCredentialOptions.yml
```diff
@@ -22,20 +22,24 @@ components:
...
         challenge:
           type: string
-          description: A challenge provided by the requesting party of the proof. For example 6e62f66e-67de-11eb-b490-ef3eeefa55f2
+          description: A challenge provided by the party requesting the proof. For example, 6e62f66e-67de-11eb-b490-ef3eeefa55f2
         domain:
           type: string
-          description: The intended domain of validity for the proof. For example website.example
+          description: The intended domain of validity for the proof. For example, website.example
+
         credentialStatus:
```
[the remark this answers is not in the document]
Wind4Greg: Okay just removed `credentialStatus`.

Wind4Greg:
@TallTed  and @msporny the one thing strange here is that I only added and removed a couple of the lines in the YAML files and somehow git/GitHub makes it look like I edited a bunch of text that I didn't touch. However, I don't mind rolling in the fixes that you indicated, it just that I didn't write any of that text for *credentialId*. Weird.

TallTed:
GitHub has been doing a lot of weird change highlighting. I *think* it's somehow attaching invisible info to invisible characters, and various edits invisibly change one or both of the invisible info and/or characters, which can mean that unchanged things are highlighted as if they'd changed, and changed things are *not* highlighted as expected.

I think it's OK to broaden the scope of PRs somewhat when that happens, especially (maybe only) when the extra changes are editorial. Others may feel it important to move the extra changes to separate PRs. I'm OK with that, too, though I'd rather not be the one to create such separate PRs.

msporny (approved):

msporny:
Normative, multiple reviews (on call), changes requested and made, no objections, merging.

Issue #383, opened 2024-05-14
proposition to follow rfc9457 for API errors


lemoustachiste:
RFC 9457 (https://www.rfc-editor.org/rfc/rfc9457.html) defines a Problem Details response for HTTP APIs.

It proposes a standard response object to explicit the error (instead of just returning error status for instance). By having an expectable shape for errors, it makes error handling easier and more consistent across various providers in the client (for instance to display feedback to the user).

To quote from the above linked document:

> A problem details object can have the following members:
>
>   o  "type" (string) - A URI reference [[RFC3986](https://www.rfc-editor.org/rfc/rfc3986)] that identifies the
      problem type.  This specification encourages that, when
      dereferenced, it provide human-readable documentation for the
      problem type (e.g., using HTML [[W3C.REC-html5-20141028](https://www.rfc-editor.org/rfc/rfc7807#ref-W3C.REC-html5-20141028)]).  When
      this member is not present, its value is assumed to be
      "about:blank".
>
>   o  "title" (string) - A short, human-readable summary of the problem
      type.  It SHOULD NOT change from occurrence to occurrence of the
      problem, except for purposes of localization (e.g., using
      proactive content negotiation; see [[RFC7231], Section 3.4](https://www.rfc-editor.org/rfc/rfc7231#section-3.4)).
>
>   o  "status" (number) - The HTTP status code ([[RFC7231], Section 6](https://www.rfc-editor.org/rfc/rfc7231#section-6))
      generated by the origin server for this occurrence of the problem.
>
>   o  "detail" (string) - A human-readable explanation specific to this
      occurrence of the problem.
>
>   o  "instance" (string) - A URI reference that identifies the specific
      occurrence of the problem.  It may or may not yield further
      information if dereferenced.

msporny:
The discussed this on the 2024-05-14 call:

@msporny Noted that we use RFC 9457 in vc-data-model. @dlongley noted that we wanted it to be easy to implement API w/o being too prescriptive on errors. There was some concern around using `type` and `name`, not URL. Maybe we want to mix those two together. We didn't have good implementation experience with it last time, didn't want to be too prescriptive. Generally, a good idea, will need to link to other issue that talks about this.

This issue is ready for a PR to say that implementations SHOULD use RFC 9457 to express problems over HTTP. We can re-use most of the text here: https://w3c.github.io/vc-data-model/#problem-details

PatStLouis:
[...]

Issue #340, opened 2023-03-21
Clarify error response code design in spec


msporny:
@ottonomy wrote:
>On the call today, we resolved to close this PR and replace it with 2 actionable PRs, for each of the above action items I researched:
> * I will make a PR to remove the specification of response content for unsuccessful status codes, replacing that with spec text describing that the response should conform to the `ErrorResponse` schema if content is returned. 
> * I will make a PR to clarify the ErrorResponse schema roughly matching the above example, except that the `pattern` should be removed from `name` and minLength, maxLength should be removed from `message`, relying on clear textual description that `name` should be short, `message` can be longer.

msporny:
We should consider using RFC 7807 for this purpose. The Data Integrity work might adopt that as an error response mechanism if used over HTTP APIs:

https://github.com/w3c/vc-data-integrity/issues/136

The same considerations over there apply here.

msporny:
The group discussed this on the 2023-08-01 telecon and agreed that the approach provided by RFC 7807 is acceptable. The error object will be updated in the spec as a SHOULD, not a MUST. There should be further detail around exposing a "log of errors" so callers know everything that went wrong/right in PRIVATE calls (like Issuer Coordinator -> Issuer Service), and when detail or error reporting might need to be redacted or not fully exposed (such as when a public endpoint is being hit... Holder -> Verifier Coordinator).

msporny:
Take a look at this for an idea on how errors might be exposed to a caller: https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1338.html#problem-details

msporny:
PR #389 has been merged, which addresses this issue. Closing.

Issue #331, opened 2023-01-10
Specify how error details are returned


msporny:
The group discussed this on 2023-01-10. @dlongley noted that he was ambivalent about whether or not everything has an error response. If there are values that need to be sent back to be acted upon vs. all information being communicated via status -- that seems to be the most important thing. Ideally, we don't invent our own thing for error codes or the error data model -- there is some RFC that Mark Nottingham might have put together for HTTP Errors, don't know if anyone uses it, don't know if it fits our use cases, might not match up with how errors are returned from some implementations today -- some implementations use `name` field to match JavaScript errors, might be problematic for a W3C spec -- a number of considerations. @jandrieu asked "what is the framework for extensibility in the responses?" -- there might be open ended things to pass back. @msporny noted that current error response has `id` which is mandatory, `description` which contains explanatory text, and `details` which can contain arbitrary infomation from the implementation. @dlongley noted that `id` is problematic and maybe we should use `name` instead. @jandrieu agreed. 

Do we want to always return an error object on an error? @TallTed noted mandating minimal requirements but allowing for larger responses -- as a troubleshooter, want detailed errors, could be configurable/switchable. That is implementation specific. Leave the door open for implementations to provide sufficient information to troubleshoot information that's erroring. @dlongley noted that we shouldn't ALWAYS return an error code until we it's clear that we should. We need to find cases where HTTP status code is insufficient and mandate detailed errors through that mechanism, but only when necessary. An implementation is allowed to return an ErrorResponse if it chooses to do so, and those errors might not be the same across implementations UNLESS it is specified to be the same by the specification. @PatStLouis noted that covering all status codes is a must, two different errors could return the same error code, would be good to know the reason for it ... for cases where implementer wants to return more detail, good to have a pre-defined schema, but leave content to implementer.

The group arrived at consensus on the following:

1. Errors will, at a minimum, be conveyed using HTTP error response codes.
2. Implementers MAY return error responses using the ErrorResponse schema, or their own, (for any HTTP error).
3. The specification MAY (at a later date) define specific ErrorResponses in specific situations.

@dlongley NOTE: We might want to just say error response is JSON and not use a suggested schema and that can be bike shedded in the PR.

[...]

Issue #141, opened 2021-03-05
Response body for POST /credentials/status


OR13:
What is it supposed to be?

I assume the updated revocation list credential.

probably the request payload should take a number instead of as string as well:

POST https://vc.example/credentials/status

```
{
  "credentialId": "https://vc.example/credentials/aabe7aca-e4aa-4e1f-b08c-a86dddc5b0c3",
  "credentialStatus": [
    {
      "type": "RevocationList2020Status",
      "status": "1"
    }
  ]
}
```

for example:

```
let revoked = credentialStatus[0].status === "1" ? true : false;
```

OR13:
Here is a postman test confirming this behavior: https://github.com/w3c-ccg/traceability-interop/pull/84

OR13:
No feedback on issue, tagging as evidence revocation is poorly understood / implemented and should be removed.

msporny:
This was discussed on the 2022-10-11 telecon. @dlongley noted that Digital Bazaar's implementation currently provides a status code, with an empty body. If there is an error, there is an error that's returned in the standard error format. StatusList2021 could return back the updated credential, but it might be advantageous to not do that (people might want implementations that update that in a background process). We might want to hear if clients need the updated value or whether or not that's necessary. If it's not necessary, we might not want response value. The question was asked if anyone thought we should be returning anything in the body of if just a status code was enough. Positive is 200, negative is 400 which should include error (if possible).

Pull request should update the POST /credentials/status endpoint such that a positive modification provides a 2xx status code with no content in the response body and a failure returns a 4xx or 5xx with the error included in the response body.

msporny:
PR #389 has been merged, which addresses this issue. Closing.

PR #389, opened 2024-06-04
Add error handling section

PatStLouis:
Addresses #383 #340 #331 and #141 

Standing questions:

[RFC 9457](https://www.rfc-editor.org/rfc/rfc9457.html) or [RFC 7807](https://datatracker.ietf.org/doc/html/rfc7807)?

What should the relationship with [verification algorithms](https://w3c.github.io/vc-data-model/#algorithms) from the spec be?

Should we define specific error codes for each [ProblemDetails](https://w3c.github.io/vc-data-model/#problem-details)?

Is it fair to assume that if an errors are returned the credential is not verified but if only warning are returned it is verified? (see list of error/warnings in the PR for details)

Keyword [status](https://datatracker.ietf.org/doc/html/rfc7807#section-3.1) vs code...

***
<a href="https://pr-preview.s3.amazonaws.com/OpSecId/vc-api/pull/389.html" title="Last updated on Jun 18, 2024, 7:20 PM UTC (bc3b5c5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/389/c02e73f...OpSecId:bc3b5c5.html" title="Last updated on Jun 18, 2024, 7:20 PM UTC (bc3b5c5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1186,25 +1186,27 @@ Connection: keep-alive
         when identifying technical friction impacting interoperability.
       </p>
       </p>
-        Other fields such as [detail](https://www.rfc-editor.org/rfc/rfc9457.html#name-detail), 
-        [instance](https://www.rfc-editor.org/rfc/rfc9457.html#name-instance) 
+        Leveraging other fields such as [detail](https://www.rfc-editor.org/rfc/rfc9457.html#name-detail), 
+        [instance](https://www.rfc-editor.org/rfc/rfc9457.html#name-instance),
         and [type](https://www.rfc-editor.org/rfc/rfc9457.html#name-type) 
-        are encouraged to be leveraged to provide more contextual feedback about the error, 
-        while being conscious of not disclosing sensitive information for security concerns.
+        is encouraged, to provide more contextual feedback about the error, 
+        while being conscious of security concerns and hence not disclosing
+        sensitive information.
       </p>
       <p>
-        Implementers should try to handle all server errors in the best of their capabilities. 
-        Endpoints should avoid returning improperly handled 500 errors in a production 
-        environment leading to potential [information disclosure](https://owasp.org/www-community/Improper_Error_Handling).
+        Implementers should handle all server errors to the best of their capabilities. 
+        Endpoints should avoid returning improperly handled 500 errors in production 
+        environments, as these may lead to [information disclosure](https://owasp.org/www-community/Improper_Error_Handling).
       </p>
       <h4>Relationship between verification and error handling</h4>
       <p>
-        While performing [verification](https://w3c.github.io/vc-data-model/#verification), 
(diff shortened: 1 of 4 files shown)


TallTed (changes requested):

TallTed (approved):

msporny:
The group discussed this on the 2024-06-04 telecon:

> RFC 9457 or RFC 7807?

RFC 9457 should be used.

> What should the relationship with verification algorithms from the spec be?

We shouldn't tie too closely to VCDM v2.0 since the API could be used to issue mDL, mDoc, and other types of digital credentials.

> Should we define specific error codes for each ProblemDetails?

Not yet. We need to make sure we don't mix "HTTP Error Codes" with "ProblemDetails" codes. We need to update the ProblemDetails object to align w/ RFC 9457 (add the "status" field). We need to also figure out (later, in another PR) what to do for things like "error revoked" when validating a VC, and so on.

> Is it fair to assume that if an errors are returned the credential is not verified but if only warning are returned it is verified? (see list of error/warnings in the PR for details)

This feels like a safe assumption to make.

> Keyword status vs code...

"status" is for "HTTP Error Codes" (200, 400, etc.) and "code" is for integer codes returned by low-level libraries.

index.html
```diff
@@ -1207,27 +1207,37 @@ <h4>Relationship between verification and error handling</h4>
...
-          "warnings": [ProblemDetails],
-          "errors": [ProblemDetails]
-        }
-    </code>
+        <h5>Errors and Warnings</h5>
+        Errors are ProblemDetails relating to cryptography, data model and malformed context.
+        Warnings are ProblemDetails relating to statuses and validity periods.
+
+        The VerificationResponse object MUST have the verified value set to false if an error is included.
+        The VerificationResponse object MUST have the verified value set to true if no errors are included.
```
TallTed, line 1227:
```suggestion
        Errors are `ProblemDetails` relating to cryptography, data model and malformed context.
        Warnings are `ProblemDetails` relating to statuses and validity periods.

        The `VerificationResponse` object MUST have the verified value set to `false` if an error is included.
        The `VerificationResponse` object MUST have the verified value set to `true` if no errors are included.
```

TallTed (changes requested):

aljones15 (approved):
LGTM

index.html
```diff
@@ -1173,6 +1174,65 @@ <h4>Exchange Examples</h4>
...
+
+        <pre class="example"
+          title="Verification Response">
+          {
+            "verified": false,
+            "document": verifiableCredential,
+            "mediaType": "vc+json+ld",
```
msporny, line 1227:
```suggestion
            "mediaType": "application/vc",
```

Issue #393, opened 2024-06-18
`workflowId` should be `id` in posted and returned workflow configuration


dlongley:
When creating a workflow, the configuration for that workflow may include, optionally, the ID for that workflow under the property `id`. The bug currently says that property is `workflowId` and it should be changed to `id`:

https://github.com/w3c-ccg/vc-api/blob/494f0a2927a7c6766730a6b4f18c8d28daa15131/exchanges.yml#L210

Similarly, the spec currently says that `workflowId` will be optionally returned, when the entire config can be optionally returned, which will include the workflow ID in the `id` property of the configuration object (just like how it can be optionally posted using that property).

msporny:
The group discussed this on the 2024-07-30 telecon:

There seems to be a typo that needs to be fixed in the specification. 

A PR should be raised to fix the typo.

msporny:
This seems to have been fixed a while ago:

https://github.com/w3c-ccg/vc-api/blob/main/exchanges.yml#L221-L223

PR #409, opened 2024-08-06
Added workflow response object to workflow creation call

PatStLouis:
Addresses #393

Replace `workflowId` in request/response and returned a workflow object.

### DIFF
--- a/exchanges.yml
+++ b/exchanges.yml
@@ -233,23 +233,19 @@ components:
             id:
               type: string
               description: The URL that uniquely identifies the created workflow.
-            stepInformation:
+            steps:
               type: object
-              description: Information about the steps required for the workflow. Returning stepInformation is REQUIRED.
+              description: One or more steps required to complete an exchange on the workflow.
               properties:
-                steps:
+                stepName:
                   type: object
-                  description: One or more steps required to complete an exchange on the workflow.
+                  description: The name of the step.
                   properties:
-                    stepName:
-                      type: object
-                      description: The name of the step.
-                      properties:
-                        step:
-                          $ref: "#/components/schemas/WorkflowStep"
-                initialStep:
-                  type: string
-                  description: The step from the above set that the exchange starts on.
+                    step:
+                      $ref: "#/components/schemas/WorkflowStep"
+            initialStep:
+              type: string
+              description: The step from the above set that the exchange starts on.
             controller:
               type: string
               description: The controller of the instance. Returning controller is OPTIONAL.
@@ -278,28 +274,24 @@ components:
           additionalProperties: false
           description: Object containing information about a created workflow.
           properties:
-            stepInformation:
+            exchanges:
+              type: array
+              description: The identifiers of the current exchanges associated with the workflow instance.
+              items:
+                type: string
+            steps:
               type: object
-              description: Information about the steps required for the workflow. Returning stepInformation is REQUIRED.
+              description: One or more steps required to complete an exchange on the workflow.
               properties:
-                exchanges:
-                  type: array
-                  description: The identifiers of the current exchanges associated with the workflow instance.
-                  items:
-                    type: string
-                steps:
+                stepName:
                   type: object
-                  description: One or more steps required to complete an exchange on the workflow.
+                  description: The name of the step.
                   properties:
-                    stepName:
-                      type: object
-                      description: The name of the step.
-                      properties:
-                        step:
-                          $ref: "#/components/schemas/WorkflowStep"
-                initialStep:
-                  type: string
-                  description: The step from the above set that the exchange starts on.
+                    step:
+                      $ref: "#/components/schemas/WorkflowStep"
+            initialStep:
+              type: string
+              description: The step from the above set that the exchange starts on.
             controller:
               type: string
               description: The controller of the instance. Returning controller is OPTIONAL.
--- a/exchanges.yml
+++ b/exchanges.yml
@@ -217,17 +217,62 @@ components:
                     type: string
                   template:
                       type: string
-            workflowId:
+            id:
               type: string
-              description: the ID that will be used for the created workflow. workflowId is OPTIONAL.
+              description: The ID that will be used for the created workflow. Passing an ID is OPTIONAL.
     CreateWorkflowResponse:
       type: object
       additionalProperties: false
-      description: Object containing information about a created workflow.
+      description: Response containing the workflowData Object created.
       properties:
-        workflowId:
-          type: string
-          description: The URL that uniquely identifies the created workflow.
+        workflowData:
+          type: object
+          additionalProperties: false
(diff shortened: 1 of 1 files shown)


exchanges.yml
```diff
@@ -217,17 +217,62 @@ components:
...
-            workflowId:
+            id:
               type: string
-              description: the ID that will be used for the created workflow. workflowId is OPTIONAL.
+              description: The ID that will be used for the created workflow. Passing an ID is OPTIONAL.
     CreateWorkflowResponse:
       type: object
       additionalProperties: false
-      description: Object containing information about a created workflow.
+      description: Response containing the workflowData Object created.
```
TallTed, line 226:
```suggestion
      description: Response containing the created workflowData Object.
```

TallTed (changes requested):

exchanges.yml
```diff
@@ -217,17 +217,62 @@ components:
...
       properties:
-        workflowId:
-          type: string
-          description: The URL that uniquely identifies the created workflow.
+        workflowData:
```
dlongley, line 228: Hmm, I think just the workflow `config` should be returned here and it looks like this is something a little different with some additional metadata is here. Certainly an implementation could do more, but I think we should only say that the workflow config is returned (i.e,. the same thing you'd get if you did a GET on the workflow ID).

```suggestion
        config:
```

exchanges.yml
```diff
@@ -217,17 +217,62 @@ components:
...
+          description: Object containing information about a created workflow.
+          properties:
+            id:
+              type: string
+              description: The URL that uniquely identifies the created workflow.
+            stepInformation:
+              type: object
+              description: Information about the steps required for the workflow. Returning stepInformation is REQUIRED.
+              properties:
+                exchanges:
```
dlongley, line 240: I think this could be a potentially huge number of exchanges for a very popular workflow -- and it isn't part of the config, so we shouldn't do this as a requirement, though some implementation might return this kind of data if that is desirable/not problematic.

exchanges.yml
```diff
@@ -217,17 +217,62 @@ components:
...
+                  properties:
+                    stepName:
+                      type: object
+                      description: The name of the step.
+                      properties:
+                        step:
+                          $ref: "#/components/schemas/WorkflowStep"
+                initialStep:
+                  type: string
+                  description: The step from the above set that the exchange starts on.
```
dlongley, line 257: I think the `stepInformation` bit should be collapsed to what's in the config, i.e., I think `steps` and `initialStep` are nested here when they shouldn't be. I will have to look at the config section to ensure we don't do that over there too. See note on "current exchanges".

exchanges.yml
PatStLouis: This is a copy of what is returned by the [GET workflowId](https://w3c-ccg.github.io/vc-api/#get-workflow-configuration) as you are suggesting with the addition of the workflowId itself under the 'id' property as mentioned in the original issue.

exchanges.yml
PatStLouis: This is a copy of the config of an existing workflow returned by the get `workflowId` route, could you provide an example of what the config should look like if this is not it?

PatStLouis (commented):
Happy to make the required changes if I can get some examples provided.

exchanges.yml
dlongley: Oh, I see we have a couple of mistakes in what's in the spec now then. There should be no `workflowData` property when creating a workflow, rather, the properties here:

https://github.com/w3c-ccg/vc-api/blob/5efdafd6f096efd6160f97a69448208c9297baf1/exchanges.yml#L190-L222

Should just be the main top-level properties in the payload, this is the "config" for the workflow.

I also see that the `GET workflowId` section is inconsistent and adds that extra `stepInformation` bit -- which we should remove (collapse it down so its nested properties are moved up a level so it matches what should be in the above payload for creating the workflow).

exchanges.yml
```diff
@@ -217,17 +217,57 @@ components:
...
-          description: The URL that uniquely identifies the created workflow.
+        workflowData:
+          type: object
+          additionalProperties: false
+          description: Object containing information about a created workflow.
+          properties:
+            id:
+              type: string
+              description: The URL that uniquely identifies the created workflow.
+            stepInformation:
```
msporny, line 236: From the call: Delete `stepInformation`

exchanges.yml
```diff
@@ -217,17 +217,57 @@ components:
...
+                  properties:
+                    stepName:
+                      type: object
+                      description: The name of the step.
+                      properties:
+                        step:
+                          $ref: "#/components/schemas/WorkflowStep"
+                initialStep:
+                  type: string
+                  description: The step from the above set that the exchange starts on.
```
msporny, line 252: Move `steps` and `initialStep` up/left one level.

exchanges.yml
```diff
@@ -217,17 +217,57 @@ components:
...
-            workflowId:
+            id:
               type: string
-              description: the ID that will be used for the created workflow. workflowId is OPTIONAL.
+              description: The ID that will be used for the created workflow. Passing an ID is OPTIONAL.
     CreateWorkflowResponse:
       type: object
       additionalProperties: false
-      description: Object containing information about a created workflow.
+      description: Response containing the created workflowData Object.
```
msporny, line 226: Change `workflowData` -> `config`.

Same changes with `workflowX` -> `X` (in general).

exchanges.yml
```diff
@@ -217,49 +217,81 @@ components:
                     type: string
                   template:
                       type: string
-            workflowId:
+            id:
```
dlongley, line 220: `id` can be optional within the `WorkflowConfig` object when creating a workflow, indicating either a preferred `id` (which some servers may honor, others may reject).

exchanges.yml
```diff
@@ -217,49 +217,81 @@ components:
...
           type: object
           additionalProperties: false
           description: Object containing information about a created workflow.
           properties:
-            stepInformation:
+            exchanges:
+              type: array
+              description: The identifiers of the current exchanges associated with the workflow instance.
+              items:
+                type: string
```
dlongley, line 281: I don't expect this (`exchanges`) to ever be part of a `WorkflowConfig` -- and it may be very large in size anyway.

exchanges.yml
```diff
@@ -217,49 +217,81 @@ components:
...
+                  template:
+                    type: string
+                    description: The template itself.
     GetWorkflowResponse:
       type: object
       additionalProperties: false
       description: Object containing information about a workflow.
       properties:
-        workflowData:
+        config:
```
dlongley, line 272: I expect this endpoint to return the `WorkflowConfig` directly as the response payload here.

msporny:
This PR was merged presuming that a WorkflowConfig object will be created to remove duplication of that object.

PR #411, opened 2024-08-07
Remove fixtures from the vc-api

PatStLouis:
Looking for feedback around missing test cases.

Once we agree on which signature suite to use (my current preference will be `Ed25519Signature2020`), I will go ahead and sign these fixtures with a did key.


msporny (changes requested):
These should go in the BSL test suite, not the VC API specification. Test fixtures should live with the test suite. 

The only exception to that rule are examples that are used in the specification, and ideally, those are dynamically generated using respec-vc.

fixtures/README.md
```diff
@@ -1,7 +1,5 @@
 # Fixtures
 
-This folder contains static test fixtures that are hosted by this test suite.
+This folder contains static BitstringStatusList test fixtures referenced by external test-suites.
 
-## revocationList.json
-
-Is a Verifiable Credential based revocation list conforming to the [RevocationStatusList2020](https://w3c-ccg.github.io/vc-status-rl-2020/) spec, whereby the credential associated to index 1 is revoked and all other credentials described by the other indexes represented in the list are not revoked.
\ No newline at end of file
+All status credentials have the following indexes set to 1 `201` `94023` `57444` and should be used for their respective purposes.
```
TallTed, line 5:
```suggestion
All status credentials should be used for their respective purposes, and have the following indexes set to 1: `201`, `94023`, `57444`.
```

TallTed (changes requested):

PatStLouis:
@msporny noted, I will migrate these. Do we want to leave 1 valid BitstringStatusList credential to use as a uri with examples in the vc-api specification?

msporny:
> @msporny noted, I will migrate these. Do we want to leave 1 valid `BitstringStatusList` credential to use as a uri with examples in the vc-api specification? Otherwise I can remove the directory entirely including the `revocationStatusList` credential present

I don't think we need an example right now, and if we do in the future, we can add it back in. Please remove the entire directory; thank you!

aljones15:
@PatStLouis just message me when this PR is ready for review again and I'll review it then. thanks for the work and sorry I didn't know about the issue with this dir.

PatStLouis:
@msporny @TallTed @aljones15 as per the discussions, I removed the fixtures directory entirely as it's undocumented and unused in the specification.

fixtures/README.md
PatStLouis: Removing this directory, I will apply your suggestion in the bitstring status list test suite repo. Thank you

msporny (approved):

aljones15 (approved):
thanks and sorry about the misunderstanding.

mkhraisha (approved):

TallTed (approved):

msporny (approved):

dlongley (approved):

PR #435, opened 2024-12-10
Add Options Extensibility section

jrhender:
Adds a new "Options Extensiblity" sub section with the information discussed here: https://github.com/w3c-ccg/vc-api/issues/335#issuecomment-1478481131

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/pull/435.html" title="Last updated on Dec 17, 2024, 8:47 PM UTC (f582120)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/435/5aff850...f582120.html" title="Last updated on Dec 17, 2024, 8:47 PM UTC (f582120)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -693,17 +693,17 @@ properties.
       </p>
       <h4>Options Extensibility</h4>
       <p>
-Implementations may extend an `options` object with additional properties.
-Property keys may be any string but they should be appropriately namespaced (for example by using URIs)
+Implementations MAY extend an `options` object with additional properties.
+Property keys can be any string but they should be appropriately namespaced (for example by using URIs)
 in order to avoid collisions with extensions in other implementations.
       </p>
       <p>
-As extension properties are implementation specific, they should not be mandatory.
+As extension properties are implementation specific, they ought not be mandatory.
 This is to maintain interoperability by avoiding clients needing to be modified to use a specific implementation.
       </p>
       <p>
-When adding an extension `options` property, consider if providing optionality to clients is necessary.
-If not, using instance configuration to vary API functionality may be a preferable approach.
+When adding an extension `options` property, consider whether providing optionality to clients is necessary.
+If not, using instance configuration to vary API functionality might be a preferable approach.
       </p>
     </section>
     <section>
--- a/index.html
+++ b/index.html
@@ -691,6 +691,20 @@ properties in order to prevent clients from passing certain data to that instanc
 Likewise, an instance configuration MAY require that clients include some `options`
 properties.
       </p>
+      <h4>Options Extensibility</h4>
+      <p>
+Implementations may extend an `options` object with additional properties.
+Property keys may be any string but they should be appropriately namespaced (for example by using URIs)
+in order to avoid collisions with extensions in other implementations.
+      </p>
+      <p>
+As extension properties are implementation specific, they should not be mandatory.
+This is to maintain interoperability by avoiding clients needing to be modified to use a specific implementation.
+      </p>
+      <p>
+When adding an extension `options` property, consider if providing optionality to clients is necessary.
+If not, using instance configuration to vary API functionality may be a preferable approach.
+      </p>
     </section>
     <section>
       <h3>Content Serialization</h3>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -691,6 +691,20 @@ <h3>Options</h3>
 Likewise, an instance configuration MAY require that clients include some `options`
 properties.
       </p>
+      <h4>Options Extensibility</h4>
+      <p>
+Implementations may extend an `options` object with additional properties.
+Property keys may be any string but they should be appropriately namespaced (for example by using URIs)
```
dlongley, line 697:
```suggestion
Implementations MAY extend an `options` object with additional properties.
Property keys can be any string but they should be appropriately namespaced (for example by using URIs)
```

index.html
```diff
@@ -691,6 +691,20 @@ <h3>Options</h3>
...
+in order to avoid collisions with extensions in other implementations.
+      </p>
+      <p>
+As extension properties are implementation specific, they should not be mandatory.
```
dlongley, line 701:
```suggestion
As extension properties are implementation specific, they ought not be mandatory.
```

index.html
```diff
@@ -691,6 +691,20 @@ <h3>Options</h3>
...
+This is to maintain interoperability by avoiding clients needing to be modified to use a specific implementation.
+      </p>
+      <p>
+When adding an extension `options` property, consider if providing optionality to clients is necessary.
+If not, using instance configuration to vary API functionality may be a preferable approach.
```
dlongley, line 706:
```suggestion
If not, using instance configuration to vary API functionality might be a preferable approach.
```

dlongley (changes requested):
Some suggested normative-word language tweaks.

TallTed (changes requested):
One change in addition to those from @dlongley.

index.html
```diff
@@ -691,6 +691,20 @@ <h3>Options</h3>
 Likewise, an instance configuration MAY require that clients include some `options`
 properties.
       </p>
+      <h4>Options Extensibility</h4>
+      <p>
+Implementations MAY extend an `options` object with additional properties.
+Property keys can be any string but they should be appropriately namespaced (for example by using URIs)
+in order to avoid collisions with extensions in other implementations.
```
dlongley, line 698: I don't know if we need to provide this advice or if it should be adjusted somewhat. We might want to discuss this bit on the next call.

It seems the danger here would actually be with a conflict with something that gets standardized, not with different implementations. This is because if a different implementation were to be swapped in, you'd certainly first stop using any previous-implementation-specific options, regardless of their names. You'd then add any new-implementation-specific options (and it then wouldn't matter whether the same names were used or not). In no case would you be using both implementations at once (or not know which one you were talking to).

dlongley (approved):
Approving, though I think we should discuss the subject of the one comment I made on a call.

index.html
TallTed: Until @dlongley's concern above is addressed, this text should be refined a bit, a la --
```suggestion
Property keys can be any string, but they should be appropriately namespaced (by using URIs, for example)
to avoid collisions with other extensions.
```

msporny: Yeah, I hesitate at saying that the extensibility mechanism is to use full URIs as well. Feels like just text strings could work fine here, and perhaps a registry? (though I shudder at the need for a registry)

TallTed: (also, I left a `should` in there, but if it remains, it should be change to `SHOULD` or `ought to`)

TallTed (changes requested):
See https://github.com/w3c-ccg/vc-api/pull/435/files#r1889145522

index.html
TallTed:
```suggestion
Property keys can be any string, but they SHOULD be appropriately namespaced (by using URIs, for example)
to avoid collisions with other extensions.
```
```suggestion
Property keys can be any string but they should be appropriately namespaced (for example by using URIs)
in order to avoid collisions with extensions in other implementations.
```

TallTed: or
```suggestion
Property keys can be any string but they ought to be appropriately namespaced (for example by using URIs)
in order to avoid collisions with extensions in other implementations.
```

msporny, line 698:
```suggestion
```

TallTed: wfm

TallTed (approved):

PR #443, opened 2025-01-29
Elaborate on service configuration

### LINKED ISSUE #267: VC API Issuer - Greenfield Excercise
@dlongley I get the impression from your comments that it might be easier to see our perspective if we could see what you wish the VC API Issuer endpoints looked like, if we were to build them from scratch, focused only on your "business use cases" and not at all focused on "demonstrating interop".

Would you be willing to sketch out in comments what you would build if we could start over here?

Addressing issues like: 

- How many endpoints?
- How many issuers / dids / keys / suites?
- Optional or required fields related to extensions (like revocation)
- API Security

We might be able to leap frog the current APIs with a clearer picture of your complete vision, and I am concerned we won't be able to make good progress trying incrementally adjust from where we are today.

msporny:
This PR attempts to address issue #267 by documenting how service instance configuration works.

Fixes #267.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/pull/443.html" title="Last updated on Feb 18, 2025, 2:19 PM UTC (55c75d5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/443/ace3da6...55c75d5.html" title="Last updated on Feb 18, 2025, 2:19 PM UTC (55c75d5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -495,7 +495,7 @@ can run their own Workflow Service to create and manage exchanges that
 realize particular workflows. Administrators configure the workflow system
 to support particular flows. Then, when the business rules justify it,
 coordinators create exchanges at their Workflow Service and give authorized
-access to those exchanges to any party. 
+access to those exchanges to any party.
         </p>
       </section>
       <section>
@@ -577,7 +577,7 @@ might contain hostile actors. As a result, conforming <a>VC API servers</a>
 require conforming <a>VC API clients</a> to utilize secure authorization
 technologies when performing certain types of requests. Each HTTP endpoint
 defined in this document specifies whether or not authorization is required
-when performing a request. With the exception of the class of forbidden 
+when performing a request. With the exception of the class of forbidden
 authorization protocols discussed later in this section, the VC API is agnostic
 regarding authorization mechanism.
       </p>
@@ -619,7 +619,7 @@ assurance is achieved in the flow to properly protect the binding.
         </li>
       </ul>
       <p>
-The rest of this section gives examples of the authorization technologies that have 
+The rest of this section gives examples of the authorization technologies that have
 been contemplated for use by conforming implementations. Other equivalent authorization
 technologies can be used. Implementers are cautioned against using non-standard
 or legacy authorization technologies.
@@ -641,7 +641,7 @@ HTTP Basic Authentication [[RFC7617]].
 If the OAuth 2.0 Authorization Framework [[RFC6749]] is utilized for authorization,
 the access tokens utilized by clients MAY be OAuth 2.0 Bearer Tokens [[RFC6750]]
 or any other valid OAuth 2.0 token type. Any valid OAuth 2.0 grant type MAY be used
-to request the access tokens. However, OAuth 2.0 MUST be implemented in the following 
+to request the access tokens. However, OAuth 2.0 MUST be implemented in the following
 way:
         </p>
         <p>
@@ -707,7 +707,7 @@ If not, using instance configuration to vary API functionality might be a prefer
     <section>
       <h3>Content Serialization</h3>
       <p>
-All entity bodies in requests and responses sent to or received from the API endpoints defined by this specification MUST be serialized as JSON and include the `Content-Type` header with a media type value of `application/json`. 
+All entity bodies in requests and responses sent to or received from the API endpoints defined by this specification MUST be serialized as JSON and include the `Content-Type` header with a media type value of `application/json`.
       </p>
     </section>
     <section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -671,16 +671,36 @@ <h4>OAuth 2.0</h4>
     <section>
       <h3>Service Instances</h3>
       <p>
-Service configuration is determined by administrators on per instance basis.
-Instance configuration can include, but is not limited to, credential format, key options, credential status mechanisms, and/or credential templates.
-Administrators are expected to configure service instances such that `options` included in client requests cannot result in incorrect action or problematic responses by the service.
+The APIs defined in this specification presume that they are attached to a
+specific instance of some kind with an associated configuration that has been
```
jrhender, line 675: To me, the "of some kind" text is superfluous
```suggestion
specific instance with an associated configuration that has been
```

index.html
```diff
@@ -671,16 +671,36 @@ <h4>OAuth 2.0</h4>
...
+put in place by a systems administrator. When a client calls an endpoint on a
+particular instance, the instance uses the configuration and options provided by
+the client to execute the action.
       </p>
       <p>
-A coordinator instance can have access to multiple service instances in order to support different use cases.
-Runtime discovery of service instance configuration is not supported by the VC API as services are expected to be known by the coordinator at the time of coordinator deployment.
+For example, the `/credentials/issue` endpoint is expected to be at the end of a
+longer URL such as `/instances/12345/credentials/issue`. In this case, it is the
```
jrhender, line 682: Does it need to be at the end of longer URL? It seems reasonable to me that one could direct to specific instances in other ways. For example, one could use subdomains to address an instance or an internal service discovery mechanism.

index.html
```diff
@@ -671,16 +671,36 @@ <h4>OAuth 2.0</h4>
...
+provided by implementations but are not necessary to expose as HTTP APIs;
+configuration can also be done through configuration files or graphical
+interfaces.
+      </p>
+      <p class="note"
+         title="Use cases can use multiple instances for complex flows">
+A coordinator instance can have access to multiple service instances in order to
+support different use cases. Runtime discovery of service instance configuration
+is not supported by the VC API as services are expected to be known by the
+coordinator at the time of coordinator deployment.
```
jrhender, line 700: IMO, the current title is too narrow. In my mind, a single coordinator could support multiple use cases and so might need multiple service instances, even the each use case flow is "simple". So having the title suggest that the main use for multiple instances is "complex" flows is too narrow.
```suggestion
      <p class="note"
         title="A coordinator can use multiple service instances">
A coordinator instance can have access to multiple service instances in order to
support different use cases or a use case with complex flows. Runtime discovery of service instance configuration
is not supported by the VC API as services are expected to be known by the
coordinator at the time of coordinator deployment.
```

jrhender (commented):
Thanks for the PR. I'd been wondering how to proceed on https://github.com/w3c-ccg/vc-api/issues/419 and maybe this PR helps move that issue forward as well.

index.html
```diff
@@ -671,16 +671,36 @@ <h4>OAuth 2.0</h4>
...
+The APIs defined in this specification presume that they are attached to a
+specific instance of some kind with an associated configuration that has been
+put in place by a systems administrator. When a client calls an endpoint on a
+particular instance, the instance uses the configuration and options provided by
+the client to execute the action.
       </p>
       <p>
-A coordinator instance can have access to multiple service instances in order to support different use cases.
-Runtime discovery of service instance configuration is not supported by the VC API as services are expected to be known by the coordinator at the time of coordinator deployment.
+For example, the `/credentials/issue` endpoint is expected to be at the end of a
```
dlongley, line 681:
```suggestion
For example, the `/credentials/issue` endpoint can be provided at the end of a
```

index.html
```diff
@@ -671,16 +671,36 @@ <h4>OAuth 2.0</h4>
...
+whether or not a status list is involved, the type of credential to issue, the
+credential format, and what additional options are possible on the endpoint.
+      </p>
+      <p>
+Software clients that call a particular instance might not have the capability
+to configure an instance, or be aware of the setup that the administrator did on
+the instance. Administration endpoints for configuring instances could be
+provided by implementations but are not necessary to expose as HTTP APIs;
+configuration can also be done through configuration files or graphical
+interfaces.
```
dlongley, line 693: Clients do need to know what is expected to happen when the admin hands them the URL, but otherwise, yes:

```suggestion
to configure an instance, or be aware of the setup that the administrator did on
the instance other than the requisite details to make appropriate use of it.
Administration endpoints for configuring instances could be
provided by implementations but are not necessary to expose as HTTP APIs;
configuration can also be done through configuration files or graphical
interfaces.
```

index.html
dlongley: We might also want to say that "Using multiple service instances, such as different workflows, can also be used to improve a deployment's security profile by limiting the privileges that a coordinator needs to operate. One way this can be done is by restricting the inputs a coordinator can provide when creating an exchange to just a few variables in a workflow-specific template." Or something along these lines.

dlongley (approved):
Approving with some suggestions.

index.html
dlongley: +1, I made a suggestion in my review.

TallTed (changes requested):

Issue #424, opened 2024-10-01
Clarify how EnvelopedVerifiableCredential and JWTs can be verified


robdefeo:
As per https://www.w3.org/TR/vc-jose-cose/#securing-vps-with-jose I can see that the request is valid for `/presentations/verify`

> ```
> {
>   "@context": [
>     "https://www.w3.org/ns/credentials/v2",
>     "https://www.w3.org/ns/credentials/examples/v2"
>   ],
>   "type": "VerifiablePresentation",
>   "verifiableCredential": [{
>     "@context": ["https://www.w3.org/ns/credentials/v2"],
>     "type": ["EnvelopedVerifiableCredential"],
[...]

PR #461, opened 2025-02-24
Verify Enveloped Credentials and Enveloped Presentations

jrhender:
Partially addresses #424 . The objective of the PR is to add support for verifying EnvelopedVCs and EnvelopedVPs. This is done by adding enveloped VCs and VPs as possible schemas to the credentials/verify and persentations/verify endpoints respectively.

### DIFF
--- a/respec-oas.js
+++ b/respec-oas.js
@@ -87,42 +87,49 @@ function buildEndpointDetails({config, document, apis}) {
 
     // schema for endpoint
     if(endpoint.requestBody) {
-      const requestSchema =
-        endpoint.requestBody.content['application/json'].schema.properties ||
-        endpoint.requestBody.content['application/json'].schema;
+      const contentTypes = Object.keys(endpoint.requestBody.content);
 
-      const schemaSummary = document.createElement('p');
-      if(requestSchema.anyOf) {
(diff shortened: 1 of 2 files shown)


msporny (approved):

msporny:
The group discussed this on the 2025-02-25 call and found that we do not expect `application/vc` or `application/vp` to be posted instead of `application/json`, but do expect `oneOf` the JSON Schemas listed to be used. The media type for the request will only ever be `application/json` with different schemas used when VCs or VPs are POSTed.

The PR needs to be updated so that the `application/vc` or `application/vp` media types aren't allowed for posting, only `application/json` and any of the schemas should be possible (based on the call being made).

msporny (changes requested):
The endpoints will receive `application/json` always -- any use of `application/vc` or `application/vp` are not allowed (even though VCs or VPs posted inside the JSON body are allowed/expected)

jrhender:
> The PR needs to be updated so that the application/vc or application/vp media types aren't allowed for posting, only application/json and any of the schemas should be possible (based on the call being made).

I made this change.

dlongley (approved):

msporny (approved):

Issue #438, opened 2024-12-17
Update /credential/status to use latest implementation feedback


msporny:
This issue is to track updating the `/credential/status` endpoint. For example:

https://gist.github.com/dlongley/165e23bad840edcec07394f9546e4503

Other implementations might be doing different things. This issue is to track what implementations are doing.

msporny:
Addressed by merging PR #462.

PR #462, opened 2025-02-25
Update credential status endpoint

PatStLouis:
To address #438

Creates a new schema for the UpdateCredentialStatusRequest payload.

Questions:
what value should the `credentialId` represent? `credential.id` or `credentialId`?


issuer.yml
```diff
@@ -72,6 +72,43 @@ components:
...
+            statusPurpose:
+              type: string
+            statusListIndex:
+              type: string
+            statusListCredential:
+              type: string
+        status:
+          type: boolean
+          description: Specifies the new status.
+        indexAllocator:
```
msporny, line 103: This feature needs to be explained more clearly in the specification (in a future PR).

msporny (approved):
LGTM

Issue #452, opened 2025-02-17
Potentially unnecessary intermediate `config` property in workflows


kezike:
I noticed that there is an intermediate `config` property in the [workflow creation](https://w3c-ccg.github.io/vc-api/#create-workflow) request and response, as well as the [workflow retrieval](https://w3c-ccg.github.io/vc-api/#get-workflow-configuration) response. Was this nesting intentional? Do we anticipate adding other properties to this schema? If not, this property may be unnecessary.

dlongley:
No, the nesting is not intentional and should be fixed to avoid any confusion. The post body *is* the configuration. It should *not* be a post body that has a `config` key in it that then has the config object as its value.

msporny:
The group discussed this on 2025-02-25:

A PR should be raised that removes the `config` key and moves the config object up to be the value that is posted.

msporny:
Fixed by merging #463.

PR #463, opened 2025-03-04
Remove "config" property from workflow definitions.

kezike:
This PR addresses Issue #452 by removing unnecessary intermediate "config" property in workflow request/response body definitions.

**NOTE:** I am still seeing the `config` property in the OpenAPI spec when I run it locally. Anyone know why that might be?

### DIFF
--- a/exchanges.yml
+++ b/exchanges.yml
@@ -183,133 +183,118 @@ components:
       additionalProperties: false
       description: Object containing information for creating a workflow.
       properties:
-        config:
+        steps:
           type: object
-          additionalProperties: false
-          description: Object containing information for creating a workflow.
+          description: One or more steps required to complete an exchange on the workflow. Passing the steps object is REQUIRED.
           properties:
-            steps:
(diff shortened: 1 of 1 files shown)


exchanges.yml
```diff
@@ -183,133 +183,118 @@ components:
...
       type: object
       additionalProperties: false
       description: Object containing information about a workflow.
       properties:
-        config:
+        exchanges:
+          type: array
+          description: The identifiers of the current exchanges associated with the workflow instance.
+          items:
+            type: string
```
dlongley, line 266: This should be removed, returning all of the current exchanges isn't scalable -- and I don't think this is readily implemented either.

dlongley (changes requested):

exchanges.yml
dlongley, line 266:
```suggestion
```

msporny (approved):

dlongley (approved):

Issue #453, opened 2025-02-17
Change `exchangeId` to `id` in exchange creation response


kezike:
I noticed that the [exchange creation endpoint](https://w3c-ccg.github.io/vc-api/#create-exchange) section specifies an `exchangeId` property in the response. I believe this should be `id` (this issue is an exchange analog to [this workflow issue](https://github.com/w3c-ccg/vc-api/issues/393))?

dlongley:
Yes, I believe there's an error here. The general rule is that the key name should be `id` if it is scoped within the exchange itself  and `exchangeId` if it is not within the exchange object itself but is being referenced by some other object.

The exchange creation endpoint should return a `location` header with the exchange ID as its value and either return a `204` response code (no content) or a `200` with the exchange information, whereby the `exchangeId` key should be `id` as stated above. Alternatively, the exchange could be returned underneath an `exchange` property to allow for other metadata to be returned -- this might be better. Either way the `id` key name would be used inside the exchange object.

msporny:
The group discussed this on the 2025-02-25 telecon: 

A PR should be raised that applies the following change: https://github.com/w3c-ccg/vc-api/issues/453#issuecomment-2663880455

dlongley:
The `id` description in the table here: https://w3c-ccg.github.io/vc-api/#get-exchange-state

...should also be fixed to say it's the full exchange ID.

msporny:
A PR should be raised to not use `exchangeId` as the property, but rather the `id` property instead.

kezike:
> A PR should be raised to not use `exchangeId` as the property, but rather the `id` property instead.

@msporny I must have missed the meeting where this was discussed, but this has already been addressed by this PR. The last few comments that Dave and I made before this comment were related to the content of the `id` field, which I still believe should be the local exchange ID.

PR #468, opened 2025-03-11
Embed exchange creation response in `exchange` property.

kezike:
This PR addresses Issue #453 by embedding the exchange creation response in the `exchange` property and renaming `exchangeId` to `id`.

@dlongley As it was [your recommendation](https://github.com/w3c-ccg/vc-api/issues/453#issuecomment-2663880455) to return the exchange in `exchange` during exchange creation (which is apparently consistent with the exchange retrieval response format), should I create an issue to do the same for workflow creation? If so, it seems the general pattern will be to include the body parameters directly in the request, and return it in an intermediate property in the response (e.g., `exchange`, `workflow`). Before we move forward, it would be good to confirm this pattern generally.

### DIFF
--- a/exchanges.yml
+++ b/exchanges.yml
@@ -374,8 +374,8 @@ components:
               type: string
               description: The local exchange ID that identifies the exchange.
             sequence:
-              type: string
-              description: A sequence number for the exchange. Set to "0" on creation.
+              type: number
+              description: A sequence number for the exchange. Set to 0 on creation.
             ttl:
               type: string
               description: The time to live for the exchange.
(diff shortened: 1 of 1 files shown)


msporny (changes requested):
The group discussed this on the 2025-03-11 telecon and requested the following changes:

* Do not return the `exchange` object when creating an exchange
* Specify that the newly created exchange location is provided via the `Location` HTTP header.
* Remove the nesting on the `exchange` object when retrieving an exchange via HTTP GET: https://github.com/w3c-ccg/vc-api/pull/468/files#diff-ee115e35c724c00de39597365d304a22894c3bc88f82a03ae205d08b8bd5b2b2R373

We need to make sure this pattern is followed by the workflow creation endpoint.

msporny (approved):

dlongley (approved):

Issue #421, opened 2024-09-13
Add `<fullExchangeId>/protocols` endpoint that serves interaction "protocols" object


dlongley:
Performing an HTTP GET on an exchange URL + `/protocols` should return a "protocols" object with the supported protocols for interacting with that exchange. For example:

```http
GET /workflows/abc/exchanges/123/protocols

{
  "protocols": {
    "vcapi": "https://vcapi.example/workflows/abc/exchanges/123",
    "OID4VCI": "openid-credential-offer://?credential_offer_uri=https%3A%2F%2Fvcapi.example%2Fworkflows%2Fabc%2Fexchanges%2F123%2Fopenid%2Fcredential-offer"
  }
}
```

dlongley:
We should also document that a coordinator website could serve any route (of their choice) from their origin that they can send via QR code or via CHAPI as an "interact" protocol:

```js
const result = await navigator.credentials.get({
  web: {
    protocols: {
      // the "interact" protocol returns more "protocols" from a VC API exchange
      interact: '<any with origin of coordinator site>'
    }
  }
});
```

That way an exchange client (e.g., digital wallet) can use origin-based security to get the list of protocol URLs directly from the coordinator website, allowing trust to be anchored in the coordinator origin, even if the coordinator serves a protocols object that points at an exchange that is hosted by some other origin.

dlongley:
Notably, if `protocols.vcapi` is given directly in a CHAPI request you can also append `/protocols` to it to get all supported protocols, so we might just want to keep doing this and not bother adding the `interact` option.

msporny:
The group discussed this on 2024-09-17:

@dlongley provided a background on what the change would do (see above). There didn't seem to be objections to the endpoint or the feature set. A PR should be raised to add this endpoint to the specification.

msporny:
A coordinator creates an exchange, the exchange always has the protocols endpoint, the coordinator can expose the endpoint on a system of their own that can proxy what the exchange returns back to the client that made the request. That ensures that the client (digital wallet) is retrieving the request through the origin that the end user is expected to trust and/or understand. That eliminates the need for doing anything like a signed authorization request (because you trust the domain) -- no keys or signing is required by the hosting domain because the trust comes from the domain certificate used to serve the .../protocols endpoint (regular trust model provided by DNS and TLS).

dlongley:
@kezike,

[...]

PR #473, opened 2025-04-09
Document interaction initiation

msporny:
This PR documents interaction initiation, which can be used by any entity (such as a holder) to initiate an interaction with another entity (such as a verifier). The mechanism is protocol agnostic and utilizes a URL, QR code, and minimal JSON objects to bootstrap into more advanced protocols.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/pull/473.html" title="Last updated on Apr 30, 2025, 2:44 PM UTC (917eadc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/473/9aa0e53...917eadc.html" title="Last updated on Apr 30, 2025, 2:44 PM UTC (917eadc)">Diff</a>


index.html
```diff
@@ -1355,6 +1367,183 @@ <h4>Exchange Examples</h4>
...
+expressed as a QR code according to [[[ISO18004]]]. To ensure broad
+interoperability, the length of the [=interaction URL=] SHOULD be as short as
+possible, SHOULD NOT exceed 400 alphanumeric characters, and MUST NOT exceed
+4,296 alphanumeric characters. An example of an interaction QR code can be found
+below:
+        </p>
+
+        <figure id="roles">
+          <img style="margin: auto; width: 33%; display: block;"
+         src="./diagrams/interaction-qr-code.png"
```
davidlehn, line 1454: Missing from PR.

index.html
```diff
@@ -1355,6 +1367,183 @@ <h4>Exchange Examples</h4>
...
+Performing a retrieval of the [=interaction URL=] results in a list of
+protocols that the local system can use to interact with the remote
+system. For example, performing an HTTP GET on the
+`https://app.example/interactions/z8n38Dp7a?iuv=1` [=interaction URL=] might
+result in the following response:
+        </p>
+
+        <pre class="example nohighlight" title="A list of protocols supported for a given interaction">
+{
+  protocols: {
```
davidlehn, line 1476:
```suggestion
  "protocols": {
```

index.html
```diff
@@ -1355,6 +1367,183 @@ <h4>Exchange Examples</h4>
...
+Bluetooth (wireless medium).
+      </p>
+
+      <section>
+        <h4>Interaction URL Format</h4>
+        <p>
+The format for the <dfn>interaction URL</dfn> MUST conform to the syntax for the
+[[[URL]]]. The interaction URL SHOULD be an HTTPS URL where the path segment
+ends with the `/interactions/` path, a transaction-specific `interactionId`, and
+the interaction URL version number which MUST be `iuv=1`. The URL SHOULD be
```
davidlehn, line 1425: Should this be more strict and describe interactionId as as a URL path segment and iuv=1 as a query param?  (use the proper terminology/links)

index.html
msporny: Added in https://github.com/w3c-ccg/vc-api/pull/473/commits/4675a21f1d9279c63892ab080d2d83ef3d166fe3.

index.html
```diff
@@ -1058,28 +1070,28 @@ <h3>Workflows and Exchanges</h3>
...
-workflow. The interaction will take place between an exchange client and the
-workflow service. Exchanges are expected to be transitory, only existing as
-long as the interaction takes to complete. The workflow service stores state
-information about each exchange such as whether the exchange is pending,
-active, or complete, as well as the current step in the workflow, any
+A VC API <dfn>exchange</dfn> represents a particular interaction based on a
+given VC API workflow. The interaction will take place between an exchange
+client and the workflow service. Exchanges are expected to be transitory, only
+existing as long as the interaction takes to complete. The workflow service
+stores state information about each exchange such as whether the exchange is
```
TallTed, line 1093:
```suggestion
stores state information about each exchange, such as whether the exchange is
```

index.html
```diff
@@ -1355,6 +1367,183 @@ <h4>Exchange Examples</h4>
       </section>
     </section>
 
+    <section>
+      <h3>Initiating Interactions</h3>
+      <p>
+It is useful for an implementation to communicate how to start interacting with
+it to another implementation. This bootstrapping process is called
+<em>initiating an <dfn>interaction</dfn></em>, which communicates what protocols
```
TallTed, line 1375:
```suggestion
<em>initiating an <dfn>interaction</dfn></em>, and communicates what protocols
```

index.html
msporny: I've made the update, let me know if the new text works for you.

index.html
```diff
@@ -1355,6 +1367,183 @@ <h4>Exchange Examples</h4>
...
+        </pre>
+        <figcaption>
+A Holder-initiated exchange using a QR Code
+        </figcaption>
+      </figure>
+
+      <p class="note" title="Multiple interaction initiation variations are possible">
+While the example above is a [=holder=]-initiated interaction, [=issuers=] and
+[=verifiers=] are also capable of initiating an interaction. Similarly, while
+the transmission of the interaction URL is performed via a QR Code
```
TallTed, line 1414:
```suggestion
the transmission of the interaction URL here is performed via a QR Code
```

index.html
```diff
@@ -1355,6 +1367,183 @@ <h4>Exchange Examples</h4>
...
+(optical medium), the interaction URL can also be transmitted over NFC or
```
TallTed, line 1415:
```suggestion
(optical medium), an interaction URL can also be transmitted over NFC or
```

index.html
```diff
@@ -1355,6 +1367,183 @@ <h4>Exchange Examples</h4>
...
+Bluetooth (wireless medium).
+      </p>
+
+      <section>
+        <h4>Interaction URL Format</h4>
+        <p>
+The format for the <dfn>interaction URL</dfn> MUST conform to the syntax for the
```
TallTed, line 1422:
```suggestion
The format of the <dfn>interaction URL</dfn> MUST conform to the syntax for the
```

index.html
```diff
@@ -1355,6 +1367,191 @@ <h4>Exchange Examples</h4>
...
+[[[URL]]] and contain an `iuv` query parameter encoding the
+interaction URL version number, which MUST be `1`. The interaction URL SHOULD be
```
TallTed, line 1424: It seems odd to mandate version `1`. Maybe allowance for a greater value will be made later? It does not seem like we're expecting this `interaction URL`'s version to be bumped by later versions of the VC API, but rather by future changes in a given implementation/deployment. But maybe I'm misunderstanding?

index.html
```diff
@@ -1355,6 +1367,191 @@ <h4>Exchange Examples</h4>
...
+
+        <pre class="example nohighlight" title="An interaction URL">
+https://app.example/interactions/z8n38Dp7a?iuv=1
+        </pre>
+      </section>
+
+      <section>
+        <h4>Interaction QR Code Format</h4>
+        <p>
+The format for an <dfn>interaction QR Code</dfn> MUST be an [=interaction URL=]
```
TallTed, line 1439:
```suggestion
An <dfn>interaction QR Code</dfn> MUST be an [=interaction URL=]
```

TallTed (changes requested):
Partial review. I'll pick it up again later.

index.html
```diff
@@ -1355,6 +1367,191 @@ <h4>Exchange Examples</h4>
...
+    "oid4vp": "https://app.example/workflows/567/exchanges/432"
+  }
+}
+        </pre>
+
+        <p>
+When the [=interaction URL=] is fetched using any unrecognized `Accept` header,
+a `text/html` document MUST be returned with directions instructing a human
+being to use specific software that understands how to process interaction URLs.
+        </p>
```
dlongley, line 1490: This section should more closely show how this integrates with VC API exchanges. More specifically, we should say that an interaction server can make its interaction map to an exchange ID -- which can be used to *produce* (on the fly) the `protocols` object. This is because an HTTP GET on `<exchangeId>/protocols` will return the object.

Therefore, a suggested implementation is for an interaction server to respond to interaction URL `X` by finding the associated exchange ID (implementation specific) and then doing an HTTP GET on `exchangeId>/protocols` and returning the result to the client. This allows interaction implementations to be quite simple and integrate nicely with VC API workflow/exchange services.

It also demonstrates how origin trust signals can be used even when a workflow/exchange service might live on a different origin. The interaction server's origin will be seen by the client (e.g., digital wallet) that receives the interaction URL -- and that software can surface this to the user (or do whatever wallet-specific trust things it wants to do).

This means that the wallet needn't be concerned with whatever origin it will subsequently interact with vis-a-vis the workflow/exchange service -- because the interaction server has securely delegated to it. This delegation is secure because the wallet and its user can first approve of the interaction server's origin and securely receive (via TLS) the protocols object that lists those delegates (services). This helps prevent vendor lock-in by pushing the VC API workflow/exchange services behind this interface.

All of this can be done without having to introduce additional encryption or key management overhead. This mechanism continues to work even if the interaction URL travels to another device (e.g., QR code), because the interaction URL points at the origin that is to be considered when making a trust decision, not the swappable workflow/exchange service. This is the first origin that the software that scans the QR code will interact with.

This would also help address: #421.

index.html
msporny: Allowance for a greater value will be made later. We are starting at version 1.

We might bump the interaction URLs version in later versions of the VC API (just like we had `v1` at the end of the context URL for for VCDM v1.0 and v1.1, and `v2` at the end of the context URL for VCDM v2.0.

One of the reasons we put this value in the URL is so that scanning systems have a simple text string that they can look for to determine if what they're looking at might be an interaction URL. Searching for a query parameter with the name `iuv` and `1` as the value gives the local system higher assurance that a GET on that URL will result in a protocols object. It's not guaranteed, but it's better than having nothing to key off of.

There are trade-offs here -- we could have used a custom protocol scheme, but then if we did that, we can't easily load it into a browser in the event that someone scans the QR Code with an app that doesn't understand interaction URLs (such as a camera app in a phone that recognizes QR Codes, but probably doesn't recognize interaction URLs).

index.html
```diff
@@ -1355,6 +1367,199 @@ <h4>Exchange Examples</h4>
...
+
+        <p class="note" title="Mapping to exchange URLs">
+Some coordinator implementations will implement the protocols endpoint as a pass
+through to a protocols endpoint for an exchange instance. For example, a GET on
+`https://app.example/interactions/z8n38Dp7a?iuv=1` will result in a pass-through
+GET on `https://app.example/workflows/123/instances/987/protocols`, which would
```
dlongley, line 1495:
```suggestion
GET on `https://app.example/workflows/123/exchanges/987/protocols`, which would
```

index.html
```diff
@@ -1355,6 +1367,199 @@ <h4>Exchange Examples</h4>
...
+            "https://www.w3.org/ns/credentials/v2",
+            "https://www.w3.org/ns/credentials/examples/v2",
+          ],
+          "type": "StudentIdCredential",
+          "credentialSubject": {
+            "studentId": ""
+          },
+        },
+        "trustedIssuer": [{
+          "required": true,
```
dlongley, line 1548: IIRC, "required" is not necessary here; if trusted issuers are provided then those are the only ones that will be accepted.

```suggestion
```

jandrieu:
Between steps 8 and 9, perhaps find a way to indicate that the next part of the flow is the Go To Website protocol (rather than VC-API, OID4, NFC, Bluetooth, etc.)

index.html
```diff
@@ -1355,6 +1367,199 @@ <h4>Exchange Examples</h4>
...
+example, performing an HTTP GET on the
+`https://app.example/interactions/z8n38Dp7a?iuv=1` [=interaction URL=] might
+result in the following response:
+        </p>
+
+        <pre class="example nohighlight" title="A list of protocols supported for a given interaction">
+{
+  "protocols": {
+    "website": "https://app.example/redirects/z8j3kfk2lQ",
+    "vc-api": "https://app.example/workflows/123/exchanges/987",
```
dlongley, line 1479: This might be just "vcapi" in existing implementations, we should check those and match what they are doing.

eric-schuh:
We should add a comment to the PR to call out that this "interaction" mechanism is not itself specific to VCs or the VC-API and can be generalized to support any interaction protocol between any two applications that want to begin an interaction. In the future there may be a better spec or standalone spec for this interaction mechanism, though the VC-API spec seems a good spot to get something like this started.

index.html
```diff
@@ -1355,6 +1367,199 @@ <h4>Exchange Examples</h4>
...
+(optical medium), an interaction URL can also be transmitted over NFC or
+Bluetooth (wireless medium).
+      </p>
+
+      <section>
+        <h4>Interaction URL Format</h4>
+        <p>
+The format of the <dfn>interaction URL</dfn> MUST conform to the syntax for the
+[[[URL]]] and contain an `iuv` query parameter encoding the interaction URL
+version number, which MUST be `1`. The interaction URL SHOULD be an HTTPS URL
```
dlongley, line 1424: The rationale for HTTPS should be explained here -- which includes all of the trust signals and benefits with origins -- and how delegation to other service providers can happen simply by retrieving the protocols object via TLS from the trusted origin (over HTTPS). This helps trust, it helps prevent vendor lock, and it does it simply. Any other URL scheme that might be considered should know about this / maybe make a statement about whether or not these same features are supported.

index.html
```diff
@@ -1355,6 +1367,199 @@ <h4>Exchange Examples</h4>
...
+      <h3>Initiating Interactions</h3>
+      <p>
+It is useful for an implementation to communicate how to start interacting with
+it to another implementation. This bootstrapping process is called
+<em>initiating an <dfn>interaction</dfn></em>, and communicates what protocols
+each implementation supports as well as how to start a particular interaction
+with the implementation.
+      </p>
+      <p>
+A high-level overview of a Holder initiating an interaction using a QR Code and
```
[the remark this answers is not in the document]
msporny: I'm guessing some of the folks more familiar with OID4 will object to this flow as "unsafe" (because of flaws in the OID4 scheme/QR Code approach)... even though the interaction URL flow does not suffer from the same flaws because it depends on HTTPS domain security for the initiation of the flow. I'll try to add a note pointing that out (as one of the other comments in this thread tries to do).

index.html
msporny: Looks likethe code is currently `vcapi`, `OID4VCI`, and `OID4VP`:

https://github.com/digitalbazaar/bedrock-vc-delivery/blob/b3aa0d178d5c12bfcb633fee8e062c19705efe1a/lib/http.js#L146

We should probably make it all lowercase in the spec and make matching case-insensitive in the code.

index.html
```diff
@@ -1355,6 +1367,198 @@ <h4>Exchange Examples</h4>
...
+example, performing an HTTP GET on the
+`https://app.example/interactions/z8n38Dp7a?iuv=1` [=interaction URL=] might
+result in the following response:
+        </p>
+
+        <pre class="example nohighlight" title="A list of protocols supported for a given interaction">
+{
+  "protocols": {
+    "website": "https://app.example/redirects/z8j3kfk2lQ",
+    "vc-api": "https://app.example/workflows/123/exchanges/987",
```
msporny, line 1479:
```suggestion
    "vcapi": "https://app.example/workflows/123/exchanges/987",
```

index.html
mkhraisha: I think i'm getting lost on the suggestion here. The text of the PR is clear to me, what are you suggesting it is changed to?
As someone who wasn't on the call when this was discussed, and is just reading the text of the spec/PR I think theres some nuance I'm missing.

mkhraisha (approved):

index.html
```diff
@@ -1376,9 +1376,54 @@ <h3>Initiating Interactions</h3>
 each implementation supports as well as how to start a particular interaction
 with the implementation.
       </p>
+      <p class="note" title="Interactions are application, use case, and protocol agnostic">
+While the specification of interactions reside in this document, the
+general approach is use case, application, and protocol agnostic. The approach
+can be used to pair two or more applications that desire to bootstrap into a
+particular protocol over any transmission medium such as a web browser,
+QR Code (optical medium), or NFC (wireless medium) where the protocol does not
+need to involve the VC API.
```
TallTed, line 1385:
```suggestion
While several interaction specifications reside in this document, the general
approach is agnostic as to use case, application, and protocol. This approach
can be used to pair two or more applications that desire to bootstrap into a
particular protocol over any transmission medium &mdash; such as a web browser,
QR Code (optical medium), or NFC (wireless medium) &mdash; where the protocol
does not need to involve the VC API.
```

index.html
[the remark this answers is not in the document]
msporny: Fixed in abad5e5155e16bd6bd29c5babc0ca74078ec303b

index.html
msporny: "Fixed" for now in f980030fdfc2704ee848b1ac3e2a01e57449466f.

msporny:
> We should add a comment to the PR to call out that this "interaction" mechanism is not itself specific to VCs or the VC-API and can be generalized to support any interaction protocol between any two applications that want to begin an interaction. In the future there may be a better spec or standalone spec for this interaction mechanism, though the VC-API spec seems a good spot to get something like this started.

Fixed in 6d2620fdf53785fa5ce81cc900e4dd902a02eee3.

index.html
msporny: Fixed in e37e305ea346a04d67b6a5fe5805ed54b0c7ccb2.

index.html
msporny: Fixed in abad5e5155e16bd6bd29c5babc0ca74078ec303b and 61f35f25a0bf25c087b516005ea41b01d933e759.

I expect that more text might be useful, but would rather do that in a future PR.

msporny:
> Between steps 8 and 9, perhaps find a way to indicate that the next part of the flow is the Go To Website protocol (rather than VC-API, OID4, NFC, Bluetooth, etc.)

Fixed in abad5e5155e16bd6bd29c5babc0ca74078ec303b.

dlongley (approved):

index.html
```diff
@@ -1355,6 +1367,237 @@ <h4>Exchange Examples</h4>
...
+      </p>
+      <p class="note" title="Interactions are application, use case, and protocol agnostic">
+While several interaction specifications reside in this document, the general
+approach is agnostic as to use case, application, and protocol. This approach
+can be used to pair two or more applications that desire to bootstrap into a
+particular protocol over any transmission medium &mdash; such as a web browser,
+QR Code (optical medium), or NFC (wireless medium) &mdash; where the protocol
+does not need to involve the VC API.
+      </p>
+      <p>
```
dlongley, line 1387:
```suggestion
      <p>
The sequence diagram below outlines an [=issuer=] generating an interaction
URL, using a QR Code to share it with a [=holder=], and proceeding with the
`vcapi` protocol:
      </p>
      <figure>
        <pre class="mermaid">
sequenceDiagram
  participant H as Holder
  participant B as Browser
  participant W as Holder Coordinator (Wallet)
  participant I as Issuer Coordinator
  autonumber
  H->>B: Click "Receive credential"
  B->>I: Request interaction QR Code
  I->>I: Generate QR Code
  I->>B: Display QR Code
  Note left of I: https://issuer.example/interactions/123?iuv=1
  H->>W: Start scanning QR Code
  W->>B: Scan QR Code
  W->>I: Get available interaction protocols
  Note left of V: GET https://issuer.example/interactions/123?iuv=1
  I->>W: Return interaction protocols
  Note left of I: (VC API, OID4, NFC, Bluetooth, etc.)
  W->>W: Select appropriate protocol
  opt "vcapi" protocol example
  W->>I: Initiate workflow exchange
  W->>H: Acquire consent to proceed
  Note right of H: issuer.example is offering credentials, accept?  
  end
        </pre>
        <figcaption>
An issuer-initiated interaction using a QR Code
        </figcaption>
      </figure>

      <p>
```

msporny:
I believe I have applied everyone's requested changes, including Dave's ones from the meeting before last. Having seen no objections to this PR as a first draft, merging.

Issue #302, opened 2022-08-31
Clarification of `GET /credentials` endpoint


kezike:
Is this endpoint supposed to return all credentials from an issuer associated with a holder in an inbox-like fashion? If so, I have some follow-up questions:
- Can issuers enable credential search via query parameters or some other mechanism? One idea that comes to mind is to accept a payload at this endpoint that includes a field to specify conditions for credential fields via JSONPaths? Examples of this form of filtering are littered throughout [the Presentation Exchange spec](https://identity.foundation/presentation-exchange). The only issue is that such a design would require a request body, which would necessitate an HTTP method other than `GET`, unless we somehow enable a search/filter mechanism via an intermediate exchange step.
- Is the expectation to use the OAuth 2.0 token (or some other auth mechanism) to identify the holder or should this request be the terminating end of an exchange workflow that begins with a DIDAuth flow?
- Should issuers enable paging via query parameters or the like?

kezike:
I believe I have gained some understanding of this endpoint at least as it pertains to the second point. The way that exchanges are currently defined, it seems like this endpoint would not necessarily need to be called as part of an exchange workflow with DID Auth, but rather that it would fetch credentials that have already been issued to a DID. As such, wallets would not need to perform DID Auth, since that would have already been done in the past and they will only pass verification during presentation if they control the subject DID. As such, OAuth seems to be the sole authentication responsibility for this endpoint (as for the others). Are these assumptions accurate?

ottonomy:
+1 to defining a pagination pattern for any list endpoint.

msporny:
The group discussed this on the 2023-06-20 telecon:

@PatStLouis noted that this could be used to get a list of credentials or a list of all issued credentials. It is not clear how this endpoint should be used. There was a discussion about what the endpoint should return when it is called. If you are an issuer coordinator, you might call this endpoint (on the issuer service) to get all credentials that you issued. If you are a holder coordinator, you might call this endpoint (on the issuer coordinator) to get all credentials that were issued to you. If you are a holder coordinator, you might call this endpoint (on the holder service) to get all of the credentials that you have stored in your digital wallet. 

For example, if you are a holder and you get a presentation request, you would want to call this endpoint (and filter based on the presentation request) to get the VCs that could be used to respond to the presentation request.

> Should issuers enable paging via query parameters or the like?

[...]

PR #477, opened 2025-04-30
Remove GET /credentials endpoint

bparth24:
Issuing Section -
Remove the Get /credentials endpoint from the specification

Related Issue https://github.com/w3c-ccg/vc-api/issues/302

**Note:** Close duplicate PR https://github.com/w3c-ccg/vc-api/pull/476 

Reviewers: @msporny

### DIFF
--- a/index.html
+++ b/index.html
@@ -867,7 +867,7 @@ The following APIs are defined for issuing a Verifiable Credential:
       </p>
 
       <table class="simple api-summary-table"
-        data-api-path="/credentials /credentials/{id}
+        data-api-path="/credentials/{id}
           /credentials/issue /credentials/status"></table>
 
       <section>
@@ -883,15 +883,6 @@ The following APIs are defined for issuing a Verifiable Credential:
 
         <div class="api-detail"
           data-api-endpoint="post /credentials/issue"></div>
-      </section>
-
-      <section>
-        <h4>Get Credentials</h4>
-        <p>
-        </p>
-
-        <div class="api-detail"
-          data-api-endpoint="get /credentials"></div>
       </section>
 
       <section>


dlongley (approved):

msporny (approved):

msporny:
Excellent, thank you for the PR @bparth24!

bparth24:
@msporny @dlongley 

I was rereading the conversation in issue #302, based on that, I've one question:

Should I also make the following changes to ensure consistency throughout the specs and make things clearer?

1. Remove the GET /credentials endpoint reference from section 3.7.1 (Issuer Coordinator endpoints)
2. Remove the GET /credentials endpoint reference from section 3.7.2 (Issuer Service endpoints)
3. Add a note in section 3.8 (Issuing) stating something like: 

> **Note:** The GET /credentials endpoint has been reserved but removed from this specification until it can be defined more properly. It was determined that:
> 
> This functionality might be better handled through workflow/exchange mechanisms rather than a direct endpoint on the issuer
> If retained in future versions, this endpoint should be limited to administrative functions
> Credential discovery and filtering for holders/verifiers should go through exchange mechanisms instead
> 
> This endpoint may be reintroduced in future versions of the specification when there is clearer implementer interest and consensus on its purpose and implementation details.

msporny:
> Remove the GET /credentials endpoint reference from section 3.7.1 (Issuer Coordinator endpoints)

Yes.

> Remove the GET /credentials endpoint reference from section 3.7.2 (Issuer Service endpoints)

Yes.

> Add a note in section 3.8 (Issuing) stating something like:

No, let's leave that out for now. While I appreciate that you are trying to explain to readers that the feature might come back, it will invite questions and discussion that we probably don't want to have now. We also want the spec to reflect the current reality, not some potential future reality that we're probably not going to achieve/do.

Thanks for thinking more deeply about this PR, @bparth24! Much appreciated. :)

bparth24:
@msporny cc: @dlongley 

Understood. Pushed a new commit [b35ef7c](https://github.com/w3c-ccg/vc-api/pull/477/commits/b35ef7cf405531d1c70d6f22c0bf1a4ff128ece1) with below changes:

> > Remove the GET /credentials endpoint reference from section 3.7.1 (Issuer Coordinator endpoints)
> 
> Yes.
> 
> > Remove the GET /credentials endpoint reference from section 3.7.2 (Issuer Service endpoints)
> 
> Yes.
> 

**Note:** I haven't updated the holder.yml file for the credentials endpoint from lines 71 to 110. Not sure if it is required.

msporny:
> Note: I haven't updated the holder.yml file for the credentials endpoint from lines 71 to 110. Not sure if it is required.

Yes, that needs to be removed as well. We don't want any of these endpoints in the .yml files either.

index.html
```diff
@@ -811,21 +811,13 @@ <h3>API Component Overview</h3>
...
         </p>
 
-      <h4>Issuer Coordinator</h4>
-      <p>
-        Below are all endpoints expected to be exposed by the Issuer Coordinator, along with the component
-        that is expected to call the endpoint
-      </p>
-      <table class="simple api-component-table"
-        data-api-path="/credentials"></table>
-
```
msporny, line 821: Let's not delete this section, we'll need it for the `/interactions/` endpoints. For now, we can just delete the `/credentials` value from the `data-api-path`.

bparth24: @msporny @dlongley 

Added the Issuer Coordinator section back in this commit [feb42f9](https://github.com/w3c-ccg/vc-api/pull/477/commits/feb42f91315869c8a18e03bab360bb6c8512f203)

msporny (approved):

Issue #406, opened 2024-07-19
Proposal to Recommend Coordinators to Expose the Same Endpoints as Services


laysakura:
Based on the description in Section 3.6, I created the following diagram: (Quoted from https://laysakura.github.io/2024/07/19/vc-api-v0.3/)

![image](https://github.com/user-attachments/assets/92494dfb-9591-4588-9423-769f89ca4a5c)

This diagram clearly illustrates the issue I'm pointing out.

---

https://github.com/w3c-ccg/vc-api/blob/066f0309d2c5e8b5a2bc972a0a60464c805e1705/index.html#L401-L403

> In the VC API, the Holder Coordinator initiates all flows. They request VCs from Issuers. They decide if, and when, to share those VCs with Verifiers.

However, there is almost no mention in this specification about the interfaces through which the Holder Coordinator communicates with Issuers and Verifiers.

While {Issuer, Verifier} Coordinators should certainly be allowed to have their own functions and interfaces, I propose that **it should be recommended that {Issuer, Verifier} Coordinators expose all endpoints of the {Issuer, Verifier} Services.**

This would allow implementers of Holder Coordinators to ensure interoperability across many {Issuer, Verifier} Coordinators.

dlongley:
The Workflows and Exchanges section of the spec is for crossing trust boundaries (e.g., Holders communicating with Issuers or Verifiers):

https://w3c-ccg.github.io/vc-api/#workflows-and-exchanges

It needs to be fleshed out more from the content we have developed in various issues but haven't created PRs for yet to the spec.

msporny:
The group discussed this on the 2024-07-23 telecon:

The discussion revolved around clarifying that workflows enable a caller to exchange credentials with a coordinator. 

A PR should be raised to highlight the notion that workflows are used to exchange credentials across trust boundaries with issuer coordinators or verifier coordinators.

PR #483, opened 2025-05-20
Add note that exchanges are not performed by coordinators.

msporny:
This PR is an attempt to address issue #406 by adding a note that exchanges are not performed by coordinators.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1099,7 +1099,12 @@ interaction with it is returned to the coordinator.
       </p>
       <p>
 The exchange URL is given to the exchange client so that it can initiate the
-exchange. Initiating the exchange does not require any authorization beyond the
+exchange. Note that while the exchange URL is given to the coordinator to
+provide to the exchange client, the actual exchange is performed between the
+exchange client and the workflow service, not the coordinator.
+      </p>
+      <p>
+Initiating the exchange does not require any authorization beyond the
 exchange URL. Depending on the workflow service implementation, exchange URLs
 can also be capability URLs (i.e., the URL is an unguessable secret such that
 only whomever is given the URL can initiate the exchange). If the workflow that


index.html
```diff
@@ -1099,7 +1099,12 @@ <h3>Workflows and Exchanges</h3>
       </p>
       <p>
 The exchange URL is given to the exchange client so that it can initiate the
-exchange. Initiating the exchange does not require any authorization beyond the
+exchange. Note that while the exchange URL is given to the coordinator to
+provide to the exchange client, the actual exchange is performed between the
+exchange client and the workflow service, not the coordinator.
```
dlongley, line 1104: Just a note here that the coordinator *could* be running the exchange client -- not sure if that needs to be called out to eliminate any confusion. In other words, there's nothing wrong with a coordinator using its own exchange client to perform the exchange.

dlongley: To give an example of when a coordinator would execute the exchange it creates:

A verifier coordinator may be a website that scans barcodes that include VCBs. It then needs to verify these, which it can do by creating an exchange using an administrator-established workflow and then executing (performing the exchange) using its own exchange client. It can then display the result to the end user on the website.

TallTed, line 1104:
```suggestion
exchange. Note that while the exchange URL is given to the coordinator for
provision to the exchange client, the actual exchange is performed between
the exchange client and the workflow service; the coordinator is not involved
after providing the exchange URL to the exchange client.
```

index.html
```diff
@@ -1099,10 +1099,10 @@ <h3>Workflows and Exchanges</h3>
...
       <p>
 The exchange URL is given to the exchange client so that it can initiate the
-exchange. Note that while the exchange URL is given to the coordinator for
-provision to the exchange client, the actual exchange is performed between
-the exchange client and the workflow service; the coordinator is not involved
-after providing the exchange URL to the exchange client.
+exchange. Note that while the exchange URL is given to the coordinator to then
+provide it to the exchange client, the actual exchange is performed between the
+exchange client and the workflow service; the coordinator is not involved after
+providing the exchange URL to the exchange client.
```
```suggestion
providing the exchange URL to the exchange client. For avoidance
of doubt, a coordinator can still use its own exchange client for any
use case that requires it to execute the exchange itself.
```

TallTed: Adjusting @dlongley's suggested addition....
```suggestion
providing the exchange URL to the exchange client. To be clear: a
coordinator can still use its own exchange client for any
use case that requires it to execute the exchange itself.
```

Issue #376, opened 2024-03-08
How to structure response metadata for `/credentials/verify` and `/presentations/verify` to include challenge verification metadata?


wes-smith:
How should the current response format for the above endpoints `{checks, warnings, errors}` be restructured to contain `challengeVerificationMetadata: {uses, verified, firstVerified}`?

msporny:
The group discussed this on the 2024-05-28 telecon:

@jandrieu noted that we post the current format in this conversation. The current format is specified here: https://github.com/w3c-ccg/vc-api/blob/main/components/VerifyCredentialResult.yml 

@dlongley noted that this part of the specification is very under specified. It's not clear how useful these structures are. @jandrieu thinks we need to restructure it, response should label itself... "These are the checks, these are the results of those checks". @dlongley agreed, "checks, warnings, and errors" are not bound to each other very well. Should be structured better -- whether an error is treated as a warning is a business rule -- provide all data to perform analysis. @PatStLouis noted that it's related to the ErrorDetails definition he's working on, but there is more work to be done here. @dlongley agreed that errors are protocol related (HTTP Errors), share as much as we can, includes "warnings" -- don't know how valuable that is (re-use old object for that), might need to figure out if all implementations need to behave the same, do we error on first error, or wait for all errors? In either case, lists all checks and any errors associated (might want every implementation to work that way) -- don't know level of prescriptive or not we want to be right now. @PatStLouis noted if you are verifying a VC, verify the proof (HTTP 422 -- unprocessable content, proof didn't verify?), do we want to match verification to existing status codes or is verifying a VC different enough that we would make our own error objects. @dlongley noted that it gets challenging when you verify more than one proof and one proof fails but the other does not. @PatStLouis wondered if we can start with one property and go from there. @dlongley suggested we figure out what the "minimum viable return value" would be. @PatStLouis would like to have just a "verified: true/false" if everything passed, false otherwise.

The group didn't think this was ready for a PR just yet and needed more work to figure out the details. We do know that the current setup of "checks/warnings/errors" is too ambiguous and needs to be fixed/changed.

msporny:
The group discussed this briefly on the 2024-06-18 call, but decided to wait for more people to be present to process it further.

msporny:
The group discussed this on the 2025-06-03 call and noted that the we addressed the issue for `/credentials/verify`, but have not done so for `/presentations/verify`. 

[...]

PR #488, opened 2025-06-21
Modernize verify presentation result.

msporny:
This PR is an attempt to address issue #376 by modernizing the result of presentation verification.

### DIFF
--- a/components/VerifyPresentationOptions.yml
+++ b/components/VerifyPresentationOptions.yml
@@ -15,17 +15,17 @@ components:
     VerifyPresentationOptions:
       type: object
       additionalProperties: false
-      description: Options for specifying how a presentation is verified
+      description: Options for specifying how a presentation is verified.
       properties:
         challenge:
           type: string
-          description: A challenge provided by the requesting party of the proof. For example 6e62f66e-67de-11eb-b490-ef3eeefa55f2
+          description: A challenge provided by the requesting party of the proof. For example, <code>6e62f66e-67de-11eb-b490-ef3eeefa55f2</code>.
         domain:
           type: string
-          description: The intended domain of validity for the proof. For example website.example
+          description: The intended domain of validity for the proof. For example, <code>website.example</code>.
         returnPresentation:
           type: boolean
-          description: Should the verified presentation be returned in the response? If true, then the verified presentation should be returned in the form in which it was verified. If false or not provided, then the verified presentation should not be returned.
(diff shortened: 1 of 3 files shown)


verifier.yml
```diff
@@ -129,11 +128,6 @@ components:
           $ref: "./components/EnvelopedVerifiablePresentation.yml#/components/schemas/EnvelopedVerifiablePresentation"
         options:
           $ref: "./components/VerifyPresentationOptions.yml#/components/schemas/VerifyPresentationOptions"
-    ProoflessVerifyPresentationRequest:
```
dlongley: It's important that we don't remove the ability to submit a VP w/o a proof that contains VCs that are to be verified. If we remove this, how do we signal that this is an acceptable use case?

dlongley (changes requested):

components/VerifyPresentationOptions.yml
```diff
@@ -15,17 +15,17 @@ components:
     VerifyPresentationOptions:
       type: object
       additionalProperties: false
-      description: Options for specifying how a presentation is verified
+      description: Options for specifying how a presentation is verified.
       properties:
         challenge:
           type: string
-          description: A challenge provided by the requesting party of the proof. For example 6e62f66e-67de-11eb-b490-ef3eeefa55f2
+          description: A challenge provided by the requesting party of the proof. For example, <code>6e62f66e-67de-11eb-b490-ef3eeefa55f2</code>.
```
TallTed:
```suggestion
          description: A challenge provided by the requesting party of the proof; for example, <code>6e62f66e-67de-11eb-b490-ef3eeefa55f2</code>.
```

components/VerifyPresentationOptions.yml
```diff
@@ -15,17 +15,17 @@ components:
...
         domain:
           type: string
-          description: The intended domain of validity for the proof. For example website.example
+          description: The intended domain of validity for the proof. For example, <code>website.example</code>.
```
TallTed:
```suggestion
          description: The intended domain of validity for the proof; for example, <code>website.example</code>.
```

TallTed (changes requested):
trivial

verifier.yml
msporny: Added presentations that don't have proofs in 90a00a2246d983182dfcc4f0142d12681a38e275.

msporny:
Addressed all comments, merging.

PR #489, opened 2025-06-21
Add /protocols endpoint to specific exchanges.

msporny:
This PR is an attempt to address issue #421 by adding the .../protocols endpoint to exchanges.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/pull/489.html" title="Last updated on Jul 26, 2025, 9:43 PM UTC (ef7fac9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/489/28eb39e...ef7fac9.html" title="Last updated on Jul 26, 2025, 9:43 PM UTC (ef7fac9)">Diff</a>

### DIFF
--- a/exchanges.yml
+++ b/exchanges.yml
@@ -346,7 +346,7 @@ components:
           properties:
             interact:
               type: string
-              description: A URL that can be used during human-in-the-loop exchange flows.
+              description: A URL that can be used during exchange flows with a human in the loop.
             vcapi:
               type: string
               description: The URL to use when initiating a VC API exchange.
--- a/exchanges.yml
+++ b/exchanges.yml
@@ -346,7 +346,7 @@ components:
           properties:
             interact:
               type: string
-              description: A URL that can be used during exchange flows with a human in the loop.
+              description: A URL that can be used during exchange flows with a human in the loop. See the section on the <a href="#dfn-interaction-url">interact URL format</a> for more details.
             vcapi:
               type: string
               description: The URL to use when initiating a VC API exchange.
--- a/exchanges.yml
+++ b/exchanges.yml
@@ -180,6 +180,36 @@ paths:
           description: Not Authorized
         "500":
           description: Internal Error
+  /workflows/{localWorkflowId}/exchanges/{localExchangeId}/protocols:
+    get:
+      summary: Gets the supported protocols for interacting with a specific exchange.
+      tags:
+        - Exchanges
+      security:
+        - networkAuth: []
+        - oAuth2: []
+        - zCap: []
+      operationId: getSupportedProtocolsConfiguration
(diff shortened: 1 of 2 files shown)


exchanges.yml
```diff
@@ -305,6 +335,27 @@ components:
...
+      additionalProperties: false
+      description: Object containing information about the protocols that can be used to perform a specific exchange.
+      properties:
+        protocols:
+          type: object
+          description: An object containing one or more protocols that can be used to perform the specific exchange.
+          properties:
+            interact:
+              type: string
+              description: A URL that can be used during human-in-the-loop exchange flows.
```
msporny: This needs a better description, @dlongley. I'm at a loss as to what to write.

index.html
```diff
@@ -1298,6 +1299,15 @@ <h4>Create Exchange</h4>
           data-api-endpoint="post /workflows/{localWorkflowId}/exchanges"></div>
       </section>
 
+      <section>
+        <h4>Get Exchange Protocols</h4>
+        <p>
+        </p>
+
```
msporny, line 1310: Maybe we need to explain how `interact` is supposed to work in this section?

exchanges.yml
dlongley: Now that we've defined the interact URL, we should just point to that section in the spec and / or reuse language from it. The value for `interact` is an interaction URL: https://w3c-ccg.github.io/vc-api/#interaction-url-format

index.html
dlongley: Or link to the interaction section? This section should say that the returned object lists all of the protocols supported by the exchange -- and that an implementation may implement an interaction URL handler by proxying/internally fetching this endpoint and returning its result to the caller.

We could use a mermaid style flowchart here to show an interaction URL being retrieved from a coordinator that internally looks up the associated exchange and fetches `<exchangeId>/protocols` and returns it to the caller. We can then explain that this allows the coordinator to delegate to an exchange service whereby the caller (client, e.g., digital wallet) can still use their trust in the coordinator origin -- as it serves (passes through) the `protocols` object.

exchanges.yml
TallTed: Not arguing against @dlongley's suggestion, but this would be better than what's here.
```suggestion
              description: A URL that can be used during exchange flows with a human in the loop.
```

msporny: Fixed via merging Ted's suggestion and this commit: 1332ea238f91eee17d9713d5cb209704aa3c9419

index.html
msporny: Fixed in ef7fac95c54552b531b89d39abe03da50f29ea9a.

msporny:
Addressed all comments, merging.

Issue #403, opened 2024-07-18
Add `presentationSchema` option to workflow step


dlongley:
A workflow step should be able to include a `presentationSchema` that will be run against a submitted presentation (if one is requested for that step).

The `presentationSchema` option should look like this:

```js
"presentationSchema": {
  "type": "JsonSchema",
  "jsonSchema": /* JSON schema here */
}
```

Other schema types can be supported with different custom properties depending on the "type" used.

msporny:
The group discussed this on 2024-07-23:

The issue was fairly straightforward and a PR should be raised for this item.

The PR should allow a `presentationSchema` option to be specified for a workflow step.

msporny:
PR #490 was merged to address this issue; closing.

PR #490, opened 2025-06-21
Add ability to specify presentationSchema to workflow step.

msporny:
This PR is an attempt to address issue #403 by adding a presentationSchema option to a workflow step.

### DIFF
--- a/exchanges.yml
+++ b/exchanges.yml
@@ -424,15 +424,22 @@ components:
         verifiablePresentationRequest:
           $ref: "./components/VerifiablePresentationRequest.yml#/components/schemas/VerifiablePresentationRequest"
         presentationSchema:
-          type: object
-          description: The presentation schema to use when validating a presentation in the current step.
-          properties:
-            types:
-              type: string
-              description: The type of presentation schema, such as <code>JsonSchema</code>.
-            jsonSchema:
-              type: object
-              description: The JSON Schema object to use if the <code>type</code> is <code>JsonSchema</code>.
+          oneOf:
+            - type: object
+              description: A JSON Schema describing the validation to be performed on the presentation.
+              properties:
+                  type:
+                    type: string
+                    description: The value MUST be <code>JsonSchema</code>.
+                  jsonSchema:
+                    type: object
+                    description: The JSON Schema object to use when validating the presentation.
+            - type: object
+              description: An alternate presentation schema format to use when validating a presentation in the current step. Additional properties are expected for a specific presentation schema type, but are beyond the scope of this specification.
+              properties:
+                types:
+                  type: string
+                  description: The type of presentation schema mechanism to use when validating the presentation.
         nextStep:
           type: string
           description: The name of the next step in the sequence. Passing nextStep is OPTIONAL. This field MUST NOT be present on the final step configuration.
--- a/exchanges.yml
+++ b/exchanges.yml
@@ -440,6 +440,23 @@ components:
                 types:
                   type: string
                   description: The type of presentation schema mechanism to use when validating the presentation.
+        verifyPresentationResponseSchema:
+          oneOf:
+            - type: object
+              description: A JSON Schema describing the validation to be performed on the result of verifying the presentation.
+              properties:
+                  type:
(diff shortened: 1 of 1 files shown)


exchanges.yml
```diff
@@ -358,6 +358,16 @@ components:
...
+        presentationSchema:
+          type: object
+          description: The presentation schema to use when validating a presentation in the current step.
+          properties:
+            types:
+              type: string
+              description: The type of presentation schema, such as <code>JsonSchema</code>.
+            jsonSchema:
+              type: object
+              description: The JSON Schema object to use if the <code>type</code> is <code>JsonSchema</code>.
```
dlongley: While we're here, we might want to also add the other optional counterpart to this schema, which is a schema to run against the result of verifying the presentation, `verifyPresentationResponseSchema`. This schema is also optional and can be used on the *result* from a workflow step that uses a VC API verifier to "verify a presentation". It has the same structure and is just a different schema that executes at a different time on a different value (on the JSON response from https://w3c-ccg.github.io/vc-api/#verify-presentation).

Related: #488 (as it better defines the result that this other optional schema would be run against).

dlongley (approved):

exchanges.yml
TallTed: What if it's not a JSON Schema? Clearly, we can't list all possible types that might ever exist, but it seems we should provide a clear way to handle an type that's not listed here (e.g., TypeSchema, JSON Constraint Rules (JCR)), including how to derive a property name (e.g., `typeScript`, `jsonConstraintRules`) -- and/or change `jsonSchema` to a more generic name, applicable to multiple (all?) possible types.

exchanges.yml
```diff
@@ -358,6 +358,16 @@ components:
...
           $ref: "./components/VerifiablePresentationRequest.yml#/components/schemas/VerifiablePresentationRequest"
+        presentationSchema:
+          type: object
+          description: The presentation schema to use when validating a presentation in the current step.
+          properties:
+            types:
+              type: string
+              description: The type of presentation schema, such as <code>JsonSchema</code>.
+            jsonSchema:
+              type: object
```
msporny:
```suggestion
            schema:
              type: string
```

This was discussed on the 2025-06-24 telecon as an alternative.

dlongley: I think the idea was that `type` determined what other properties would appear -- and for the `JsonSchema` type, the property `jsonSchema` is used to express the schema. This mirrors what is done in the VC-JSON-SCHEMA spec, see for example (the credential subject is a JSON schema):

https://www.w3.org/TR/vc-json-schema/#example-example-email-credential-schema

TallTed: I don't see anything in the text that communicates what you describe ... which *might* be OK, but I remain concerned that only providing detail about JSON Schema suggests that this is the *only* schema language that might be in play ... but I'm pretty sure that is not the intent, so I want to see at least one other example, which could be a dummy type.

msporny:
I'm currently blocked here... don't know what the group wants. Here seem to be the options:

* Define JUST the usage of JSON Schema for now.
* Define a more generic interface that allows the use of any schema language.

I think we can do the first bullet now (to be concrete) without blocking the second bullet from happening in the future. Just need guidance from the group on what to do.

msporny:
Add rules to the JSON Schema so that they match concretely on the following:

```
type: JsonSchema
jsonSchema: string
```

but also allow an "anyOf" match that does the following:

```
type: string
```

Explain that other schema languages are allowed. If you use JSON Schema, it matches the first item.

exchanges.yml
msporny: Fixed in 933a6ee42848132c10b5200cf6d86ee1a79a77a8.

msporny: Fixed in 677ec30771faccbe662e2ea9e0df13c38bb37fd7.

exchanges.yml
msporny: Fixed in 677ec30771faccbe662e2ea9e0df13c38bb37fd7.

TallTed (approved):
works for me

Issue #407, opened 2024-07-21
Details of `DELETE /credentials/{id}` endpoint should be added


laysakura:
https://w3c-ccg.github.io/vc-api/#issuer-service includes `DELETE /credentials/{id}` but any detailed API spec is not written.

msporny:
The group discussed this on the 2024-07-23 telecon:

The specification clearly does not document the `DELETE /credentials/{id}` endpoint and needs to do so.

A PR should be raised to document the `DELETE /credentials/{id}` endpoint.

msporny:
PR #491 has been merged to address this issue, closing.

PR #491, opened 2025-06-22
Add credential deletion section.

msporny:
This PR is an attempt to address issue #407 by adding a credential deletion section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/pull/491.html" title="Last updated on Jul 27, 2025, 9:23 PM UTC (7dcfcd3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/491/dbf8a82...7dcfcd3.html" title="Last updated on Jul 27, 2025, 9:23 PM UTC (7dcfcd3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -493,8 +493,8 @@ The <dfn>issuer service</dfn> takes requests to issue VCs from authorized Issuer
 Coordinators and returns well-formed, signed VCs. This
 service MUST have access to private keys (or key services which utilize private
 keys) in order to create the proofs for those VCs. The API between the Issuer
-service and its associated key service is believed to be out of scope for the VC
-API, but may be addressed by WebKMS or similar specifications.
+service and its associated cryptographic key management service is
+out of scope for the VC API.
         </p>
         <p>
 The <dfn>verifier service</dfn> takes requests to verify Verifiable Credentials
--- a/index.html
+++ b/index.html
@@ -490,7 +490,7 @@ network endpoints of services are defined herein.
         </p>
         <p>
 The <dfn>issuer service</dfn> takes requests to issue VCs from authorized Issuer
-Coordinators and returns well-formed, signed Verifiable Credentials. This
+Coordinators and returns well-formed, signed VCs. This
 service MUST have access to private keys (or key services which utilize private
 keys) in order to create the proofs for those VCs. The API between the Issuer
 service and its associated key service is believed to be out of scope for the VC
@@ -500,21 +500,21 @@ API, but may be addressed by WebKMS or similar specifications.
 The <dfn>verifier service</dfn> takes requests to verify Verifiable Credentials
 and Verifiable Presentations and returns the result of checking their proofs and
 status (if present). The service only checks the authenticity and timeliness of
-the VC; leaving the Verifier Coordinator to finish Applying any business rules
-needed.
+the VC, leaving the Verifier Coordinator to finish applying any necessary
+business rules.
         </p>
         <p>
 The <dfn>holder service</dfn> takes requests to create Verifiable Presentations
 from an optional set of VCs and returns well-formed, signed Verifiable
 Presentations containing those VCs. These VPs are used with Issuers to
-demonstrate control over DIDs prior to issuance and with Verifiers to present
-specific VCs.
+demonstrate control over DIDs prior to VC issuance and with Verifiers to
+present specific VCs.
         </p>
       </section>
       <section>
         <h3>Status Service</h3>
         <p>
-The <dfn>status service</dfn> provides a privacy-preserving means for publishing
+The <dfn>status service</dfn> provides a privacy-preserving means of publishing
 and checking the status of any Verifiable Credentials issued by the Issuer.
 Implementers of verifier services are encouraged to understand the privacy
 implications of checking status by referring to the respective status
@@ -522,14 +522,14 @@ specification used by the verifiable credential.
         </p>
         <p>
 For specific mechanisms by which to manage Verifiable Credential statuses, it's
-recommended to refer to external well known specifications, such as the
+recommended to refer to well-known external specifications, such as the
 [[VC-BITSTRING-STATUS-LIST]].
         </p>
       </section>
       <section>
         <h3>Storage Services</h3>
         <p>
-<strong>Storage Service (Issuer) &bull;Storage Service (Verifier) &bull; Storage
+<strong>Storage Service (Issuer) &bull; Storage Service (Verifier) &bull; Storage
 Service (Holder)
 </strong>
         </p>
@@ -560,11 +560,11 @@ back-end storage platforms.
         <p>
 The <dfn>workflow service</dfn> provides a way for coordinators to automate
 specific interactions for specific users. Each role (Holder, Issuer, and
-Verifier) can run their own Workflow Service to create and manage exchanges that
+Verifier) can run their own workflow service to create and manage exchanges that
 realize particular workflows. Administrators configure the workflow system to
 support particular flows. Then, when the business rules justify it, coordinators
-create exchanges at their Workflow Service and give authorized access to those
-exchanges to any party.
+create exchanges at their workflow service and give access to those exchanges
+to any authorized party.
         </p>
       </section>
       <section>
@@ -934,9 +934,9 @@ error if `credential` values that contain existing `proof` values are provided.
         <p>
 An [=issuer service=] or a [=holder service=] might store an issued
 [=verifiable credential=] for an extended period of time. When this is done,
-it can be useful to delete such a [=verifiable credential=]. An [=issuer=]
-might need to do so because of regulatory requirements such as
-<a href="https://en.wikipedia.org/wiki/Right_to_be_forgotten">
+it can be useful to delete such a [=verifiable credential=]; for instance,
+an [=issuer=] might need to do so because of regulatory requirements such
+as <a href="https://en.wikipedia.org/wiki/Right_to_be_forgotten">
 the right to be forgotten</a>. See Section [[[#deletion]]] for additional
 considerations related to the removal of [=verifiable credentials=] from
 systems.
--- a/index.html
+++ b/index.html
@@ -489,48 +489,49 @@ interfaces, such interfaces are out of scope for the VC API. Only the
 network endpoints of services are defined herein.
         </p>
         <p>
-The Issuer Service takes requests to issue VCs from authorized Issuer Coordinators
-and returns well-formed, signed Verifiable Credentials. This service MUST
-have access to private keys (or key services which utilize private keys)
-in order to create the proofs for those VCs. The API between the Issuer
-service and its associated key service is believed to be out of scope for
-the VC API, but may be addressed by WebKMS or similar specifications.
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -489,48 +489,49 @@ <h3>Services</h3>
...
         </p>
         <p>
-The Verifier service takes requests to verify Verifiable Credentials and
-Verifiable Presentations and returns the result of checking their proofs
-and status (if present). The service only checks the authenticity and
-timeliness of the VC; leaving the Verifier Coordinator to finish Applying any business rules needed.
+The <dfn>verifier service</dfn> takes requests to verify Verifiable Credentials
+and Verifiable Presentations and returns the result of checking their proofs and
+status (if present). The service only checks the authenticity and timeliness of
+the VC; leaving the Verifier Coordinator to finish Applying any business rules
```
```suggestion
the VC; leaving the Verifier Coordinator to finish applying any business rules
```

index.html
```diff
@@ -489,48 +489,49 @@ <h3>Services</h3>
...
+Implementers of verifier services are encouraged to understand the privacy
+implications of checking status by referring to the respective status
+specification used by the verifiable credential.
         </p>
         <p>
-          For specific mechanisms by which to manage Verifiable Credential statuses,
-          it's recommended to refer to external well known specifications, such as the
-          [[VC-BITSTRING-STATUS-LIST]].
+For specific mechanisms by which to manage Verifiable Credential statuses, it's
+recommended to refer to external well known specifications, such as the
```
```suggestion
recommended to refer to external well-known specifications, such as the
```

index.html
```diff
@@ -928,6 +929,23 @@ <h4>Update Status</h4>
...
+        <h4>Delete a Specific Credential</h4>
+        <p>
+An [=issuer service=] or a [=holder service=] might store an issued
+[=verifiable credential=] for an extended period of time. When this is done,
+it can be useful to delete such a [=verifiable credential=]. An [=issuer=]
+might need to do so because of regulatory requirements such as
+<a href="https://en.wikipedia.org/wiki/Right_to_be_forgotten">
+the right to be forgotten</a>. See Section [[[#deletion]]] for additional
+considerations related to the removal of [=verifiable credentials=] from
+systems.
```
dlongley, line 942: +1 to this. We might want to consider some additional feature which would be more of a "prune" operation than a "delete", whereby the VC isn't deleted entirely, but its `credentialId` and any status entries information remain (information required for the application of uniqueness constraints and the maintenance of status lists, etc.).

dlongley (approved):

index.html
TallTed: The word order I suggest makes it (somewhat) less likely that `well-known specifications` will be read as a unit, thus not referencing only specifications that relate to the `.well-known` space.
```suggestion
recommended to refer to well-known external specifications, such as the
```

index.html
```diff
@@ -489,48 +489,49 @@ <h3>Services</h3>
...
         </p>
         <p>
-The Issuer Service takes requests to issue VCs from authorized Issuer Coordinators
-and returns well-formed, signed Verifiable Credentials. This service MUST
-have access to private keys (or key services which utilize private keys)
-in order to create the proofs for those VCs. The API between the Issuer
-service and its associated key service is believed to be out of scope for
-the VC API, but may be addressed by WebKMS or similar specifications.
+The <dfn>issuer service</dfn> takes requests to issue VCs from authorized Issuer
+Coordinators and returns well-formed, signed Verifiable Credentials. This
```
TallTed: It seems odd to have this `Verifiable Credentials` where the rest of the paragraph uses the `VC` abbreviation. I'm suggesting to abbreviate this one, to make the paragraph internally consistent, believing that there's a `Verifiable Credentials (VCs)` or similar earlier in the document. The whole document should optimally be made consistent, likely with `<dfn>` tags applied to the appropriate (usually first) instance and `<a>` tags applied to the rest, and using `VC` in most instances.
```suggestion
Coordinators and returns well-formed, signed VCs. This
```

index.html
```diff
@@ -489,48 +489,49 @@ <h3>Services</h3>
...
+service MUST have access to private keys (or key services which utilize private
+keys) in order to create the proofs for those VCs. The API between the Issuer
+service and its associated key service is believed to be out of scope for the VC
+API, but may be addressed by WebKMS or similar specifications.
```
TallTed: Is there an appropriate link for WebKMS, to serve with `<a >` or `[[? ]]`, as appropriate?

index.html
```diff
@@ -489,48 +489,49 @@ <h3>Services</h3>
...
         <p>
-The Verifier service takes requests to verify Verifiable Credentials and
-Verifiable Presentations and returns the result of checking their proofs
-and status (if present). The service only checks the authenticity and
-timeliness of the VC; leaving the Verifier Coordinator to finish Applying any business rules needed.
+The <dfn>verifier service</dfn> takes requests to verify Verifiable Credentials
+and Verifiable Presentations and returns the result of checking their proofs and
+status (if present). The service only checks the authenticity and timeliness of
+the VC; leaving the Verifier Coordinator to finish Applying any business rules
+needed.
```
TallTed: As used here, the semicolon should separate two complete sentences. The `leaving...` section could be reworded to start with something like `the Verifier Coordinator is left to finish` support this use, but going with the comma (as I've suggested below) is the smaller change to make the sentence correct.
```suggestion
the VC, leaving the Verifier Coordinator to finish applying any necessary
business rules.
```

index.html
```diff
@@ -489,48 +489,49 @@ <h3>Services</h3>
...
-  The Holder service takes requests to create Verifiable Presentations
-  from an optional set of VCs and returns well-formed, signed Verifiable
-  Presentations containing those VCs. These VPs are used with Issuers to
-  demonstrate control over DIDs prior to issuance and with Verifiers to
-  present specific VCs.
+The <dfn>holder service</dfn> takes requests to create Verifiable Presentations
+from an optional set of VCs and returns well-formed, signed Verifiable
+Presentations containing those VCs. These VPs are used with Issuers to
+demonstrate control over DIDs prior to issuance and with Verifiers to present
+specific VCs.
```
```suggestion
demonstrate control over DIDs prior to VC issuance and with Verifiers to
present specific VCs.
```

index.html
```diff
@@ -489,48 +489,49 @@ <h3>Services</h3>
...
       </section>
       <section>
         <h3>Status Service</h3>
         <p>
-        The Status Service provides a privacy-preserving means for publishing
-        and checking the status of any Verifiable Credentials issued by the
-        Issuer. Implementers of verifier services are encouraged to understand the
-        privacy implications of checking status by referring to the respective
-        status specification used by the verifiable credential.
+The <dfn>status service</dfn> provides a privacy-preserving means for publishing
```
```suggestion
The <dfn>status service</dfn> provides a privacy-preserving means of publishing
```

index.html
```diff
@@ -489,48 +489,49 @@ <h3>Services</h3>
...
+[[VC-BITSTRING-STATUS-LIST]].
         </p>
       </section>
       <section>
         <h3>Storage Services</h3>
         <p>
-          <strong>Storage Service (Issuer) &bull;Storage Service
-            (Verifier) &bull; Storage Service (Holder)
-          </strong>
+<strong>Storage Service (Issuer) &bull;Storage Service (Verifier) &bull; Storage
```
```suggestion
<strong>Storage Service (Issuer) &bull; Storage Service (Verifier) &bull; Storage
```

index.html
```diff
@@ -557,13 +558,13 @@ <h3>Storage Services</h3>
...
-The Workflow Service provides a way for coordinators to automate specific
-interactions for specific users. Each role (Holder, Issuer, and Verifier)
-can run their own Workflow Service to create and manage exchanges that
-realize particular workflows. Administrators configure the workflow system
-to support particular flows. Then, when the business rules justify it,
-coordinators create exchanges at their Workflow Service and give authorized
-access to those exchanges to any party.
+The <dfn>workflow service</dfn> provides a way for coordinators to automate
+specific interactions for specific users. Each role (Holder, Issuer, and
+Verifier) can run their own Workflow Service to create and manage exchanges that
```
```suggestion
Verifier) can run their own workflow service to create and manage exchanges that
```

index.html
```diff
@@ -557,13 +558,13 @@ <h3>Storage Services</h3>
...
+realize particular workflows. Administrators configure the workflow system to
+support particular flows. Then, when the business rules justify it, coordinators
+create exchanges at their Workflow Service and give authorized access to those
+exchanges to any party.
```
```suggestion
create exchanges at their workflow service and give access to those exchanges
to any authorized party.
```

index.html
```diff
@@ -928,6 +929,23 @@ <h4>Update Status</h4>
...
       </section>
 
+      <section>
+        <h4>Delete a Specific Credential</h4>
+        <p>
+An [=issuer service=] or a [=holder service=] might store an issued
+[=verifiable credential=] for an extended period of time. When this is done,
+it can be useful to delete such a [=verifiable credential=]. An [=issuer=]
+might need to do so because of regulatory requirements such as
+<a href="https://en.wikipedia.org/wiki/Right_to_be_forgotten">
```
```suggestion
it can be useful to delete such a [=verifiable credential=]; for instance,
an [=issuer=] might need to do so because of regulatory requirements such
as <a href="https://en.wikipedia.org/wiki/Right_to_be_forgotten">
```

TallTed (changes requested):
Editorial tweaks, mostly minor, a couple more broad.

index.html
msporny: @PatStLouis noted that deleting the VC would need to delete any "linked" information -- including possibly revoking the credential? 

The group discussed some of this on the 2025-06-24 call and came to the following conclusions:

* The difference between DELETE and "prune" is that the latter does not revoke the credential, and the former revokes the credential.
* Deleting any "higher level" linked information is the responsibility of the issuer coordinator and not the issuer service.
* Deleting might not be a good option if there are audit requirements (especially due to regulatory burdens).

dlongley: I don't know that an issuer instance would necessarily "know" to "revoke" (or change any other status bit for any other status "purpose") when deletion is performed.

I think an individual implementation could implement some stops/checks that would prevent deletion via a special configuration, but I don't think it would necessarily be easy to have some default behavior based solely on the "revocation" purpose. What about "suspension"? What about other custom purposes? It seems setting an expectation or special case for "revocation" would lead to more confusion with the DELETE function, not less.

I would say, instead, that we should keep the deletion operation at this level separate and simple. We should say that custom configurations can prevent deletion for any reason (including checking arbitrary status values or for other existing "linked" information) and provide an appropriate error that would signal that a client needs to run other operations using the VC API (e.g., flip status bits) in order to enable deletion. This approach would also avoid creating multiple ways / mechanisms for changing status.

PatStLouis: +1 to @dlongley response

index.html
msporny: There is, but probably not worth citing now because it's incomplete: https://w3c-ccg.github.io/webkms/

msporny: Fixed in 7dcfcd394d169f576c911b7b4c9b97db8f5a93eb.

msporny:
I think I've addressed all comments, merging.

Issue #486, opened 2025-05-20
Update yaml schemas displaying a Verifiable Credential to 2.0


PatStLouis:
This affects some of the properties such as the `issuanceDate` and `expirationDate`

PR #492, opened 2025-07-08
Update examples to use vcdm 2.0 context and properties

PatStLouis:
Addresses #486

Updates the context URLs, some properties and adds some definitions.


dlongley (approved):
The `proofValue` values need to be updated to at least start with the proper multibase character, otherwise this LGTM.

Issue #413, opened 2024-08-20
Create WorkflowConfig object to re-use across API calls


PatStLouis:
Following #409, we want to create a workflowConfig object to re-use across calls

PatStLouis:
Point of discussion, the `steps` property.

The model currently looks like this:
```
steps [object]
    One or more steps required to complete an exchange on the workflow. The steps object MUST be an object of the following form:

    stepName [object]
        The name of the step. The stepName object MUST be an object of the following form:

        step [object]
            Object containing information about a workflow step. The step object MUST be either an object of the following form: 
```

While I understand what it's meant to represent, the way it's modeled is confusing such as the description of the `stepName` being the name of the step while being an object.

msporny:
The group discussed this on the 2024-09-10 call:

@PatStLouis noted his comment above as a concern. @dlongley noted that "steps" is an object where they keys are the name of the steps. Lots of discussion, but end result was to modify the text and possibly the code to ensure that the output is easier to read.

kezike:
As I was reviewing the latest version of the [workflow creation](https://w3c-ccg.github.io/vc-api/#create-workflow) and [workflow retrieval](https://w3c-ccg.github.io/vc-api/#get-workflow-configuration) endpoints, I noticed that there is an intermediate `step` property. This seems unnecessary and may be error prone for implementers. Could we remove that?

dlongley:
@kezike,

Yes, the "intermediate `step` property" you mention is a bug somewhere in the yaml or in how it's processed. It will be removed. The "steps" object has keys that are the step names and values that are the step objects. There should be no intermediate `step` property at all.

BigBlueHat:
Closed via #494

PR #494, opened 2025-07-11
Remove intermediate step property from workflow steps

bparth24:
This PR addresses issue #413 - Remove intermediate step property from workflow steps

## 📋 Changes Made

- [x] Removed confusing intermediate `step` property from workflow schema (#413)
- [x] Simplified workflow steps structure to use direct step name keys
- [x] Updated `CreateWorkflowRequest` and `GetWorkflowResponse` schemas in `exchange.yml` file
- [x] Improved schema clarity by using `theNameOfTheStep` placeholder

@msporny @dlongley ready for review.

Additionally, I have encountered a rendering issue in the generated documentation. In the `createWorkflow` endpoint, all attributes after `initialStep` are displayed a bit weirdly in the HTML output. I have been unable to identify the root cause of this rendering problem.

**Could you please:**
1. Verify that my understanding and implementation of issue #413 is correct
2. Help identify what might be causing the documentation rendering issues with the attributes following `initialStep`

I want to ensure that both the schema fix and the documentation display are working as expected before this is merged.

### DIFF
--- a/exchanges.yml
+++ b/exchanges.yml
@@ -191,14 +191,10 @@ components:
       properties:
         steps:
           type: object
-          description: One or more steps required to complete an exchange on the workflow. Passing the steps object is REQUIRED.
+          description: One or more steps required to complete an exchange on the workflow. Passing the steps object is REQUIRED. The keys are step names and values are step configurations.
           properties:
-            stepName:
-              type: object
-              description: The name of the step.
-              properties:
-                step:
-                  $ref: "#/components/schemas/WorkflowStep"
+            theNameOfTheStep:
+              $ref: "#/components/schemas/WorkflowStep"
         initialStep:
           type: string
           description: The step from the above set that the exchange starts on. Passing intialStep is REQUIRED.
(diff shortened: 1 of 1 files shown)


exchanges.yml
```diff
@@ -191,14 +191,10 @@ components:
       properties:
         steps:
           type: object
-          description: One or more steps required to complete an exchange on the workflow. Passing the steps object is REQUIRED.
+          description: One or more steps required to complete an exchange on the workflow. Passing the steps object is REQUIRED. The keys are step names and values are step configurations.
```
TallTed:
```suggestion
          description: One or more steps required to complete an exchange on the workflow. Passing the steps object is REQUIRED. The keys are step names and the values are step configurations.
```

exchanges.yml
```diff
@@ -228,14 +224,10 @@ components:
       properties:
         steps:
           type: object
-          description: One or more steps required to complete an exchange on the workflow.
+          description: One or more steps required to complete an exchange on the workflow. The keys are step names and values are step configurations.
```
TallTed:
```suggestion
          description: One or more steps required to complete an exchange on the workflow. The keys are step names and the values are step configurations.
```

TallTed (changes requested):
Trivial. Add a couple of missing words.

bparth24:
> Trivial. Add a couple of missing words.

Thank you, @TallTed, for catching those.

ottonomy:
Should probably use `additionalProperties` like:
```
properties:
  steps:
    type: object
    description: One or more steps required to complete an exchange on the workflow. The keys are step names and the values are step configurations.
    additionalProperties:
      $ref: "#/components/schemas/WorkflowStep"
      ```

dlongley (approved):
Ideally there would be a fix to do this as @ottonomy suggests, but short of that, this is an improvement.

TallTed (approved):

Issue #422, opened 2024-09-24
Add support for proof chains to VC API POST credential/issue endpoint


aljones15:
A credential issuance endpoint can be posted to multiple times with initially an unsigned credential and then later a signed credential which will have a new proof added to the VC which could use the `proofId` and `previousProof` options.
https://w3c-ccg.github.io/vc-api/#issue-credential

The schema for POST `credentials/issue` should have an optional `proof` property

the options object should now have optional options `previousProof` String or Array<String> or `proofId` for specifying a proofId for a proof.

dlongley:
+1 to most of this proposal.

> the options object should now have optional options previousProof String or Array or proofId for specifying a proofId for a proof.

This is the only part that give me heartburn, since the issuer may be attaching more than one proof or there may be more than one previous proof, and these options might be insufficient / mismatched with the requirements. It might be better for `proofId` to simply be auto-generated for each proof and that every previous proof that the client submits in the credential be included in `previousProof`. We should figure out if we need more complexity than this or not in what we standardize (with, of course, individual implementations continuing to be able to add custom options as it makes sense).

PatStLouis:
question: should the issuer verify exiting proofs before adding a proof set/chain? It's not in the spec but this would make an issuer instance output invalid credentials...

dlongley:
@PatStLouis,

> should the issuer verify exiting proofs before adding a proof set/chain?

No, I don't think it should do this -- and I think it's on the client if it uses an issuer endpoint to produce something invalid. There are many ways "invalid" (in some sense) output could be generated due to configuration issues and other problems and, IMO, we can't and shouldn't try to patch them all by spreading all the various responsibilities around to each component/endpoint.

TallTed:
> should the issuer verify exiting proofs before adding a proof set/chain?

I think this will likely depend on the deployment. Some will want to require it, others won't find it important. More business logic.

msporny:
The group discussed this on the 2025-01-14 telecon:

[...]

PR #499, opened 2025-07-22
Update Proof description & Add proofId, previousProof properties

bparth24:
This PR also addresses **#422** by clarifying how the `/credentials/issue` endpoint handles credentials with existing proofs as agreed upon in the 2025-05-27 call.

**Important Note:** Created a new PR for additional changes. This PR is related to PR https://github.com/w3c-ccg/vc-api/pull/495#issuecomment-3104499743

## 📋 Files & Changes Made

1. Update `components/Credential.yml` file - enhance `UnsecuredCredential` proof property description to `An optional proof or array of proofs for credentials that are secured using proof sets or chains. When present, the issuer instance configuration determines how these existing proofs are processed (appended to create proof sets/chains, or trigger an error).`
2. Update `components/IssueCredentialOptions.yml` - add optional `proofId` and `previousProof` properties  

**Section 3.8.1 Before Updates:**

Screenshot 2025-07-22 at 3 17 18 PM

**Section 3.8.1 After Updates**

Screenshot 2025-07-22 at 3 11 00 PM

________

@dlongley cc: @msporny @BigBlueHat @TallTed @PatStLouis  -- Ready for review.

### DIFF
--- a/components/Credential.yml
+++ b/components/Credential.yml
@@ -91,8 +91,11 @@ components:
           description: The subject
         "proof":
           type: object
-          description: An optional proof for credentials that are secured using proof sets or chains.
-          oneOf: 
+          description: >
+            An optional proof or array of proofs for credentials that are secured using proof sets or chains. 
+            When present, the issuer instance configuration determines how these existing proofs are processed 
+            (appended to create proof sets/chains, or trigger an error).
+          oneOf:
             - type: object
             - type: array
               items: 
--- a/components/IssueCredentialOptions.yml
+++ b/components/IssueCredentialOptions.yml
@@ -38,6 +38,17 @@ components:
             auto-generate a credentialId if one is not supplied by the issuer
             coordinator, because doing so could create a partitioning error if the
             result is never received by the client.
+        proofId:
+          type: string
+          description: >
+            An optional identifier for the proof being added by this issuer instance. 
+            If not provided, the issuer service MAY auto-generate a proof identifier 
+            based on its configuration.
+        previousProof:
+          type: object
+          description: >
+            Optional reference to previous proof(s) when creating proof chains. 
+            The issuer instance configuration determines how this is processed.
       example:
         {
           "credentialId": "example.com/ad5d541f-db7a-4bff-97e1-d403ce403767",


components/Credential.yml
```diff
@@ -91,8 +91,11 @@ components:
           description: The subject
         "proof":
           type: object
-          description: An optional proof for credentials that are secured using proof sets or chains.
-          oneOf: 
+          description: >
+            An optional proof or array of proofs for credentials that are secured using proof sets or chains. 
+            When present, the issuer instance configuration determines how these existing proofs are processed 
+            (appended to create proof sets/chains, or trigger an error).
```
TallTed:
```suggestion
            When present, the configuration of the issuer instance determines how these existing proofs are
            processed (e.g., append to create proof sets/chains, or trigger an error).
```

components/IssueCredentialOptions.yml
```diff
@@ -38,6 +38,17 @@ components:
...
+          type: string
+          description: >
+            An optional identifier for the proof being added by this issuer instance. 
+            If not provided, the issuer service MAY auto-generate a proof identifier 
+            based on its configuration.
+        previousProof:
+          type: object
+          description: >
+            Optional reference to previous proof(s) when creating proof chains. 
+            The issuer instance configuration determines how this is processed.
```
TallTed:
```suggestion
            The configuration of the issuer instance determines how this is processed.
```

components/IssueCredentialOptions.yml
```diff
@@ -38,6 +38,17 @@ components:
...
+          type: string
+          description: >
+            An optional identifier for the proof being added by this issuer instance. 
+            If not provided, the issuer service MAY auto-generate a proof identifier 
+            based on its configuration.
+        previousProof:
+          type: object
+          description: >
+            Optional reference to previous proof(s) when creating proof chains. 
+            The configuration of the issuer instance determines how this is processed.
```
msporny: Hmm, interested @dlongley's take on this as an implementer.

This should probably be a `string` that refers to a previous `proofId`. Specifying `proofId` feels useful too, but it should probably be a globally unique identifier with a scheme prefix, such as `urn:uuid:d6764dfc-6b33-11f0-8e93-63e78a307c56` or `https://domain.example/proofs#82978473729973`.

components/IssueCredentialOptions.yml
```diff
@@ -38,6 +38,17 @@ components:
             auto-generate a credentialId if one is not supplied by the issuer
             coordinator, because doing so could create a partitioning error if the
             result is never received by the client.
+        proofId:
+          type: string
```
msporny: This should probably be a globally unique identifier with a scheme prefix, such as `urn:uuid:d6764dfc-6b33-11f0-8e93-63e78a307c56` or `https://domain.example/proofs#82978473729973`.

dlongley: TL; DR: Hmm, I don't think we should specify this at all, actually. I don't think we're looking for interop here yet and we don't have good enough incubation on whether specifying this in this way will match common use.

Many of my following comments favor avoiding specifying any settings/features as top-level options like this when they clearly can only target one specific proof.

---

Just looking at this now, I wouldn't expect `proofId` to be something that one specifies via these options, but that is instead dynamically generated. There would also be an immediate problem for issuer instances that need to generate more than one proof. So, I think this is better done as a proof-scoped issuer instance configuration setting. You would indicate, in the config, that you want the issuer instance, when it generates a particular proof (for which there might be several), to include in it a `proofId`. More details than just that are generally needed too -- and I would expect all of this to need scoping.

For example, some implementations today have a section in the issuer instance config where the settings for multiple proofs can be individually specified. This includes an entry that is scoped to each proof to be generated, including settings for the cryptosuite to be used and any other related values like references to keys / zcaps, other flags.

Since each entry in this section is scoped to a proof that is to be created -- just sending a top-level `proofId` in as an option from the client would probably create an impedance mismatch for anything but the simplest case. Instead, these entries could be augmented to allow: indicating whether a proof chain ought to be created or not, some optional "selector" for identifying any previous proof, something to determine proof ordering for all new / to be created proofs, and perhaps a `proofId` template of sorts.

Notably, this is related to the `previousProof` option below too. I don't think should be specified yet either -- and wouldn't work well the way it presently is in this PR for similar reasons. Just thinking about how one would use such an option: they would would have to write some other software that would detect / select the correct previous proof before making their issue call, shifting more responsibility onto a party that might be previously designed to focus on the data that goes into the VC, not its cryptography layer. In many scenarios, efforts have been made to allow this party to rely on abstracted away, predetermined, and consistent rules for the crypto parts (i.e., separation of concerns).

Furthermore, I imagine that, over time, a "selector" primitive would emerge to allow for code reuse when different `previousProof` selections need to be made, but all of the "parsing and finding" code is the same. But, even then, the option (as specified below) would only work for the case that there is exactly one `previousProof` of interest. And -- this would be for just one subset of your new proofs where your issuer instance config would have to know where to plug in that single value.

When a developer needs to handle more than this highly specific case they could: create more complicated top-level "options" that allow scoping `previousProof` values -- and enable those to match up with their existing issuer instance config -- or -- more simply, add the ability to specify a proof-scoped "selector" in the existing issuer instance config. The latter would have the benefit of reclaiming the aforementioned responsibility / separation of concerns.

That's how I see this progressing -- for a more efficient, reusable design anyway. Other devs might be pressured to opt for a "quick" (short term) fix where everything is rewritten for the specific use case -- possibly leading to vendor lock-in and tech debt. We don't want to help that situation occur by specifying something sub-optimal because we didn't have the requisite implementation experience.

So, I would expect we want more feedback from implementers that have use cases with a need to extend proof chains. They need to experiment more here to arrive at something we know is sufficiently reusable. Maybe that will result in some kind of scoped-"selector" that can go in an issuer instance config -- or some other way forward. I don't think we know what that is just yet, but I do think it won't be what is presently specified in this PR.

Let's wait on specifying anything until we have implementers saying they are ready to test for interop on a flexible, reusable approach.

components/IssueCredentialOptions.yml
dlongley: See above: https://github.com/w3c-ccg/vc-api/pull/499/files#r2240519905

bparth24:
@dlongley @msporny 

VC API Call 29-07-2025 group decided to avoiding specifying settings/features such as `proofId` and `previousProof` in top-level options.

@bparth24 to remove those attributes and push the change.

bparth24:
@dlongley cc: @msporny @TallTed 

Removed the `proofId` and `previousProof` properties. Ready for review and merge once review looks okay.

dlongley (approved):

TallTed (approved):

msporny (approved):

msporny:
@bparth24 This PR needs a rebase to resolve merge conflicts.

Issue #456, opened 2025-02-19
Potentially improper format for workflow `createAuthorizationRequest` property


kezike:
I noticed that the `createAuthorizationRequest` property in `workflow.steps[STEP_NAME].openId.createAuthorizationRequest` found in the [workflow creation](https://w3c-ccg.github.io/vc-api/#create-workflow) request and response as well as the [workflow retrieval](https://w3c-ccg.github.io/vc-api/#get-workflow-configuration) response is specified as a string. I believe that this should be a boolean. Is that correct?

dlongley:
No, it actually is correct as a string. The value identifies the name of the variable that the auto-generated authorization request will be stored in for subsequent use in the exchange. The text should be updated to make this clear.

kezike:
Ah OK, where the "auto-generated authorization request" is an OID4* authorization request?

Also, I think that we should update both the description of this property as well as the name of the property itself. Since we have another step-level `create*` property that is a boolean (`createChallenge`), it seems inconsistent to use that verb for this property, which is not a boolean.

dlongley:
> Ah OK, where the "auto-generated authorization request" is an OID4* authorization request?

Yeah, that's right.

The `createAuthorizationRequest` key is nested under `openId` and is for auto-generating an "OpenID for Verifiable Presentations Authorization Request" from the combination of a VPR (for example, see: https://github.com/digitalbazaar/oid4-client/blob/v4.3.0/lib/oid4vp.js#L300) and other exchange information. Optionally, additional, specific OID4VP parameters can be included in the step's `openId` section so they will be used in the generated authorization request.

We could flush out how to integrate VC API exchanges with OID4VP a bit more in the spec so how to generate a compliant authorization request is more clear. Historically, the OID4VP spec was going through significant flux -- so that we didn't want to write anything down that would just need to be later revised or that would entrench improper implementations. But, it seems that may have settled at this point so we could proceed.

> Also, I think that we should update both the description of this property as well as the name of the property itself. Since we have another step-level `create*` property that is a boolean (`createChallenge`), it seems inconsistent to use that verb for this property, which is not a boolean.

We could consider doing this, though it will affect existing implementations.

msporny:
The group discussed this on the 2025-02-25 telecon:

A PR should be raised that explains what the value of the `createAuthorizationRequest` is supposed to contain (with a concrete example).

msporny:
PR #502 has been raised to address this issue. This issue will be closed once that PR has been merged.

PR #502, opened 2025-07-28
Clarify `openId` values.

msporny:
This PR is an attempt to address issue #456 by adding language that clarifies what the `createAuthorizationRequest` property does.

I still don't understand what example we might want to include here, so need some help on that. Do we want to do something like this?

https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5.4

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/pull/502.html" title="Last updated on Aug 3, 2025, 8:11 PM UTC (af21354)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/502/be5922a...af21354.html" title="Last updated on Aug 3, 2025, 8:11 PM UTC (af21354)">Diff</a>


exchanges.yml
```diff
@@ -432,6 +432,7 @@ components:
           properties:
             createAuthorizationRequest:
               type: string
-              description: If present, auto-generate the OID4VP authorization request from the VPR. Either this or authorizationRequest must be present, but not both.
+              description: If present, identifies the name of the template variable (in <code>variables</code> for an <a href="#create-exchange">exchange</a>) that the auto-generated OpenID for Verifiable Presentations Authorization Request will be stored in for subsequent use in the exchange. Either this or <code>authorizationRequest</code> MUST be present, but not both.
```
dlongley:
```suggestion
              description: If present, identifies the name of the template variable (in <code>variables</code> for an <a href="#create-exchange">exchange</a>) that the auto-generated OpenID for Verifiable Presentations Authorization Request will be stored in for subsequent use in the exchange. Either this or <code>authorizationRequest</code> MUST be present, but not both. The value of the named variable
can be lazily generated when the Authorization Request is required, i.e., after the client of the exchange
chooses to use the OID4VP protocol. This feature allows the Authorization Request to be generated
by reusing values, e.g., a Verifiable Presentation Request, that only need to be specified once across
multiple protocol offerings.
```

exchanges.yml
```diff
@@ -432,6 +432,7 @@ components:
           properties:
```
kezike, line 432: Unable to add this comment to the `openId` line, so I will place it here. Can we rename this property to `oid4vp` or `OID4VP`, since that is a more accurate description and could be confusing with respect to the workflow’s `authorization.oauth2` schema?

dlongley: Hmm, well OAuth2 is specifically different from OpenID.

Additionally, the `openId` section covers *both* OID4VP and OID4VCI -- and some steps might require both of these at once. I imagine we'd have to have two separate sections for them, and if anything is shared across the protocols, we might have unnecessary duplication (that usually leads to bugs). So I'm not sure if possible confusion between `authorization.oauth2` and `.openId` is a strong enough reason to rename or split those sections (or to rename the `openId` section to something like `oid4` which may itself lead to other confusions?).

dlongley:
An example authorization request:

```json
{
  "response_type": "vp_token",
  "presentation_definition": {
    "id": "f1bd224a-0fed-469a-b64a-dad7cf63fd98",
    "input_descriptors": [
      {
        "id": "77d50c71-95ef-442c-a372-f32e17634fab",
        "constraints": {
          "fields": [
            {
              "path": [
                "$['@context']"
              ],
              "filter": {
                "type": "array",
                "contains": {
                  "type": "string",
                  "const": "https://www.w3.org/ns/credentials/v2"
                }
              }
            },
            {
              "path": [
                "$['type']"
              ],
              "filter": {
                "type": "array",
                "contains": {
                  "type": "string",
                  "const": "VerifiableCredential"
                }
              }
            },
            {
              "path": [
                "$['credentialSubject']['name']"
              ],
              "filter": {
                "type": "string"
              }
            }
          ]
        },
        "purpose": "We require a name credential to display your name when you post messages."
      }
    ]
  },
  "response_mode": "direct_post",
  "client_metadata": {
    "vp_formats_supported": {
      "ldp_vc": {
        "proof_type_values": [
          "DataIntegrityProof",
          "Ed25519Signature2020"
        ],
        "cryptosuite_values": [
          "ecdsa-rdfc-2019",
          "eddsa-rdfc-2022"
        ]
      }
    },
    "vp_formats": {
      "jwt_vp": {
        "alg": [
          "EdDSA",
          "Ed25519",
          "ES256",
          "ES384"
        ]
      },
      "jwt_vp_json": {
        "alg": [
          "EdDSA",
          "Ed25519",
          "ES256",
          "ES384"
        ]
      },
      "di_vp": {
        "proof_type": [
          "ecdsa-rdfc-2019",
          "eddsa-rdfc-2022",
          "Ed25519Signature2020"
        ]
      },
      "ldp_vp": {
        "proof_type": [
          "ecdsa-rdfc-2019",
          "eddsa-rdfc-2022",
          "Ed25519Signature2020"
        ]
      }
    }
  },
  "client_id": "https://some.example/workflows/z1A7ojitBF3pGpVzW17MgU7zm/exchanges/z1ACKwiBiXmuGjpKAdKMpBZXB/openid/client/authorization/response",
  "client_id_scheme": "redirect_uri",
  "response_uri": "https://some.example/workflows/z1A7ojitBF3pGpVzW17MgU7zm/exchanges/z1ACKwiBiXmuGjpKAdKMpBZXB/openid/client/authorization/response",
  "nonce": "z1ACKwiBiXmuGjpKAdKMpBZXB"
}
```

exchanges.yml
msporny: At present, I'm going to keep the "openId" section the same for the reasons that @dlongley suggested -- it's not clear that changing it to "OID4VP" will be better.

I'd like to get this PR merged so we at least have a step improvement in the spec. @kezike, please raise an issue if you disagree or think we should be tracking this to be resolved later.

Issue #509, opened 2025-08-05
Update presentation verification response schema


PatStLouis:
Following the decision made on #284, it was discussed that the presentation verification response schema should be updated to reflect each credential verification response separately in addition to the presentation verification.

This issue is to track discussions about what this schema should look like before opening a PR.

msporny:
PR #510 has been merged, closing.

PR #508, opened 2025-08-01
Clarify VP verification scope and add implementer flexibility.

bparth24:
- Updates specification text to explicitly state VP verification includes all contained VCs by default
- Updates POST /presentations/verify endpoint to reflect comprehensive verification of presentations and contained credentials
- Adds extensibility for implementer options to disable credential verification for debugging purposes
- Documents compositional API design enabling use of /credentials/verify for individual credential verification when VP credential checking is disabled
- Clarifies verification vs validation boundaries with business rules handled by calling applications
- Enhances HTTP response descriptions in /presentations/verify endpoint to clarify process success vs verification outcomes

To my understanding above updates are sufficient.

----
**Updates are highlighted**
----

_Section 3.9.2_
----

VP Verify Description

Screenshot 2025-08-01 at 3 48 33 PM

_`verifier.html` displays from `verifier.yml`_
----
Screenshot 2025-08-01 at 3 39 33 PM

[
Screenshot 2025-08-01 at 3 46 54 PM
](url)

---- 

Question for reviewers: Do we need to update the schema in the `/components/VerifyPresentationResult.yml` file?
  
Addresses #284.

@dlongley @msporny @TallTed - ready for review.

***
<a href="https://pr-preview.s3.amazonaws.com/bparth24/vc-api/pull/508.html" title="Last updated on Aug 6, 2025, 3:12 PM UTC (01ae6be)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/508/be5922a...bparth24:01ae6be.html" title="Last updated on Aug 6, 2025, 3:12 PM UTC (01ae6be)">Diff</a>

Issue #284, opened 2022-04-14
Does verifying a presentation include verifying its VCs?


jandrieu:
When educating clients in this work, I've always said that verifying a VP includes verifying the VCs entailed within.

However, Section 3.3.2 https://w3c-ccg.github.io/vc-api/#verify-presentation doesn't state whether that is correct or not.

Whatever the intention, we should clarify it in the text. 

For client-side simplicity, I'd recommend a VP endpoint that DOES verify the VCs, but if we do that we need more resilient responses. In particular, we should be able to report independently on each VC.

Further, for some of the use cases under discussion, it would be exceptionally useful to know why a verification failed. Did it fail the signature check? The expiry check? Or the status check?

jandrieu:
Also, does verifying the VP verify that the subject of the VC is the signer of the VP?

I think not. I think that is a validation rule applied by the Verifier App, not a function verified by the Verifier service. However, the application of this rule could be in either function.

mkhraisha:
> When educating clients in this work, I've always said that verifying a VP includes verifying the VCs entailed within.
> However, Section 3.3.2 https://w3c-ccg.github.io/vc-api/#verify-presentation doesn't state whether that is correct or not.
> Whatever the intention, we should clarify it in the text.

Verifying a VP should verify the VCs within, +1 to update the spec.

> For client-side simplicity, I'd recommend a VP endpoint that DOES verify the VCs, but if we do that we need more resilient responses. In particular, we should be able to report independently on each VC.
> Further, for some of the use cases under discussion, it would be exceptionally useful to know why a verification failed. Did it fail the signature check? The expiry check? Or the status check?
 ```
 {
   "verified":false,
   "checks":[
      "string"
   ],
   "warnings":[
      "string"
   ],
   "errors":[
      "credential X failed to verify signature",
      "credential Y failed a status check"
   ]
}
 ```

clehner:
Previously: #111

clehner:
> VP verification will NOT apply any business rules about any identifiers in the VCs and the signer of the VP. This means that holder==subject tests are the domain of validation (not verification) and performed by the verifier app, not the verifier service.

OK. What about VP holder ↔ VP proof verification method? (Analogous to VC issuer ↔ VC proof verification method?)

> What we don't have right now is an endpoint that adequately responds so callers can understand why the VP failed verification. IMO, an http response code of 200 is insufficient.
>
[...]

PR #510, opened 2025-08-05
Enhance VP verification response schema.

bparth24:
- Adds summary counts providing quick overview of credential verification results
- Implements detailed per-VC breakdown showing verification status for each credential
- Includes granular verification details for status checks, validity periods, proof verification, and schema validation
- Provides transparency about which VCs were checked vs skipped with explanatory reasons
- Adds comprehensive error reporting with problemDetails at presentation, credential, and proof levels
- Enables understanding of which proofs were verified and verification factors checked per credential
- Changes to file - `/components/VerifyPresentationResult.yml`

Addresses #509.

----
----

Screenshot 2025-08-05 at 2 54 06 PM

Screenshot 2025-08-05 at 2 54 25 PM

Screenshot 2025-08-05 at 2 54 40 PM

Screenshot 2025-08-05 at 2 54 50 PM

----
----

@dlongley @msporny @TallTed @PatStLouis - Ready for review

### DIFF
--- a/components/VerifyPresentationResult.yml
+++ b/components/VerifyPresentationResult.yml
@@ -19,72 +19,303 @@ components:
       properties:
         verified:
           type: boolean
-          description: Overall verification assertion of the VerifiablePresentation. This is set to `True` if no errors were detected during the verification process; otherwise, `False`. See error handling section for more guidance around errors, warnings, and validation and verification steps.
+          description: Overall verification assertion of the VerifiablePresentation. This is set to `True` if the VP and all checked VCs passed verification; otherwise, `False`. See error handling section for more guidance around errors, warnings, and validation and verification steps.
         verifiablePresentation:
           type: object
           description: The VerifiablePresentation used as the verification input.
         problemDetails:
           type: array
-          description: Array consisting of ProblemDetails objects.
+          description: Array consisting of ProblemDetails objects for overall presentation-level issues.
           items:
             $ref: "./ProblemDetails.yml#/components/schemas/ProblemDetails"
         results:
           type: object
           additionalProperties: false
-          description: Verification results to be included as a more verbose output. Keys are mapped to properties from the Verifiable Credential Data Model which are subject to validations.
+          description: Detailed verification results for both presentation and all credentials contained within it, by default. Keys are mapped to properties from the Verifiable Credential Data Model which are subject to validations.
           properties:
-            challenge:
+            presentation:
               type: object
-              description: If provided in the verification options, the result of validating the security challenge across all proofs provided.
+              additionalProperties: false
+              description: Verification results specific to the VerifiablePresentation itself (not its contained VCs).
               properties:
-                valid:
-                  type: boolean
-                  description: Result of validating the security challenge across all proofs provided.
-                input:
-                  type: string
-                  description: The challenge input string.
-            domain:
-              type: object
-              description: If provided in the verification options, the result of validating the security domain across all proofs provided.
-              properties:
-                valid:
-                  type: boolean
-                  description: Result of validating the security domain across all proofs provided.
-                input:
-                  type: string
-                  description: The security domain input string.
-            holder:
-              type: object
-              description: Result from validating the holder, if any.
-              properties:
-                valid:
-                  type: boolean
-                  description: Result of validating the holder.
-                input:
-                  type: string
-                  description: The holder input string.
-            proof:
-              type: array
-              description: Results from validating the proof objects if any.
-              items:
-                type: object
-                description: Results from validating a proof object.
-                properties:
-                  valid:
-                    type: boolean
-                    description: Results of validating the proof object.
-                  input:
+                challenge:
+                  type: object
+                  description: If provided in the verification options, the result of validating the security challenge across all proofs provided.
+                  properties:
+                    valid:
+                      type: boolean
+                      description: Result of validating the security challenge across all proofs provided.
+                    input:
+                      type: string
+                      description: The challenge input string.
+                domain:
+                  type: object
+                  description: If provided in the verification options, the result of validating the security domain across all proofs provided.
+                  properties:
+                    valid:
+                      type: boolean
+                      description: Result of validating the security domain across all proofs provided.
+                    input:
+                      type: string
+                      description: The security domain input string.
+                holder:
+                  type: object
+                  description: Result from validating the holder, if any.
+                  properties:
+                    valid:
+                      type: boolean
+                      description: Result of validating the holder.
+                    input:
+                      type: string
+                      description: The holder input string.
+                proof:
+                  type: array
+                  description: Results from validating the proof objects if any.
+                  items:
                     type: object
-                    description: The proof object.
+                    description: Results from validating a proof object.
+                    properties:
+                      valid:
+                        type: boolean
+                        description: Results of validating the proof object.
+                      input:
+                        type: object
+                        description: The proof object.
+                      problemDetails:
+                        type: array
+                        description: Issues sepecific to this proof.
+                        items:
+                          $ref: "./ProblemDetails.yml#/components/schemas/ProblemDetails"
+            credentials:
+              type: object
+              description: Verification results for the credentials contained in the presentation.
+              additionalProperties: false
+              properties:
+                summary:
+                  type: object
+                  description: Summary counts providing a quick overview of credential verification results.
+                  additionalProperties: false
+                  properties:
+                    totalCount:
+                      type: integer
+                      description: Total number of VCs in the presentation.
+                    checkedCount:
+                      type: integer
+                      description: Number of VCs that were actually verified.
+                    skippedCount:
+                      type: integer
+                      description: Number of VCs that were skipped due to options or configuration.
+                    passedCount:
+                      type: integer
+                      description: Number of checked VCs that passed verification.
+                    failedCount:
+                      type: integer
+                      description: Number of checked VCs that failed verification.
+                results:
+                  type: array
+                  description: Detailed verification results for each VC in the presentation, in the same order as they appear in the verifiableCredential array.
+                  items:
+                    type: object
+                    description: Verification results for a single VC.
+                    additionalProperties: false
+                    properties:
+                      index:
+                        type: integer
+                        description: Index of this VC in the VP's verifiableCredential array.
+                      wasChecked:
+                        type: boolean
+                        description: Whether this VC was actually verified. If false, detailed verification properties will not be present.
+                      verified:
+                        type: boolean
+                        description: Whether this specific VC pass verification. Only meaningful if wasChecked is true.
+                      skipReason:
+                        type: boolean
+                        description: Explanation of why verification was skipped for this VC. Only present if wasChecked is false.
+                      validFrom:
+                        type: object
+                        description: Results of validating the validFrom property, if present in the verifiableCredential. Only present if wasChecked is true.
+                        properties:
+                          valid:
+                            type: boolean
+                            description: Validation result.
+                          input:
+                            type: string
+                            description: The validFrom value.
+                      validUntil:
+                        type: object
+                        description: Results of validating the validUntil property, if present in the verifiableCredential. Only present if wasChecked is true.
+                        properties:
+                          valid:
+                            type: boolean
+                            description: Validation result.
+                          input:
+                            type: string
+                            description: The validUntil value.
+                      credentialSchema:
+                        type: array
+                        description: Results from validating the credentialSchema objects if any. Only present if wasChecked is true.
+                        items:
+                          type: object
+                          description: Results from validating a credentialSchema object.
+                          properties:
+                            valid:
+                              type: boolean
+                              description: Results of validating the credentialSchema object.
+                            input:
+                              type: string
+                              description: The credentialSchema object.
+                      credentialStatus:
+                        type: array
+                        description: Results from validating the credentialStatus objects if any. Only present if wasChecked is true.
+                        items:
+                          type: object
+                          description: Results of validating a credentialStatus object.
+                          properties:
+                            value:
+                              type: integer
+                              description: Value of the statusEntry Bit.
+                            valid:
+                              type: boolean
+                              description: Results of validating the credentialStatus object.
+                            input:
+                              type: string
+                              description: The credentialStatus object.
+                      proof:
+                        type: array
+                        description: Results from validating the proof objects if any. Only present if wasChecked is true.
+                        items:
+                          type: object
+                          description: Results of validating a proof object.
+                          properties:
+                            valid:
+                              type: boolean
+                              description: Results of validating the proof object.
+                            input:
+                              type: object
+                              description: The proof object.
+                            problemDetails:
+                              type: array
+                              description: Issues specific to this proof.
+                              items:
+                                $ref: "./ProblemDetails.yml#/components/schemas/ProblemDetails"
+                      problemDetails:
+                        type: array
+                        description: Issues specific to this individual VC, present regardless of wasChecked status.
+                        items:
+                          $ref: "./ProblemDetails.yml#/components/schemas/ProblemDetails"
       example: {
         "verified": false,
         "verifiablePresentation": {},
-        "problemDetails": [{
-          "title": "PARSING_ERROR",
-          "type": "https://www.w3.org/TR/vc-data-model#PARSING_ERROR"
-        }],
+        "problemDetails": [
+          {
+            "title": "PARSING_ERROR",
+            "type": "https://www.w3.org/TR/vc-data-model#PARSING_ERROR"
+          }
+        ],
         "results": {
-          "holder": {},
-          "proof": []
+          "presentation": {
+            "challenge": {
+              "valid": true,
+              "input": "d436f0c8-fbd9-4e48-bbb2-55fc5d0920a8"
+            },
+            "domain": {
+              "valid": true,
+              "input": "example.com"
+            },
+            "holder": {
+              "valid": true,
+              "input": "did:example:123"
+            },
+            "proof": [{
+              "valid": true,
+              "input": {},
+              "problemDetails": []
+            }
+           ]
+          },
+          "credentials": {
+            "summary": {
+              "totalCount": 3,
+              "checkedCount": 2,
+              "skippedCount": 1,
+              "passedCount": 1,
+              "failedCount": 1
+            },
+            results: [
+              {
+                "index": 0,
+                "wasChecked": true,
+                "verified": true,
+                "validFrom": {
+                  "valid": true,
+                  "input": "2020-03-16T22:37:26.544Z"
+                },
+                "validUntil": {
+                  "valid": true,
+                  "input": "2025-03-16T22:37:26.544Z"
+                },
+                "credentialStatus": [
+                  {
+                    "value": 0,
+                    "valid": true,
+                    "input": {}
+                  }
+                ],
+                "proof": [
+                  {
+                    "valid": true,
+                    "input": {},
+                    "problemDetails": []
+                  }
+                ],
+                "problemDetails": []
+              },
+              {
+                "index": 1,
+                "wasChecked": true,
+                "verified": false,
+                "validFrom": {
+                  "valid": true,
+                  "input": "2020-03-16T22:37:26.544Z"
+                },
+                "validUntil": {
+                  "valid": false,
+                  "input": "2022-03-16T22:37:26.544Z"
+                },
+                "credentialStatus": [
+                  {
+                    "value": 1,
+                    "valid": false,
+                    "input": {}
+                  }
+                ],
+                "proof": [
+                  {
+                    "valid": false,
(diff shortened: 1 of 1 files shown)


components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
       properties:
         verified:
           type: boolean
-          description: Overall verification assertion of the VerifiablePresentation. This is set to `True` if no errors were detected during the verification process; otherwise, `False`. See error handling section for more guidance around errors, warnings, and validation and verification steps.
+          description: Overall verification assertion of the VerifiablePresentation. This is set to `True` if the VP and all checked VCs passed verification; otherwise, `False`. See error handling section for more guidance around errors, warnings, and validation and verification steps.
```
TallTed: Unless there's a way to specify "only check these VCs among those contained in the VP", `all checked` -> `all contained` Also, tweak to the `around` items listing.
```suggestion
          description: Overall verification assertion of the VerifiablePresentation. This is set to `True` if the VP and all contained VCs passed verification; otherwise, `False`. See error handling section for more guidance around errors and warnings, and validation and verification steps.
```

TallTed: I think we'd decided not to abbreviate?
```suggestion
          description: Overall verification assertion of the VerifiablePresentation. This is set to `True` if the VerifiablePresentation and all checked VerifiableCredentials passed verification; otherwise, `False`. See error handling section for more guidance around errors, warnings, and validation and verification steps.
```

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
           type: array
-          description: Array consisting of ProblemDetails objects.
+          description: Array consisting of ProblemDetails objects for overall presentation-level issues.
           items:
             $ref: "./ProblemDetails.yml#/components/schemas/ProblemDetails"
         results:
           type: object
           additionalProperties: false
-          description: Verification results to be included as a more verbose output. Keys are mapped to properties from the Verifiable Credential Data Model which are subject to validations.
+          description: Detailed verification results for both presentation and all credentials contained within it, by default. Keys are mapped to properties from the Verifiable Credential Data Model which are subject to validations.
```
TallTed:
```suggestion
          description: Detailed verification results for both VerifiablePresentation and all VerifiableCredentials contained within it, by default. Keys are mapped to properties from the Verifiable Credential Data Model which are subject to validations.
```

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
           properties:
-            challenge:
+            presentation:
               type: object
-              description: If provided in the verification options, the result of validating the security challenge across all proofs provided.
+              additionalProperties: false
+              description: Verification results specific to the VerifiablePresentation itself (not its contained VCs).
```
TallTed:
```suggestion
              description: Verification results specific to the VerifiablePresentation itself (not its contained VerifiableCredentials).
```

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                        description: The proof object.
+                      problemDetails:
+                        type: array
+                        description: Issues sepecific to this proof.
+                        items:
+                          $ref: "./ProblemDetails.yml#/components/schemas/ProblemDetails"
+            credentials:
               type: object
-              description: Result from validating the holder, if any.
+              description: Verification results for the credentials contained in the presentation.
```
TallTed:
```suggestion
              description: Verification results for the VerifiableCredentials contained in the VerifiablePresentation.
```

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
-                type: object
-                description: Results from validating a proof object.
-                properties:
-                  valid:
-                    type: boolean
-                    description: Results of validating the proof object.
-                  input:
+                summary:
+                  type: object
+                  description: Summary counts providing a quick overview of credential verification results.
```
TallTed:
```suggestion
                  description: Summary counts providing a quick overview of VerifiableCredential verification results.
```

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                  additionalProperties: false
+                  properties:
+                    totalCount:
+                      type: integer
+                      description: Total number of VCs in the presentation.
```
TallTed:
```suggestion
                      description: Total number of VerifiableCredentials in the VerifiablePresentation.
```

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                      description: Number of VCs that were actually verified.
+                    skippedCount:
+                      type: integer
+                      description: Number of VCs that were skipped due to options or configuration.
+                    passedCount:
+                      type: integer
+                      description: Number of checked VCs that passed verification.
+                    failedCount:
+                      type: integer
+                      description: Number of checked VCs that failed verification.
```
TallTed:
```suggestion
                      description: Number of VerifiableCredentials that were actually verified.
                    skippedCount:
                      type: integer
                      description: Number of VerifiableCredentials that were skipped due to options or configuration.
                    passedCount:
                      type: integer
                      description: Number of checked VerifiableCredentials that passed verification.
                    failedCount:
                      type: integer
                      description: Number of checked VerifiableCredentials that failed verification.
```

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                results:
+                  type: array
+                  description: Detailed verification results for each VC in the presentation, in the same order as they appear in the verifiableCredential array.
```
TallTed:
```suggestion
                  description: Detailed verification results for each VerifiableCredential in the VerifiablePresentation, in the same order as they appear in the verifiableCredential array.
```

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                        description: Whether this VC was actually verified. If false, detailed verification properties will not be present.
+                      verified:
+                        type: boolean
+                        description: Whether this specific VC pass verification. Only meaningful if wasChecked is true.
+                      skipReason:
+                        type: boolean
+                        description: Explanation of why verification was skipped for this VC. Only present if wasChecked is false.
+                      validFrom:
+                        type: object
+                        description: Results of validating the validFrom property, if present in the verifiableCredential. Only present if wasChecked is true.
```
TallTed:
```suggestion
                    description: Verification results for a single VerifiableCredential.
                    additionalProperties: false
                    properties:
                      index:
                        type: integer
                        description: Index of this VerifiableCredential in the VerifiablePresentation's verifiableCredential array.
                      wasChecked:
                        type: boolean
                        description: Whether this VerifiableCredential was actually verified. If false, detailed verification properties will not be present.
                      verified:
                        type: boolean
                        description: Whether this specific VerifiableCredential pass verification. Only meaningful if wasChecked is true.
                      skipReason:
                        type: boolean
                        description: Explanation of why verification was skipped for this VerifiableCredential. Only present if wasChecked is false.
                      validFrom:
                        type: object
                        description: Results of validating the validFrom property, if present in the verifiableCredential. Only present if wasChecked is true.
```

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                              type: object
+                              description: The proof object.
+                            problemDetails:
+                              type: array
+                              description: Issues specific to this proof.
+                              items:
+                                $ref: "./ProblemDetails.yml#/components/schemas/ProblemDetails"
+                      problemDetails:
+                        type: array
+                        description: Issues specific to this individual VC, present regardless of wasChecked status.
```
TallTed:
```suggestion
                        description: Issues specific to this individual VerifiableCredential, present regardless of wasChecked status.
```

components/VerifyPresentationResult.yml
bparth24: @TallTed I will make sure not to use abbreviation and correct all other places as well.

bparth24: > Unless there's a way to specify "only check these VCs among those contained in the VP", `all checked` -> `all contained` Also, tweak to the `around` items listing.

@TallTed Yes, there will be options for users to check/select VCs among those contained in the VP. This was mentioned in Issue #284 & PR #508. But `contained` word looks okay to me. No strong opinion.

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
       properties:
         verified:
           type: boolean
-          description: Overall verification assertion of the VerifiablePresentation. This is set to `True` if no errors were detected during the verification process; otherwise, `False`. See error handling section for more guidance around errors, warnings, and validation and verification steps.
+          description: Overall verification assertion of the VerifiablePresentation. This is set to `True` if the VerifiablePresentation and all contained VerifiableCredentials passed verification; otherwise, `False`. See error handling section for more guidance around errors and warnings, and validation and verification steps.
```
TallTed: "all contained" is only appropriate if all VCs within VP will be checked, i.e., if the user cannot select *some* VCs to check.
```suggestion
          description: Overall verification assertion of the VerifiablePresentation. This is set to `True` if the VerifiablePresentation and all checked VerifiableCredentials passed verification; otherwise, `False`. See error handling section for more guidance around errors and warnings, and validation and verification steps.
```

TallTed (changes requested):
Just a tiny reversion.

TallTed (approved):

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                      description: Number of VerifiableCredentials that were actually verified.
+                    skippedCount:
+                      type: integer
+                      description: Number of VerifiableCredentials that were skipped due to options or configuration.
+                    passedCount:
+                      type: integer
+                      description: Number of checked VerifiableCredentials that passed verification.
+                    failedCount:
+                      type: integer
+                      description: Number of checked VerifiableCredentials that failed verification.
```
msporny: This doesn't seem like something we want to do for the following reasons:

* You can get `totalCount` by just getting the length of the credentials array.
* You shouldn't skip any checks for any proofs AFAICT, that could lead to severe security vulnerabilities.
* You can get checked/passed/failed by looking at the `verified` field for each credential.

I don't think we should have a `summary` object as all the information is redundant.

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                results:
+                  type: array
+                  description: Detailed verification results for each VerifiableCredential in the VerifiablePresentation, in the same order as they appear in the verifiableCredential array.
+                  items:
                     type: object
-                    description: The proof object.
+                    description: Verification results for a single VerifiableCredential.
+                    additionalProperties: false
+                    properties:
+                      index:
```
msporny: We can't presume that the index means anything for set-based proofs, and order is not guaranteed. We should remove `index` from here.

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                    skippedCount:
+                      type: integer
+                      description: Number of VerifiableCredentials that were skipped due to options or configuration.
+                    passedCount:
+                      type: integer
+                      description: Number of checked VerifiableCredentials that passed verification.
+                    failedCount:
+                      type: integer
+                      description: Number of checked VerifiableCredentials that failed verification.
+                results:
```
msporny: We probably shouldn't have a `results` object... we don't have the same for `presentation`. It should probably just be rolled up into `credentials`.

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                  items:
                     type: object
-                    description: The proof object.
+                    description: Verification results for a single VerifiableCredential.
+                    additionalProperties: false
+                    properties:
+                      index:
+                        type: integer
+                        description: Index of this VerifiableCredential in the VerifiablePresentation's verifiableCredential array.
+                      wasChecked:
```
msporny: Why would you not check a proof?

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                        type: boolean
+                        description: Whether this VerifiableCredential was actually verified. If false, detailed verification properties will not be present.
+                      verified:
+                        type: boolean
+                        description: Whether this specific VerifiableCredential pass verification. Only meaningful if wasChecked is true.
+                      skipReason:
```
msporny: Why would you skip checking a proof that is vital for security?

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                              type: string
+                              description: The credentialSchema object.
+                      credentialStatus:
+                        type: array
+                        description: Results from validating the credentialStatus objects if any. Only present if wasChecked is true.
+                        items:
+                          type: object
+                          description: Results of validating a credentialStatus object.
+                          properties:
+                            value:
```
msporny: Hmm, this seems like you would need to pass in the expected status in the check... I don't think we can report on 'valid' or not without that. We might need something like: `expectedValue`.

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
       example: {
         "verified": false,
         "verifiablePresentation": {},
-        "problemDetails": [{
-          "title": "PARSING_ERROR",
-          "type": "https://www.w3.org/TR/vc-data-model#PARSING_ERROR"
-        }],
+        "problemDetails": [
+          {
+            "title": "PARSING_ERROR",
```
msporny: I don't think this is what we meant by "human-readable" string in the spec. Something like this might be better:

```json
{
  "type": "https://www.w3.org/TR/vc-data-model#PARSING_ERROR",
  "title": "Parsing error",
  "description": "There was a parsing error on line 32, column 5: A comma was expected after the \" character."
}
```

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+              "valid": true,
+              "input": "example.com"
+            },
+            "holder": {
+              "valid": true,
+              "input": "did:example:123"
+            },
+            "proof": [{
+              "valid": true,
+              "input": {},
```
msporny: Why is this blank and not the proof that was checked?

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                    "input": {}
+                  }
+                ],
+                "proof": [
+                  {
+                    "valid": false,
+                    "input": {},
+                    "problemDetails": [
+                      {
+                        "title": "CRYPTOGRAPHIC_SECURITY_ERROR",
```
msporny: Not easily human readable

components/VerifyPresentationResult.yml
```diff
@@ -19,72 +19,303 @@ components:
...
+                        "type": "https://www.w3.org/TR/vc-data-model#CRYPTOGRAPHIC_SECURITY_ERROR"
+                      }
+                    ]
+                  }
+                ],
+                "problemDetails": [
+                  {
+                    "title": "CREDENTIAL_EXPIRED",
```
msporny: Not easily human readable. See this for a better example:

https://www.rfc-editor.org/rfc/rfc9457#appendix-C

Remember that people will copy/paste these examples into production code. Any bad practice we provide here will be multiplied a hundred times over in deployed software.

msporny (changes requested):
Thank you for a PR that we can discuss, @bparth24.

There are significant design issues with the response format that we'll have to discuss during the next call. 

I also note inconsistencies in `True` vs `true`, use of backticks (which won't work in OAS files when rendered in ReSpec, you have to use &lt;code> tags instead, not referencing specs using HTTP URLs (VCDM v2), and not wrapping type names in code blocks. Please make a pass to fix all of that because it'll create spec debt that we'll just have to clean up later.

components/VerifyPresentationResult.yml
bparth24: @msporny 

**On the summary redundancy:** I completely agree with your assessment. My initial instinct was the same - that the summary contains entirely derivable information that clients can calculate from the detailed results array. I don't have a strong preference for keeping it, and removing it would simplify the schema significantly.

**Regarding the skipping functionality:** I want to clarify that this wasn't intended for bypassing security checks. The skipping capability was included specifically because it was mentioned in Issue #284 as beneficial for troubleshooting and debugging scenarios where developers need to test only VP-level verification while isolating credential-specific issues.

However, I acknowledge your security concern. Depending on the workflow or use case, skipping proof checks could introduce complexity and potential misuse. 

**Next steps:** Remove the summary object and reconsider whether skipping functionality is necessary for the initial implementation.

components/VerifyPresentationResult.yml
bparth24: @msporny 

Noted, and I agree with you about removing the `index` property.

components/VerifyPresentationResult.yml
bparth24: @msporny 

Good point - without context about expectations, the `valid` field is unclear.

**Option 1:** Add `expectedValue`
```
expectedValue:
  type: integer
  description: Expected value for this credential to be considered valid.
```

**Option 2:** Add `statusPurpose` for context
```
statusPurpose:
  type: string
  description: Purpose of this status check (e.g., "revocation", "suspension").
```

Additionally, I noticed the `input` type should be `object` and not `string`, which I will correct together.

components/VerifyPresentationResult.yml
bparth24: @msporny I missed it. Added inputs.

bparth24:
@msporny 

`I also note inconsistencies in `True` vs `true`, use of backticks (which won't work in OAS files when rendered in ReSpec, you have to use <code> tags instead, not referencing specs using HTTP URLs (VCDM v2), and not wrapping type names in code blocks. Please make a pass to fix all of that because it'll create spec debt that we'll just have to clean up later.`

Addressed in commit - https://github.com/w3c-ccg/vc-api/pull/510/commits/f4b15badd83dba5e5a7e642c1a7813168659178c

components/VerifyPresentationResult.yml
bparth24: @msporny 

> Not easily human readable. See this for a better example:
> 
> https://www.rfc-editor.org/rfc/rfc9457#appendix-C
> 
> Remember that people will copy/paste these examples into production code. Any bad practice we provide here will be multiplied a hundred times over in deployed software.

Addressed in commit - https://github.com/w3c-ccg/vc-api/commit/f4b15badd83dba5e5a7e642c1a7813168659178c

components/VerifyPresentationResult.yml
bparth24: @msporny 

> Not easily human readable

Addressed in commit - https://github.com/w3c-ccg/vc-api/commit/f4b15badd83dba5e5a7e642c1a7813168659178c

components/VerifyPresentationResult.yml
bparth24: @msporny 

> Why is this blank and not the proof that was checked?

Addressed in commit - https://github.com/w3c-ccg/vc-api/commit/f4b15badd83dba5e5a7e642c1a7813168659178c

components/VerifyPresentationResult.yml
bparth24: @msporny 

> I don't think this is what we meant by "human-readable" string in the spec. Something like this might be better:
> 
> ```json
> {
>   "type": "https://www.w3.org/TR/vc-data-model#PARSING_ERROR",
>   "title": "Parsing error",
>   "description": "There was a parsing error on line 32, column 5: A comma was expected after the \" character."
> }
> ```

Addressed in commit - https://github.com/w3c-ccg/vc-api/commit/f4b15badd83dba5e5a7e642c1a7813168659178c

bparth24:
@bparth24 to review [Verifiable Credential Schema](https://github.com/w3c-ccg/vc-api/blob/main/components/VerifiableCredential.yml)
- ensure it has all the fields required
- Ideally, Verifiable Presentation will use VC as a sub-schema (clean up results object)
- remove summary object

bparth24:
> The group discussed this on the 2025-08-12 telecon.
> 
> @bparth24 to review [Verifiable Credential Schema](https://github.com/w3c-ccg/vc-api/blob/main/components/VerifiableCredential.yml)
> 
> * ensure it has all the fields required
> * Ideally, Verifiable Presentation will use VC as a sub-schema (clean up results object)
> * remove summary object

Addressed all points as discussed above in this commit https://github.com/w3c-ccg/vc-api/pull/510/commits/0f22f53fadea28f58e02d78f1f255d8afae8bdf6.

@msporny cc: @TallTed @dlongley Ready for review.

msporny (approved):

kmk142789 (approved):

Issue #496, opened 2025-07-20
Define OID4VCI and OID4VP protocol identifiers


ottonomy:
The Interactions / protocols object has an example property `oid4vp`, but a different issue in this repo used the example `OID4VP`. 

For clarity, I would love to see a section added for each OID4VP and OID4VCI to describe how an [OID4VP](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html) (`openid4vp://`) or [OID4VCI](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html) (`openid-credential-offer://`) launch url could be added to the protocols list.

This may require some collaboration or input from the OpenID Federation community.

msporny:
PR #512 has been raised to address this issue.

PR #512, opened 2025-08-10
Fix "oid4vp" -> "OID4VP" in interaction protocol identifier.

msporny:
This PR is an attempt to address issue #496 by fixing the interaction protocol identifier for OID4VP.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/pull/512.html" title="Last updated on Aug 17, 2025, 7:16 PM UTC (7a3195b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/512/eebda3f...7a3195b.html" title="Last updated on Aug 17, 2025, 7:16 PM UTC (7a3195b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1007,18 +1007,18 @@ all <a>verifiable credentials</a> contained within it.
 The verification process includes:
         </p>
         <ul>
-          <li>Verifying the presentation's own proof (including domain 
+          <li>Verifying the presentation's own proof (including domain
 and challenge validation)</li>
-          <li>Verifying each contained verifiable credential's proof, 
+          <li>Verifying each contained verifiable credential's proof,
 status, and validity period(s)</li>
-          <li>Checking that the holder in the presentation matches 
+          <li>Checking that the holder in the presentation matches
 the verification method used in the presentation's proof</li>
         </ul>
 
         <p>
-Business rule validation (such as verifying that credential subjects 
-match the presentation holder, or authorization policies about who can 
-present which credentials) is outside the scope of this verification 
+Business rule validation (such as verifying that credential subjects
+match the presentation holder, or authorization policies about who can
+present which credentials) is outside the scope of this verification
 endpoint and should be performed by the calling application.
         </p>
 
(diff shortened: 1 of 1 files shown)


msporny:
> I see 6 chars changed in 1 line. I see 10 other lines with EOL whitespace removed.

Yes, unfortunately, someone doing PRs doesn't have "strip trailing spaces" turned on, so when I save, it strips trailing spaces, which then shows up in the next PR. We can ignore those lines... I usually quietly do this on main before doing new feature branches, but it didn't happen this time.

index.html
```diff
@@ -1792,16 +1792,16 @@ <h4>Interaction Protocols Response</h4>
   "protocols": {
     "website": "https://app.example/redirects/z8j3kfk2lQ",
     "vcapi": "https://saas.example/workflows/123/exchanges/987",
-    "oid4vp": "openid4vp://?client_id=https%3A%2F%2Fapp.example%2Fworkflows%2F123%2Fexchanges%2F987%2Fopenid%2Fclient%2Fauthorization%2Fresponse&request_uri=https%3A%2F%2Fapp.example%2Fworkflows%2F123%2Fexchanges%2F987%2Fopenid%2Fclient%2Fauthorization%2Frequest'"
+    "OID4VP": "openid4vp://?client_id=https%3A%2F%2Fapp.example%2Fworkflows%2F123%2Fexchanges%2F987%2Fopenid%2Fclient%2Fauthorization%2Fresponse&request_uri=https%3A%2F%2Fapp.example%2Fworkflows%2F123%2Fexchanges%2F987%2Fopenid%2Fclient%2Fauthorization%2Frequest'"
```
kezike, line 2287: Should we add `OID4VCI` here as well?

dlongley: Maybe in a separate example -- since it would be unusual to see both in the same response. OID4VCI and OID4VP are split between issuance and verification, so if the exchange involves any issuance, you'd see OID4VCI, and if not, that's when you'd see OID4VP. I doubt there's a use case for having both for the same exchange -- and it would probably be confusing.

kezike: Agreed! I just wanted to make sure that we were capturing it somewhere, but just noticed that we at least have it in the schema in the `Get Exchange Protocols` section, so may not be necessary to have both examples 👍🏾

Issue #514, opened 2025-08-12
Change 'valid' to something else in verification results


msporny:
The current usage of the property `valid` is confusing to developers because it is not clear if business validation or syntax validation, or some other form of validation is occurring. We should rename the property to be more specific about the action that is being performed; verification that the field value is conformant to what the specification requires.

PR #518, opened 2025-08-17
Replace incorrect use of "valid" with "verified".

msporny:
This PR is an attempt to address issue #514 by replacing the incorrect use of "valid" with "verified".

### DIFF
--- a/components/VerifyCredentialResult.yml
+++ b/components/VerifyCredentialResult.yml
@@ -19,96 +19,96 @@ components:
       properties:
         verified:
           type: boolean
-          description: Overall verification assertion of the VerifiableCredential. This is set to <code>true</code> if no errors were detected during the verification process; otherwise, <code>false</code>. See error handling section for more guidance around errors, warnings, and validation and verification steps.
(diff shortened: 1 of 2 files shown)


dlongley (approved):

components/VerifyCredentialResult.yml
```diff
@@ -19,96 +19,96 @@ components:
...
               type: array
-              description: Results from validating the credentialStatus objects if any.
+              description: Results from verifying the credentialStatus objects if any.
               items:
                 type: object
-                description: Results of validating a credentialStatus object.
+                description: Results of verifying a credentialStatus object.
                 properties:
                   value:
                     type: integer
```
msporny:
```suggestion
                    type: any
```

Other status mechanisms can have different values.

components/VerifyCredentialResult.yml
```diff
@@ -19,96 +19,96 @@ components:
...
                     description: Value of the statusEntry Bit.
```
msporny:
```suggestion
                    description: The specific status value associated with the status entry.
```

Issue #515, opened 2025-08-12
JSON-LD with comments before, and sameline, and after are very confusing


TallTed:
_Originally posted by @TallTed in https://github.com/w3c-ccg/vc-api/pull/511#discussion_r2267720081_

https://github.com/w3c-ccg/vc-api/blob/a6034a0c4c11e99fec7eb6c46d77dd6efd86b9ac/index.html#L1116-L1157

These lines are very confusing at first read. Comments before, and sameline, and after? After reading a few times, it becomes clear that (for instance)
* 1117 applies to the whole object in which it is contained, 1116-1143
* 1118 applies only to the line in which it is contained, 1118
* 1119-1120 apply to the line which follows them, 1121

It will be non-trivial to rework this and other examples, but it would be a benefit to everyone downstream of us specification authors, so I think it would be worthwhile. The rework would be to make the comments *always* (or nearly so) have the same physical relation (i.e., _precede_ or _sameline_ or _follow_) to the property (or whatever) to which they are meant to apply, and possibly add some empty lines to clearly separate one intended cluster of JSON-LD with its (non-JSON-LD) `//`-delimited comments. It might also help to include the key (e.g., `credentialQuery`, `required`, `reason`; possibly even retaining the quotation marks that are seen in the non-commented portion JSON-LD, e.g., `"reason"`) in the comment(s) which apply to that key.

PR #519, opened 2025-08-17
Fix credential query examples.

msporny:
This PR is an attempt to address issue #515 by fixing the credential query examples.


index.html
```diff
@@ -1157,51 +1154,36 @@ <h3>Query By Example</h3>
...
-            {
-              "required": true,
-              "issuer": "urn:some:required:issuer"
-            }
-          ]
+  "query": [{
+    "type": "QueryByExample",
+    "credentialQuery": [{
+      <span class="comment">// (Optional) Defaults to 'true' if omitted</span>
+      "required": false,
```
dlongley: We can keep `required` in here for now, but I think it might be an anti-pattern to "ask for things that aren't required". It's being used to differentiate between "AND" and "OR" but could also be used to just ask for "other unnecessary things". I'm not sure if we can easily differentiate these, but it's worth discussing/looking into. Maybe "required" can be kept and only clearly defined as relating to query language of choice instead of pertaining to the contents being requested -- or maybe we need a different name/term for this.

It might be better for privacy for a second request to be made for "optional" things (but treating them as required in that second request) that is, for example, gated by telling the user, on a verifier coordinator website, that they can provide some optional information to continue if they'd like.

This kind of friction that arises from cleanly separating required requests from optional ones, such that they always appear as "extra asks" might be a good idea.

---

Instead of "required", and option could be to express it as something like "choice" with a string value identifying the particular choice. This might also allow for quicker selection in the verifier's evaluation engine on the other end, e.g., "choice": "A", "choice": "B" ... with an option for the responder to include these tags/IDs somewhere in their response.

index.html
```diff
@@ -1445,15 +1427,22 @@ <h3>The DID Authentication Response Format</h3>
       <section>
         <h3>Authorization Capability Request</h3>
 
+        <p class="issue">
+It is highly unlikely that Authorization Capability queries and responses
+will be standardized at this time. This example is provided to demonstrate that
+verifiable credentials and mDLs/mdocs are not the only type of credential
+that can be queried.
```
dlongley: No need to guess/estimate probabilities in this issue text:

```suggestion
Authorization Capability queries and responses might not be standardized
at this time. This example is provided to demonstrate that verifiable
credentials and mDLs/mdocs are not the only type of credential that can be
queried.
```

index.html
```diff
@@ -1445,15 +1427,22 @@ <h3>The DID Authentication Response Format</h3>
...
 Capabilities or "zcaps" in the Verifiable Presentation.
         </p>
 
-        <pre class="example nohighlight" title="A zCap Request query">
+        <pre class="example nohighlight" title="An Authorization Capability Request query">
 {
   query: [{
-    type: 'ZcapQuery',
+    type: 'AuthorizationCapabilityQuery',
     capabilityQuery: [{
```
dlongley: I can't make a suggestion below for this, but since we're here, "invoker" and "delegator" should be removed and replaced with just "controller".

index.html
msporny: Fixed in efc11d5fd42a55b890d903300b18dbb0ef9af0b6. Went with `group`, which we can bikeshed in future PRs.

index.html
msporny: Partial fix in 486c97127e20d360e317da300a8091579648afc9.

Issue #513, opened 2025-08-10
Ensure "callers of endpoints" properly references workflow instances instead of (or in preference to) issuer coordinators


dlongley:
I suspect we never did a second pass on these tables after adding the workflow service.

For example, in section [3.7.2](https://w3c-ccg.github.io/vc-api/#issuer-service) the workflow service should be shown as the first choice for making a call to the `/credentials/issue` endpoint, even if there might be cases where an issuer coordinator would use that endpoint directly. We should preference the newer workflow architecture that provides better separation of concerns and reusability (and better wallet protocol management, etc.) here.

The same should apply to where verifier coordinators are referenced as directly calling verification endpoints instead of letting this flow through a layer where different end user/wallet protocols can be selected, etc.

msporny:
PR #525 has been raised to address this issue.

eric-schuh:
@msporny I believe we can close this issue as it was resolved in https://github.com/w3c-ccg/vcalm/pull/565

PR #525, opened 2025-09-01
Update API Component Overview to include workflows.

msporny:
This PR is an attempt to address issue #513 by adding the Workflow Service to callers of existing issuing/verifying APIs.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/pull/525.html" title="Last updated on Sep 2, 2025, 7:20 PM UTC (b2c1c0e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-api/525/1575102...b2c1c0e.html" title="Last updated on Sep 2, 2025, 7:20 PM UTC (b2c1c0e)">Diff</a>

### DIFF
--- a/holder.yml
+++ b/holder.yml
@@ -25,6 +25,7 @@ paths:
       x-expectedCaller:
         - Holder Coordinator
         - Issuer Coordinator
+        - Workflow Service
       operationId: getCredential
       parameters:
         - $ref: "./components/parameters/path/ObjectId.yml"
@@ -112,6 +113,7 @@ paths:
       operationId: getPresentation
       x-expectedCaller:
         - Holder Coordinator
+        - Workflow Service
       parameters:
         - $ref: "./components/parameters/path/ObjectId.yml"
       responses:
@@ -205,6 +207,7 @@ paths:
       operationId: createPresentation
       x-expectedCaller:
         - Holder Coordinator
+        - Workflow Service
       description: Creates a presentation and returns it in the response body.
       requestBody:
         content:
--- a/issuer.yml
+++ b/issuer.yml
@@ -56,7 +56,7 @@ paths:
         - zCap: []
       operationId: updateCredentialStatus
       x-expectedCaller:
-        - Issuer Service
+        - Issuer Coordinator
       description: Updates the status of an issued credential.
       requestBody:
         content:


holder.yml
```diff
@@ -22,7 +22,9 @@ paths:
        - oAuth2: []
        - zCap: []
       summary: Gets a credential or verifiable credential by ID. To get a credential that does not have credential.id set but has an associated credentialId value, pass credentialId instead.
-      x-expectedCaller: "Issuer Coordinator"
+      x-expectedCaller:
+        - Holder Coordinator
+        - Issuer Coordinator
```
dlongley, line 27: A workflow service could also conceivably use this endpoint.

```suggestion
        - Issuer Coordinator
        - Workflow Service
```

holder.yml
```diff
@@ -107,7 +110,8 @@ paths:
        - oAuth2: []
        - zCap: []
       operationId: getPresentation
-      x-expectedCaller: "Holder Coordinator"
+      x-expectedCaller:
+        - Holder Coordinator
```
dlongley, line 115: A workflow service could also conceivably use this endpoint.

```suggestion
        - Holder Coordinator
        - Workflow Service
```

holder.yml
```diff
@@ -196,7 +203,8 @@ paths:
        - oAuth2: []
        - zCap: []
       operationId: createPresentation
-      x-expectedCaller: "Holder Coordinator"
+      x-expectedCaller:
+        - Holder Coordinator
```
dlongley, line 209: A workflow service could also conceivably use this endpoint.

```suggestion
        - Holder Coordinator
        - Workflow Service
```

issuer.yml
```diff
@@ -53,7 +55,8 @@ paths:
         - oAuth2: []
         - zCap: []
       operationId: updateCredentialStatus
-      x-expectedCaller: "Issuer Service"
+      x-expectedCaller:
+        - Issuer Service
```
dlongley: Hmm, this seems like it would be the issuer coordinator that would use this endpoint (and that the endpoint would reside on the Status Service).

```suggestion
        - Issuer Coordinator
```

dlongley (changes requested):

eric-schuh:
My only question is how we want to go about updating the tables themselves, the function in the respec-oas that creates the tables should be able to handle an array of endpoints fine, but we will also need to update the index.html with the new Workflow Service component, here:

dlongley (approved):

kezike:
@dlongley To be clear, I am not describing the discovery use case. I agree that coordinators should know the structure of templates. What I am saying is that it would help to verify that the credential prep records that the coordinator has created in the past still match against a valid template (by validating that it matches an existing template ID or schema in the workflow). This could be useful to ensure that changes in a workflow are captured in credential prep records.

issuer.yml
kezike, line 1: I was looking for the `GET /credentials/{id}` and `DELETE /credentials/{id}` endpoints in `issuer.yml` because the endpoints in the issuer service table [here](https://w3c-ccg.github.io/vc-api/#issuer-service) need to be updated to include workflow service as an expected caller. However, I found that the `/credentials/{id}` path only exists on `holder.yml`, which is presumably where these paths are being sourced for this table, per [this line](https://github.com/w3c-ccg/vc-api/blob/381a4e420e2b4a48671e13305b321c179c3b975d/index.html#L858) in `index.html`. So, it seems that we need a few things:

1. `GET /credentials/{id}` endpoint in `issuer.yml` with the workflow service added as an expected caller
2. `DELETE /credentials/{id}` endpoint in `issuer.yml` with the workflow service added as an expected caller
3. [OPTIONAL] Disambiguation between the `/credentials/{id}` definitions in `index.html`, such that the proper one is pulled in for different tables (issuer vs. holder)

kezike, line 1: I noticed that the API component section (currently [Section 3.7](https://w3c-ccg.github.io/vc-api/#api-component-overview)) is a bit out of sync with the API definition sections (currently Sections 3.8 to 3.12) with respect to the distinction between issuer responsibilities and status responsibilities. In the API component section, the `POST /credentials/status` endpoint is listed under the status service. Meanwhile, in the API definition sections, this endpoint is listed under the issuer service.

It feels to me like this endpoint should be listed in a new `status.yml` file that would map to a new API definition section for the status service, which would also bring the API component section and the API definition sections into alignment. I know we've discussed this a bit in the past, but I could not find an open issue for it. My questions:

1. Do we agree that there should be a distinction between issuer and status services in the API definition sections?
2. If yes, should I create an issue for this to avoid bloating the scope of this PR? If no, how should we address the misalignment between the API component section and the API definition sections?

cc @dlongley

eric-schuh: I would be in favor of doing the disambiguation option listed (and can do the work if the group agrees). Currently this definition lives one time in the holder.yml, but according to prior work the endpoint itself exists for both the Issuer Service and Holder Service. Adding the /credentials/{id} endpoint to the issuer.yml would also help with reading clarity of the OAS files since the Issuer Service is expected to have this endpoint, and would let us clean up the x-expectedCaller flags.

eric-schuh:
Keeping an eye on the following Google doc for any discussion about how to update these tables: https://docs.google.com/document/d/1D9PR28HmHsgAwPC2RBaiVwZ5Wt6AjYJY6SeD_M2oLVc/edit?tab=t.0#heading=h.v8s0fky2ksrq 

I got some feedback from Dave about a couple of the endpoints, and have updated the tables to reflect his suggestions. My plan is to put together a PR to reflect the state it settles by the end of this week for discussion next Tuesday (Sept 23)

Issue #530, opened 2025-09-09
"website" protocol is awkwardly named


msporny:
The "website" protocol in the specification is awkwardly named. The URL provided for the protocol is really a request for an invitation to a website. The object posted to the URL is an response to an invitation (which is the website that the sender would like the receiver to be redirected to). 

We should rename "website" to "inviteRequest" and adjust the language accordingly.

msporny:
PR #531 has been merged; closing.

PR #531, opened 2025-09-09
Update "website" interaction protocol to "inviteRequest".

msporny:
This PR is an attempt to address issue #530 by renaming the "website" protocol to the "inviteRequest" protocol.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vcalm/pull/531.html" title="Last updated on Sep 13, 2025, 6:28 PM UTC (cad03a7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vcalm/531/6b772ac...cad03a7.html" title="Last updated on Sep 13, 2025, 6:28 PM UTC (cad03a7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2296,7 +2296,7 @@ result in either of the following responses:
   }
 }
         </pre>
-        
+
         <pre class="example nohighlight" title="The degenerate case where the list of protocols only includes the inviteRequest option.">
 {
   "protocols": {
@@ -2304,7 +2304,7 @@ result in either of the following responses:
   }
 }
         </pre>
-              
+
         <p>
 The protocol response enables protocol execution to be delegated to third-party
 service providers through the HTTPS domain trust model, similar to the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2290,7 +2290,7 @@ <h4>Interaction Protocols Response</h4>
         <pre class="example nohighlight" title="A list of protocols supported for a given interaction">
 {
   "protocols": {
-    "website": "https://app.example/redirects/z8j3kfk2lQ",
+    "inviteRequest": "https://app.example/requests/z8j3kfk2lQ",
```
dlongley: I think this makes it more clear that you can reuse the same exchange for this, "vcapi", and "OID4VP":

```suggestion
    "inviteRequest": "https://saas.example/workflows/123/exchanges/987/invite-request/response",
```

index.html
```diff
@@ -2324,19 +2324,19 @@ <h4>Interaction Protocols Response</h4>
...
+specific interaction. If the `inviteRequest` interaction protocol is selected,
+the local system sends data using an HTTP POST to instruct the remote system
+where to send the individual. An example of the POST data is shown below:
         </p>
 
-        <pre class="example nohighlight" title="A website interaction protocol redirect request">
+        <pre class="example nohighlight" title="A response to an invitation request">
 {
-  "url": "https://website.example/interactions/8372974",
+  "url": "https://website.example/transactions/8372974",
```
dlongley: Hmm, I'd like to avoid reintroducing "transactions" anywhere in the spec to avoid confusion, though I realize this could really be any value for this particular case. How about "checkouts"?

```suggestion
  "url": "https://website.example/checkouts/8372974",
```

eric-schuh: I think "checkouts" gets a little to specific in terms of use cases this invitation-request pattern can be used for. Could we use "invitation"? As in the entity that goes to the /invite-request/response endpoint provides a URL with an /invitation/{invitationID}?

dlongley: I'd be ok with "invitation" -- I'd just be worried that people would think it then has to be "invitation", which is why having something different (or perhaps just two different examples?) would be helpful.

index.html
eric-schuh: Updated some language and added an additional example protocols object, while keeping the example of the above form in the following commit: https://github.com/w3c-ccg/vcalm/pull/531/commits/cfe5030d7c517af3ab7df54ca89aae241a359327

index.html
eric-schuh: Updated some language and added an additional example of the inviteResponse object in the following commit: https://github.com/w3c-ccg/vcalm/pull/531/commits/cfe5030d7c517af3ab7df54ca89aae241a359327

Issue #457, opened 2025-02-19
Underspecification of format for workflow `credentialTemplates` property


kezike:
I noticed that the `credentialTemplates` property in `workflow.steps[STEP_NAME].credentialTemplates` found in the [workflow creation](https://w3c-ccg.github.io/vc-api/#create-workflow) request and response as well as the [workflow retrieval](https://w3c-ccg.github.io/vc-api/#get-workflow-configuration) response is specified as an array, but it does not specify the format of each item in the specification text (although the OpenAPI definition does specify that the fields should be `type` and `template`). Additionally, since this is an array at the workflow level without IDs, it is unclear how the templates should be applied to an exchange/step. Should this property reside at the exchange level instead? If not, could we at least add an ID property to the items and/or modify the format of `credentialTemplates` to a map with keys being the step names?

dlongley:
Each object in `credentialTemplates` should have a schema like this:

```
{
  title: 'Typed Template',
  type: 'object',
  required: ['type', 'template'],
  additionalProperties: false,
  properties: {
    id: {
      type: 'string'
    },
    type: {
      type: 'string',
      // with other options possible too
      enum: ['jsonata']
    },
    template: {
      type: 'string'
    }
  }
}
```

The `id` property is optional but if present can be used to reference each template in an exchange for issuance purposes. Alternatively the index of the template can be used. Each exchange `step` can define an array of `issueRequests` where each element is an object that references one of these credential templates; the objects have a schema like this:

```
{
  title: 'Issue Request Parameters',
  type: 'object',
  oneOf: [{
    required: ['credentialTemplateId']
  }, {
    required: ['credentialTemplateIndex']
  }],
  additionalProperties: false,
  properties: {
    credentialTemplateId: {
      type: 'string'
    },
    credentialTemplateIndex: {
      type: 'number'
    },
    // optionally specify different variables from `exchange.variables`
    variables: {
      oneOf: [{type: 'string'}, {type: 'object'}]
    }
  }
}
```

kezike:
OK, according to your response, there are two missing items:

1. `credentialTemplates[].id` in both the specification text and the OpenAPI definition
2.  `workflow.steps[STEP_NAME].issueRequests` in both the specification text and the OpenAPI definition

[...]

Issue #446, opened 2025-02-17
What is the schema for exchange `variables`?


kezike:
What is the schema for exchange `variables`? I have seen examples where the keys are the step name (e.g., `variables[exchange.step].results.verifiablePresentation`) and I have seen other examples where the keys are something else (e.g., `results`) and the values are step-scoped data (e.g., `variables.results[exchange.step].verifiablePresentation`).

dlongley:
We should clarify what pieces of this should be considered standard (i.e., every implementation must implement in a specific way). At present, the only restriction on `variables` is for it to be an object, but given that implementations have used it to also store step results to allow template access to the result of any other step, we should probably reserve the `results` key and define it in a particular way for interoperability.

If we do that, we would say that `results` is reserved and, if used, it is an object with step names as keys and the exchange server sets the result of each step (e.g., `{verifiablePresentation, ...}`) as the value. The other properties could be left up to implementations to use as desired.

Another option would be to store results in another place and allow implementations to merge it into variables for template processing, but this just seems like it would be de facto reserving *some property* that would be overwritten (probably `results`) in the variables object to enable implementations to do this anyway.

We could also reserve some other top-level property in variables to allow for future expansion of the spec w/any other implementation-common features. We might consider whether to move `results` underneath that property, considering backwards compatibility concerns with existing implementations.

kezike:
Yes, that would be good to clarify, as well as errors, which I have seen stored in `variables` under the scope of a step as `error`/`errors`. Perhaps, that also belongs under `results`? Don't have a strong opinion on this at the moment, but just wanted to call it out as another field I have seen used in `variables`.

msporny:
The group talked about this on the 2025-02-18 telecon. @dlongley noted that we might want to reserve some key names underneath the variables object. The current YAML states the following:

```
variables:
  type: array
  description: Template variables to be used in the exchange.
  items:
    type: object
    properties:
      type:
        type: string
        description: The type of template.
      template:
        type: string
        description: The template itself.
```

... which is wrong. The `variables.type` should be an `object`, not an `array`. Each item is wrong as well. It should be something closer to the following (no strong schema/typing necessary):

```
variables:
  type: object
  description: Template variables to be used in the exchange.
```

PR #549, opened 2025-09-10
Define reserved exchange variables.

kezike:
This PR attempts to define reserved exchange variables:

- `inputs`: single or multiple credential inputs for each exchange step
- `results`: result Verifiable Presentations for each exchange step
- `errors`: errors for each exchange step

Open to all feedback!


kezike:
NOTE - For some reason, the autogenerated preview linked above indicates that the tables are no longer rendering (despite a proper rendering locally). I will address that soon.

exchanges.yml
```diff
@@ -470,3 +468,50 @@ components:
...
+    ExchangeVariables:
+      type: object
+      description: Variables needed for an exchange.
+      properties:
+        inputs:
+          type: object
+          description: Credential inputs for each step of an exchange.
+          properties:
+            STEP_NAME:
+              $ref: '#/components/schemas/ExchangeStepInputs'
```
dlongley: I'm not so sure about this one -- since it may cause implementers to think they can't share variables across steps. Indeed, even the `results` variables can be used in steps. Do we need a specific "inputs" field here? Every variable is available to every step template.

exchanges.yml
```diff
@@ -470,3 +468,50 @@ components:
...
+          properties:
+            STEP_NAME:
+              $ref: '#/components/schemas/ExchangeStepResults'
+        errors:
+          type: object
+          description: Errors from each step of an exchange.
+          nullable: true
+          properties:
+            STEP_NAME:
+              $ref: '#/components/schemas/ExchangeStepErrors'
```
dlongley: Should `errors` be variables? Would they ever appear or be used in steps? Is the main use case here for exchange creators to be able to read the errors that occurred? If so -- perhaps this should instead be an optional property of an exchange.

In our implementation, we just implement `lastError` as an optional property on an exchange and it's just for the most recent step -- as no other steps can happen when there's an error anyway. Perhaps that's worth documenting instead of this.

Notably, this would be different from `results` in variables. While the presence of `results` allows the exchange creator to read results, that's really just a side effect of the driving purpose for having them be in "variables". Results are often expected to used in later step templates (e.g., if the client provided a VC with values X and Y in one step, those might go into another credential that is being issued -- or into a schema for another request in a subsequent step). There's no optional `results` property on an exchange since it would then be redundant with the results already being in "variables" -- and there are storage size constraints to consider.

dlongley (changes requested):

msporny:
@kezike I spent a decent chunk of today refactoring the entire repository to merge all OAS definitions into a single `oas.yaml` file. The merge went well and the structure of the repo is so much better as a result. Unfortunately, that also means that you'll have to rebase and move your edits over to `oas.yaml`. Looking at the PR, it shouldn't take you that long to migrate. Apologies for the timing, I wanted to get this done before the next meeting and was able to merge everything except for your PR before doing it.

TallTed (approved):
lgtm

exchanges.yml
kezike: The `inputs` field, as I have conceived it, is the home for the credentials that would be issued by that step. This would have the same format as the `workflow.step.issueRequests` field [defined here](https://github.com/w3c-ccg/vcalm/issues/457) (hence why `inputs` can be an object or an array, though I am open to making it an array for simplicity). Meanwhile, I see `results` as the home of data that is sent from a holder to the service during an exchange participation request (e.g., a VP, with holder, challenge, domain, etc.). Without using a designated field like `inputs` for credential data, I am curious to know how you are providing the credential data for each step.

kezike: Also, it is worth noting that my main reason for reopening Issue #446 - and this PR addressing it - is because I felt like we needed a field like `inputs`/`credentials`/`whatever` to hold the credential data, since the issuer coordinator would need to know which variable field to place credential data in the exchange records. Please let me know if I am missing something in my understanding.

exchanges.yml
kezike: Re `errors` variable: I am least opinionated on this variable (in fact, I almost omitted it entirely). Your read is correct in that it is intended to give feedback to exchange creators (e.g., issue coordinators), if they want it. Your point on an optional `exchange.lastError` field is well taken. Are you proposing that we add that to the OAS or should we omit it altogether and allow implementations to handle that as they please?

Re `results` variable: Have we defined a standard mechanism to perform the data population procedure you described in this statement (this seems like an important thing to specify):

> Results are often expected to used in later step templates (e.g., if the client provided a VC with values X and Y in one step, those might go into another credential that is being issued -- or into a schema for another request in a subsequent step)

exchanges.yml
dlongley: > The `inputs` field, as I have conceived it, is the home for the credentials that would be issued by that step. This would have the same format as the `workflow.step.issueRequests` field [defined here](https://github.com/w3c-ccg/vcalm/issues/457) (hence why `inputs` can be an object or an array, though I am open to making it an array for simplicity).

Hmm, I found this confusing because `issueRequests` already has a `variables` field where the variables for a specific credential template for the step can be specified either by reference (using a string pointing to any other variable in the exchange) or directly (using an object with whatever you want). And this is already nested / co-located with the specific issue request -- so I'm still not quite following what `inputs` does differently from this.

Rather, it seems that it is actually one way to make use of this feature, but not necessarily what people may prefer. If I'm following, it seems to require one to adopt a particular pattern to use it (see below), which may work well for some use cases and not others.

> Without using a designated field like `inputs` for credential data, I am curious to know how you are providing the credential data for each step.

The step template can use whatever variables it wants to, and it could be designed like this (for example):

```
{
  ...,
  "issueRequests": [{
    "credentialTemplateId": "urn:employee-credential-template",
    "variables": employeeCredentialVars
  }, ...]
}
```

You could use `inputs`, if you'd like, with this pattern:

```
{
  ...,
  "issueRequests": [{
    "credentialTemplateId": "urn:credential-template-1",
    "variables": inputs.stepName
  }, ...]
}
```

But I find that a little strange because, as you say, you'd expect to need the `inputs.stepName` to be an array to match this pattern against multiple requests, or you'd need some other kind of secondary mapping under `inputs.stepName` to identify which issue request to map to. It implies that the coordinator knows more about the specific step template's structure and order than it really needs to, whereas the semantic variable name `employeeCredentialVars` doesn't have that restriction.

I think people may prefer to communicate a set of variables for a coordinator to send like:

```
{
  "employeeCredentialVars": {
    "name": "Jane Doe",
    ...
  },
  "membershipCredentialVars": {
    "memberId": "1234",
    ...
  },
  "skillCredentialVars": {
    "skill": "Windsurfing",
    ...
  }
}
```

Notably, the above could be rearranged a number of different ways, but none of them would require the coordinator creating the exchange to know about step or issue request order in the step template, etc.

I think this is better than instead asking them to send:

```
{
  "inputs": {
    "myStep1": [{
      "name": "Jane Doe",
      ...
    }, {
      "memberId": "1234",
      ...
    }],
    "myStep2": {
      "skill": "Windsurfing",
      ...
    }
  }
}
```

I might be misunderstanding, but it seems like this is what the `inputs` proposal would look like for, say, a two step process that issues two VCs in the first step and one in the second.

Therefore, it seems to me that `inputs` is a bit of a leaky abstraction and the name `inputs` itself doesn't make it clear what the inputs are for by the name (hence my original confusion around the fact that all variables are "inputs"). Of course, you've explained it now that the intention is for `inputs` to be for specific issue requests within a given step, but what about "inputs" for other things -- like protocol parameters, challenges, VPRs, etc.?

exchanges.yml
dlongley: I would be ok with suggesting and optional `lastError` property of the exchange that can be used to express error information. I think we'll then need to say how to express its value as well, making that more strict too. The group can decide to what level we want to express these things.

exchanges.yml
```diff
@@ -0,0 +1,517 @@
...
+        openId:
+          type: object
+          description: Optional information to trigger OID4VP
+          properties:
+            createAuthorizationRequest:
+              type: string
+              description: If present, identifies the name of the template variable (in <code>variables</code> for an <a href="#create-exchange">exchange</a>) that the auto-generated OpenID for Verifiable Presentations Authorization Request will be stored in for subsequent use in the exchange. Either this or <code>authorizationRequest</code> MUST be present, but not both. The value of the named variable can be lazily generated when the Authorization Request is required, i.e., after the client of the exchange chooses to use the OID4VP protocol. This feature allows the Authorization Request to be generated by reusing values, e.g., a Verifiable Presentation Request, that only need to be specified once across multiple protocol offerings.
+            authorizationRequest:
+              type: object
+              description: If present, an OID4VP Authorization Request for subsequent use in the exchange. Either this or <code>createAuthorizationRequest</code> MUST be present, but not both.
```
TallTed: Why sometimes "OID4VP" and sometimes "OpenID for Verifiable Presentations"? Better to be consistent, which I would suggest be either "OID4VP (OpenID for Verifiable Presentations)" or just "OID4VP" if there's a guaranteed easily-accessible place that the reader can see the expansion.

exchanges.yml
```diff
@@ -0,0 +1,517 @@
...
+        nextStep:
+          type: string
+          description: The name of the next step in the sequence. Passing nextStep is OPTIONAL. This field MUST NOT be present on the final step configuration.
+        openId:
+          type: object
+          description: Optional information to trigger OID4VP
+          properties:
+            createAuthorizationRequest:
+              type: string
+              description: If present, identifies the name of the template variable (in <code>variables</code> for an <a href="#create-exchange">exchange</a>) that the auto-generated OpenID for Verifiable Presentations Authorization Request will be stored in for subsequent use in the exchange. Either this or <code>authorizationRequest</code> MUST be present, but not both. The value of the named variable can be lazily generated when the Authorization Request is required, i.e., after the client of the exchange chooses to use the OID4VP protocol. This feature allows the Authorization Request to be generated by reusing values, e.g., a Verifiable Presentation Request, that only need to be specified once across multiple protocol offerings.
```
TallTed: Will all tools working with this file know that `description` will contain `HTML` and present it properly?

exchanges.yml
kezike: OK, that's fine with me. As for the results, since this variable is dynamically populated, it feels to me that there would still need to be either a predefined set of fields that can be placed in it for all exchanges or some other config-based way to convey what values should be placed in results. At the moment, it is not clear how this should be done. To be more concrete, at the moment, I am dynamically placing challenge and domain in a step's results from a DID Authentication steps like this:

```
{
  "results": {
    "STEP_NAME": {
      "verifiablePresentation": {
        "challenge": "35634c13-1737-4107-b07c-39ddfb454f0d",
        "domain": "https://vcalm.example.com"
      }
    }
  }
}
```

It is not clear to me how to specify arbitrary result values for any type of step, according to your claim above:

> Results are often expected to used in later step templates (e.g., if the client provided a VC with values X and Y in one step, those might go into another credential that is being issued -- or into a schema for another request in a subsequent step)

exchanges.yml
kezike: Not sure, but it was there prior to this PR and seems to be working fine with the tools we are currently using:

Screenshot 2025-09-21 at 3 15 08 AM

index.html
```diff
@@ -1720,9 +1720,11 @@ <h3>Workflows and Exchanges</h3>
...
 pending, active, or complete, as well as the current step in the workflow, any
 workflow-specific variables and data, and any verifiable presentations and
-credentials received while the exchange executes. While there is no technical
-limitation on the number of steps in a workflow, implementers might want to use
-a default maximum number of steps to prevent bugs.
+credentials received while the exchange executes. Implementers are free to use
+whatever variables they need for their implementation in addition to the
+reserved variables defined in Section [[[#create-exchange]]]. While there is no
+technical limitation on the number of steps in a workflow, implementers might
+want to use a default maximum number of steps to prevent bugs.
```
```suggestion
want to enforce a default maximum number of steps to prevent bugs.
```

oas.yaml
```diff
@@ -1031,10 +1031,36 @@ components:
...
+    ExchangeError:
+      type: object
+      description: An error encountered throughout the course of an exchange.
+      properties:
+        message:
+          type: string
+          description: A message describing an exchange error.
+        status:
+          type: integer
+          description: A status number describing an exchange error.
```
dlongley: We should converge on an error style and use it consistently. We're using this elsewhere:

https://github.com/w3c-ccg/vcalm/blob/main/components/ProblemDetails.yml#L15-L31

oas.yaml
```diff
@@ -1031,10 +1031,36 @@ components:
...
+          description: Results from each step of an exchange.
+          properties:
+            STEP_NAME:
+              $ref: '#/components/schemas/ExchangeStepResults'
+    ExchangeStepResults:
+      type: object
+      description: Results from a step of an exchange.
+      properties:
+        verifiablePresentation:
+          $ref: "./components/VerifiablePresentation.yml#/components/schemas/VerifiablePresentation"
```
dlongley, line 1052: Notably, if `OID4VCI` or `OID4VP` (or some other protocol) is used, some other information would be captured. So we could list other fields like this:

```
// populated if OID4VCI/OID4VP was used
openId: {
  clientProfileId,
  authorizationRequest,
  presentationSubmission
},
// populated if `inviteRequest` was used
inviteRequest: {inviteResponse}
```

index.html
```diff
@@ -1720,9 +1720,11 @@ <h3>Workflows and Exchanges</h3>
 stores state information about each exchange, such as whether the exchange is
 pending, active, or complete, as well as the current step in the workflow, any
 workflow-specific variables and data, and any verifiable presentations and
-credentials received while the exchange executes. While there is no technical
-limitation on the number of steps in a workflow, implementers might want to use
-a default maximum number of steps to prevent bugs.
+credentials received while the exchange executes. Implementers are free to use
+whatever variables they need for their implementation in addition to the
```
```suggestion
whatever variables they need or wish for their implementation in addition to the
```

exchanges.yml
dlongley:
```
{
  template: {
    "foo": {
      "value": results.STEP_NAME.verifiablePresentation.verifiableCredential[0].credentialSubject.foo
    }
  }
}```

Comment (kezike) [reply exchanges.yml]: I have prepared [this gist](https://gist.github.com/kezike/fa58b7acbb98cba6f6aeb853188695a3) to provide example workflow configurations that can be included in the spec. You will notice that the second example attempts to use `stepTemplate` in a DID Authentication step configuration, per your recommendation. However, it is incomplete because it is still unclear how this can be used to direct the appropriate values into `variables.results`. DID Authentication seemed like a great example to use because it is a common step definition and it requires the use of dynamically populated values.

File: oas.yaml (hunk 6 above)
Comment (kezike) [reply oas.yaml]: @dlongley Between the `ProblemDetails` schema you've suggested here and and the [ErrorResponse schema](https://github.com/w3c-ccg/vcalm/blob/main/components/ErrorResponse.yml#L15-L36), which would you say is more appropriate for this use case?

File: index.html (hunk 8 above)
Comment (kezike) [reply index.html]: @TallTed Can we capture "need" and "wish" with "desire" instead?

File: oas.yaml (hunk 7 above)
Comment (kezike) [reply oas.yaml:1052]: @dlongley If we add these new variables, would we also need to specify the detailed schema for all of the fields (e.g., `openId.authorizationRequest.response_type`, `open.authorizationRequest.presentation_definition`, etc.) or would the `object`/`string` types suffice along with an explanatory note in the description?

File: index.html [hunk 9]
```diff
@@ -1891,7 +1893,180 @@ <h4>Create Workflow</h4>
...
 
         <p>
-For the `openId` options above, specifically the values
+Most workflows will contain information about steps, credential templates,
+and authorization. In some cases, a workflow will also reference the
+reserved `results` variable at different points in its configuration
+(typically in credential templates). This variable holds data that
+the workflow service receives from clients throughout the course
+of an exchange. Here is a basic workflow configuration that contains
+a DID Authentication step and a credential delivery step:
```
kezike: @dlongley Do you think this is a decent enough explanation/motivation for the `results` variable given our discussions in this PR and elsewhere?

index.html
```diff
@@ -1891,7 +1893,180 @@ <h4>Create Workflow</h4>
...
+        </pre>
+
+        <p>
+Some workflows will need to reference custom variables in one or more
+of their steps. In this case, the coordinator will need
+to provide the appropriate value(s) for the variable(s) when it
+creates an exchange. Here is an example of a workflow configuration
+with a templated DID Authentication step
+(Notice the `verifiablePresentationRequest` and `callbackUrl` variables
+referenced in the `didAuthRequest` step):
```
kezike, line 1997: @dlongley Do you think this is a decent enough explanation/motivation for custom variables given our discussions in this PR and elsewhere?

exchanges.yml
kezike: To close the loop on this thread, the examples in that gist have been updated and included in this PR based on feedback from the last call.

exchanges.yml
kezike: To close the loop on this thread, we agreed a few calls ago that a reserved `inputs` variable is unnecessary, since the coordinator can place the credential variables directly at the root scope of `exchange.variables`.

index.html
```diff
@@ -1891,7 +1893,180 @@ <h4>Create Workflow</h4>
           data-api-endpoint="post /workflows"></div>
 
         <p>
-For the `openId` options above, specifically the values
+Most workflows will contain information about steps, credential templates,
+and authorization. In some cases, a workflow will also reference the
```
dlongley: I wouldn't say we know what most workflows will or won't have, but we can definitely call out what they can contain:

```suggestion
Workflows can contain information about steps, credential templates,
and authorization. In some cases, a workflow will also reference the
```

index.html
dlongley: I left a minor edit, otherwise, yes, thanks!

index.html
```diff
@@ -1891,7 +1893,180 @@ <h4>Create Workflow</h4>
...
+          "achievement": {
+            "id": sampleAchievementCredential.credentialSubject.achievement.id,
+            "type": ["Achievement"],
+            "achievementType": sampleAchievementCredential.credentialSubject.achievement.achievementType,
+            "name": sampleAchievementCredential.credentialSubject.achievement.name,
+            "description": sampleAchievementCredential.credentialSubject.achievement.description,
+            "alignment": sampleAchievementCredential.credentialSubject.achievement.alignment
+          }
+        }
+      }"
```
dlongley, line 1983: I think this template should be changed to be nested under a top-level `"credential"` property. This will also not contradict the ability for a template to output `"options"` if desired along side it.

While it's true that an implementation could try and guess whether the template renders directly to a credential or renders to the payload to be sent to the VC API `/credentials/issue` endpoint, we should have all the examples in the spec avoid doing that but being clear that the output rendered by the template should be `{credential, options?}`.

index.html
```diff
@@ -1891,7 +1893,180 @@ <h4>Create Workflow</h4>
...
+        }
+      ]
+    }
+  },
+  "credentialTemplates": [
+    {
+      "id": "caffb919-053a-4bfa-beba-80567904f842",
+      "type": "jsonata",
+      "template": "{
+        "@context": [
```
dlongley: Same comment as above.

oas.yaml
```diff
@@ -635,7 +635,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/ErrorResponse"
+                $ref: "./components/ProblemDetails.yml#/components/schemas/ProblemDetails"
```
kezike, line 638: @dlongley As I was making this change I noticed that this `/exchanges/{exchangeId}` endpoint path definition seems out of place (similar to what we discussed last call when @eric-schuh asked about it in the component table. Should I remove this and the `/endpoint` endpoint path definition above (Note: the description of the `GET /exchanges` definition indicates that it would be used for exchange discovery, which is not specified in the spec for `GET /workflows/{localWorkflowId}/exchanges/{localExchangeId}`), and which seems like odd language for an admin endpoint and for ephemeral resources like an exchange?

PR #550, opened 2025-09-16
Added yaml to support the interaction mechanism endpoints

eric-schuh:
Add two endpoints, along with their required components. These are:

- /interactions/{interactionId}
- /{inviteId}/invite-request/response

A couple of things I would like double checked due to my unfamiliarity with yaml and OAS:

1. In the component table, for the "Protocols" object, is the "otherFutureAPI" property appropriate? I believe we aren't planning on specifying a supported list of protocols, so I was trying to show that the object is extendable.
2. In the summary of the {inviteId}/invite-request/response, I tried to capture a note from @dlongley that in the case someone implements the entire VCALM we would expect inviteID to be an exchangeId. Is this the right spot for this kind of message or should it be moved elsewhere?

### DIFF
--- a/oas.yaml
+++ b/oas.yaml
@@ -1232,7 +1232,11 @@ components:
       type: object
       description: Object containing the available options for continuing the current interaction.
       required:
-        oneOf: ["inviteRequest", "vcaml"]
+        oneOf: 
+          - type: string
+              description: The value MUST be <code>inviteRequest</code>.
+          - type: string
+              description: The value MUST be <code>vcaml</code>.
       properties:
         inviteRequest:
           type: string
@@ -1257,9 +1261,7 @@ components:
         url:
           type: string
           description: A URL where the recipient can continue the interaction, should they choose.
-          examples:
-            - https://website.example/checkout/8372974
-            - https://website.example/forms/7864165 
+          example: https://website.example/checkout/8372974
         purpose:
           type: string
           description: A reason for the interaction that is human readable and usable in a concent interaction with the user recieving the inviteResponse object. 
--- a/oas.yaml
+++ b/oas.yaml
@@ -772,6 +772,7 @@ paths:
           required: true
           schema:
             type: string
+            enum: ["1"]
           description: Informs a reader of the Interaction URL of the Interaction API version, and allows for quick disambiguation of the URL when scanned from a QR Code.
       requestBody:
         description: A request for the protocols object to negotiate an interaction protocol. 
--- a/oas.yaml
+++ b/oas.yaml
@@ -1238,7 +1238,7 @@ components:
           example: https://website.example/invitation/87654321/invite-request/response
         vcaml:
           type: string
-          description: A VC-API exchanges URL that will allow the initiating user of the current interaction to continue with a VC-API credential exchange.
+          description: A VC-API exchange URL that will allow the initiating user of the current interaction to continue with a VC-API credential exchange.
           example: https://vcapi.example/workflows/z1A1FjMfnG/exchanges/z19mxakBFecZ"
         otherFutureAPI:
           type: string
--- a/oas.yaml
+++ b/oas.yaml
@@ -1233,23 +1233,25 @@ components:
       description: Object containing the available options for continuing the current interaction.
       required:
         oneOf: 
-          - type: string
-              description: The value MUST be <code>inviteRequest</code>.
-          - type: string
-              description: The value MUST be <code>vcaml</code>.
+          - $ref: "#/componenets/schemas/inviteRequest"
+          - $ref: "#/componenets/schemas/vcalm"
       properties:
         inviteRequest:
-          type: string
-          description: A URL that is intended to take the initiating user of the current interaction to a website to continue the interaction beyond this API.
-          example: https://website.example/invitation/87654321/invite-request/response
-        vcapi:
-          type: string
-          description: A VC-API exchanges URL that will allow the initiating user of the current interaction to continue with a VC-API credential exchange.
-          example: https://vcapi.example/workflows/z1A1FjMfnG/exchanges/z19mxakBFecZ
+          - $ref: "#/componenets/schemas/inviteRequest"
+        vcaml:
+          - $ref: "#/componenets/schemas/vcalm"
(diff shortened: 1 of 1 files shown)


oas.yaml
```diff
@@ -744,6 +744,70 @@ paths:
...
+                $ref: "#/components/schemas/Protocols"
+        "400":
+          description: Invalid or malformed input
+        "401":
+          description: Not Authorized
+        "500":
+          description: Internal Error
+  /{inviteId}/invite-request/response:
+    post:
+      summary: Provides a location for a particular invitation response to be submitted to the by the entitiy trying to start an interaction. In the case that an implimenter has implimented the full VCALM, it is expected that the {inviteId} parameter will be equivilant to the {exchangeId}.
```
TallTed:
```suggestion
      summary: Provides a location for a particular invitation response to be submitted to the by the entity trying to start an interaction. When an implementer has implemented the full VCALM, it is expected that the {inviteId} parameter will be equivalent to the {exchangeId}.
```

oas.yaml
```diff
@@ -1231,32 +1231,24 @@ components:
...
-          schema:
-            - $ref: "#/components/schemas/inviteRequest"
+          type: string
+          description: A URL that is intended to take the initiating user of the current interaction to a website to continue the interaction beyond this API.
+          example: https://website.example/invitation/87654321/invite-request/response
         vcaml:
-          schema:
-            - $ref: "#/components/schemas/vcalm"
+          type: string
+          description: A VC-API exchanges URL that will allow the initiating user of the current interaction to continue with a VC-API credential exchange.
```
TallTed:
```suggestion
          description: A VC-API exchange URL that will allow the initiating user of the current interaction to continue with a VC-API credential exchange.
```

oas.yaml
```diff
@@ -744,6 +744,88 @@ paths:
...
+          schema:
+            oneOf: 
+              - type: string
+              - type: integer
+          description: The unique interaction identifier for the current interaction. 
+        - in: query
+          name: iuv
+          required: true
+          schema:
+            type: string
```
ottonomy, line 774: Add `enum: ["1"]` to enforce the specific value required.

eric-schuh: As far as the requirement to use the /interaction/{interactionId} endpoint is concerned. I don't think it should be required, since as you pointed out the iuv query pram will indicate a URL is supposed to be an interaction URL. 

Would it be enough if I updated the description and summary of this endpoint to make it clear that implementers can choose a different endpoint if they choose, as long as the iuv parameter is included?

eric-schuh: I will add `enum: ["1"]` to the iuv parameter block as well.

eric-schuh: The yaml is updated per this suggestion.

oas.yaml
```diff
@@ -744,6 +744,89 @@ paths:
...
+                $ref: "#/components/schemas/Protocols"
+        "400":
+          description: Invalid or malformed input
+        "401":
+          description: Not Authorized
+        "500":
+          description: Internal Error
+  /{inviteId}/invite-request/response:
+    post:
+      summary: Provides a location for a particular invitation response to be submitted to the by the entity trying to start an interaction. When an implementer has implemented the full VCALM, it is expected that the {inviteId} parameter will be equivalent to the {exchangeId}.
```
msporny:
```suggestion
      summary: Provides a location for a particular invitation response to be submitted to the by the entity trying to start an interaction. When an implementer has implemented the full specification, it is expected that the {inviteId} parameter will be equivalent to the {exchangeId}.
```

oas.yaml
```diff
@@ -744,6 +744,89 @@ paths:
...
+      x-expectedCaller: Whoever POSTed to the URL containing /interaction/{interactionId}
+      parameters:
+        - in: path
+          name: inviteId
+          required: true
+          schema:
+            oneOf: 
+              - type: string
+              - type: integer
+          description: The unique inviteId that the interaction requester uses to select the invite response protocol from the set of advertised options. When an implementer has implemented the full VCALM, this value will be an {exchangeId}, and that the invite-request/response endpoint lives on the Workflow Service.
```
msporny:
```suggestion
          description: The unique inviteId that the interaction requester uses to select the invite response protocol from the set of advertised options. When an implementer has implemented the full specification, this value will be an {exchangeId}, and that the invite-request/response endpoint lives on the Workflow Service.
```

oas.yaml
```diff
@@ -1146,3 +1229,44 @@ components:
...
+      description: Object containing the available options for continuing the current interaction.
+      properties:
+        inviteRequest:
+          type: string
+          description: A URL that is intended to take the initiating user of the current interaction to a website to continue the interaction beyond this API.
+          example: https://website.example/invitation/87654321/invite-request/response
+        vcaml:
+          type: string
+          description: A VC-API exchange URL that will allow the initiating user of the current interaction to continue with a VC-API credential exchange.
+          example: https://vcapi.example/workflows/z1A1FjMfnG/exchanges/z19mxakBFecZ"
```
msporny:
```suggestion
          example: https://vcapi.example/workflows/z1A1FjMfnG/exchanges/z19mxakBFecZ
```

oas.yaml
```diff
@@ -1146,3 +1229,44 @@ components:
...
+        otherFutureAPI:
+          type: string
+          description: A URL that is intended to take the initiating user of the current interaction to some other, as yet to be defined, API to continue the interaction and credential exchange.
+          example: https://otherFutureAPI.example/otherFutureAPIParameter/0987
```
msporny:
```suggestion
          example: https://otherFutureAPI.example/collection/0987
```

oas.yaml
```diff
@@ -1146,3 +1229,44 @@ components:
...
+      oneOf:
+        - required:
+          - inviteRequest
+        - required:
+          - vcalm
+    InviteResponse:
+      type: object
+      description: Object a continuation URL for an interaction, along with a purpose for a given interaction.
```
msporny:
```suggestion
      description: Object for a continuation URL for an interaction, along with a purpose for a given interaction.
```

msporny (approved):

Issue #558, opened 2025-09-30
Cross-component Workflow Services


eric-schuh:
For diagram simplicity reasons, it makes sense that our [API Components diagram](https://w3c-ccg.github.io/vcalm/#components) shows each major component having its own workflow service. The current language that describes the workflow service also implies that each major component has its own workflow service, it currently reads:

"A workflow service provides a way for coordinators to automate specific interactions for specific users. Each role ([holder](https://www.w3.org/TR/vc-data-model-2.0/#dfn-holders), [issuer](https://www.w3.org/TR/vc-data-model-2.0/#dfn-issuers), and [verifier](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifier)) can run their own workflow service to create and manage exchanges that realize particular workflows. Administrators configure the workflow system to support particular flows. Then, when the business rules justify it, coordinators create exchanges at their workflow service and give access to those exchanges to any authorized party."

Unless I have a wrong notion, I believe it is expected that the workflow services will often need to support workflows that involve both verification and issuance as part of the same workflow. In these cases the "workflow service" should be interfacing with potentially all three coordinator components, not just a single one. 

Assuming the above is correct, and I'm not misunderstanding something, I'd like to update the workflow service description to make it clear that workflow services, and workflows can interface with multiple coordinators and are not restricted to being an "issuer workflow service", "verifier workflow service", or "holder workflow service" which seems to be implied in both our architecture diagram and language currently.

eric-schuh:
Recording Ted's statement for the PR: "One software package may serve multiple roles, but these roles are not required to be served by a single library or other software package."

PR #559, opened 2025-10-07
Update API Component Tables

eric-schuh:
Since [PR#525](https://github.com/w3c-ccg/vcalm/pull/525) is quite old and still has the old yaml file construction I felt it was easier to create a new PR for this update.

The respec-oas repo will also need to have the following PR accepted for this update to work: https://github.com/digitalbazaar/respec-oas/pull/1

The Component Tables have been updated to match the current state of the google doc, here: https://docs.google.com/document/d/1D9PR28HmHsgAwPC2RBaiVwZ5Wt6AjYJY6SeD_M2oLVc/edit?tab=t.0

While I was going through the oas.yaml to make sure the x-expectedCaller endpoints were correct I found 3 endpoints that were missing the flag entirely, and were not in the Google Doc linked above. These are currently all listed as part of the Workflow Service table:

GET /workflows/{localWorkflowId}/exchanges/{localExchangeId}/protocols
GET /exchanges
POST /exchanges/{exchange-id}

My question for the group before we accept this PR is: Is this the correct location for these, and are the listed expectedCallers correct?

***
<a href="https://pr-preview.s3.amazonaws.com/eric-schuh/vcalm/pull/559.html" title="Last updated on Oct 26, 2025, 5:26 PM UTC (53b4b7b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vcalm/559/7c31416...eric-schuh:53b4b7b.html" title="Last updated on Oct 26, 2025, 5:26 PM UTC (53b4b7b)">Diff</a>

PR #562, opened 2025-10-14
Update Architecture section to include a statement about implementing multiple components

eric-schuh:
This PR is attempting to address https://github.com/w3c-ccg/vcalm/issues/558 by adding a paragraph at the start of the Architecture section that details why the architecture described was selected, and that any implement is free to implement any combination of components as is appropriate for their use case.

***
<a href="https://pr-preview.s3.amazonaws.com/eric-schuh/vcalm/pull/562.html" title="Last updated on Oct 26, 2025, 5:33 PM UTC (5606aef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vcalm/562/7c31416...eric-schuh:5606aef.html" title="Last updated on Oct 26, 2025, 5:33 PM UTC (5606aef)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -515,14 +515,19 @@ All implementations MAY provide functionality beyond this specification.
   <section>
     <h2>Architecture</h2>
       <p>
-The below architecture description, as well as that seen in the diagram in the 
+The architecture description below, as well as that seen in the diagram in the 
 <a href="#architecture-overview"></a> section, are representative of one possible 
-architecture for these software components. One software package may serve multiple 
-roles, but these roles are not required to be served by a single library or other 
-software package. For the purposes of this specification, each component will be 
-treated independently, both for clarity of reading, and to highlight that each 
-component can be implemented separately from any other, assuming some other software 
-is fulfilling the missing roles. 
+architecture for these software components. This representation was chosen to
+make categorization of the different features in the specification clear, enabling
+each component to be treated independently for both clarity of reading and
+to highlight that each component can be implemented separately from any
+other. However, this is not a prescriptive architecture; implementers can use as
+many or as few software packages as desired to serve multiple roles; roles are
+not required to be served by a single library or other software package nor are
+software libraries or packages required to provide for particular roles in isolation
+of others. It is also an expectation that single deployments of software or
+system components can be for the purpose of multiple roles as needed for a
+given use case.
       </p>
     <section>
       <h3>Coordinators</h3>


components/parameters/path/ExchangeId.yml
```diff
@@ -1,4 +1,4 @@
-name: exchange-id
+name: exchangeId
```
dlongley, line 1: This change seems unrelated and I believe is going into another PR, so the branch for this PR might need a rebase on that other one to prevent this from appearing in the diff.

index.html
```diff
@@ -901,52 +909,71 @@ <h3>Component Overview</h3>
...
+along with the component that is expected to call the endpoint.
       </p>
       <table class="simple api-component-table"
         data-api-path="/credentials/verify /presentations/verify /challenges"></table>
 
-      <h4>Holder Service</h4>
+      <h4>Holder Coordinator</h4>
+      <p>
+Below are all endpoints expected to be exposed by the [=holder coordinator=], along with
+the component that is expected to call the endpoint.
```
dlongley, line 954:
```suggestion
All endpoints below are expected to be exposed by the [=holder coordinator=], along with
the component that is expected to call the endpoint.
```

index.html
```diff
@@ -515,7 +514,16 @@ <h2>Terminology</h2>
...
-
+      <p>
+The below architecture description, as well as that seen in the diagram in the 
+<a href="#architecture-overview"></a> section, are representative of one possible 
+architecture for these software components. One software package may serve multiple 
+roles, but these roles are not required to be served by a single library or other 
+software package. For the purposes of this specification, each component will be 
+treated independently, both for clarity of reading, and to highlight that each 
+component can be implemented separately from any other, assuming some other software 
+is fulfilling the missing roles.
```
dlongley: Hmm, I tried to reorder this to emphasize that the expression of the architecture in the spec is for reader clarity and categorization -- and that particular expressed organizations aren't prescriptive, and then moved the "you can reorganize as it makes sense for your use case" to the end and added a bit. Could probably use some more wordsmithing, but I think at least this ordering gets at the goal a little better.

```suggestion
The architecture description below, as well as that seen in the diagram in the 
<a href="#architecture-overview"></a> section, are representative of one possible 
architecture for these software components. This representation was chosen to
make categorization of the different features in the specification clear, enabling
each component to be treated independently for both clarity of reading and
to highlight that each component can be implemented separately from any
other. However, this is not a prescriptive architecture; implementers can use as
many or as few software packages as desired to serve multiple roles; roles are
not required to be served by a single library or other software package nor are
software libraries or packages required to provide for particular roles in isolation
of others. It is also an expectation that single deployments of software or
system components can be for the purpose of multiple roles as needed for a
given use case.
```

components/parameters/path/ExchangeId.yml
eric-schuh: Yes, I forgot to branch my fork before pushing so the updates in https://github.com/w3c-ccg/vcalm/pull/559 appear here as well since they aren't accepted yet. The only update that is new here, and not in PR559 is the last one https://github.com/w3c-ccg/vcalm/pull/562/commits/a680cd20707be5ed03fa2d655ff9633006d7fdf6

index.html
TallTed: @dlongley — Your revision suggests that `the component that is expected to call the endpoint` is `expected to be exposed by the [=holder coordinator=]` which I think is not correct. I wanted to suggest a rephrasing of my own, but neither the [source `index.html`](https://github.com/w3c-ccg/vcalm/pull/562/files) nor the [Preview](https://pr-preview.s3.amazonaws.com/eric-schuh/vcalm/pull/562.html) shows the full content, so I cannot decipher what the rephrasing should be....

dlongley: Hmm, yeah, I was just trying to improve readability. We'll need a different suggestion or to figure out how to make it more clear what the scope is here for "below".

Issue #347, opened 2023-08-31
Add callback url to exchanges


brianorwhatever:
To avoid polling architecture should we think about a way to include a `callbackUrl` for the workflow to hit when an exchange has been complete?

TallTed:
I suggest adding a question mark to the end of this issue's title, and closing #346 as a dupe of this issue, as this issue is more completely described.

brianorwhatever:
Oops definitely accidentally duped this. Thanks for the catch!

dlongley:
One reason to add a callback URL might be to try and eliminate a need to poll an exchange for updates to its state. I imagine that's the main draw.

However, even if we had a callback URL, it seems to me that using it could fail for any variety of reasons, leading to a need to poll an exchange anyway. So if we were thinking that having a callback URL could allow for a push-only notification experience, that might be not be correct. It may instead just reduce the amount of polling necessary in the average case.

If this reasoning is correct, then we should consider the added complexity of having this feature according to this lesser appealing "average case" trade off instead of considering that it would eliminate polling entirely.

Other considerations around "callback URLs" involve what kind of authz would be required (could we keep it simple with capability URLs?), how many times can they be called, with what request data and what response data, etc.

jrhender:
I think this issue might be a duplicate (it might not though) of https://github.com/w3c-ccg/vc-api/issues/295 . I'd say it certainly relates to it.

msporny:
This is a duplicate of #295.

msporny:
The group discussed this on 2024-05-28 call. 

@PatStLouis noted that this is a more specific case of issue #295 and perhaps we could discuss in #295. Cover callback URLs in issue #295. @dlongley recommended closing #295 and keeping this one since it's more specific. The most useful place to put callbacks right now is in exchanges. @jandrieu +1 to solve the more specific use case (this issue). @PatStLouis reading #295, use case identified, report to external system on progress in an exchange, somewhat of a different use case. Make sure threads in other issue are captured here. @dlongley didn't think that there is a use case to report back on progress of issuance (atomic operation of issuance API). Use cases are around exchanges, other issue has been opened for a long time, it's still open/linked if it's not exchange related, let's close to make progress. @jandrieu noted that point of exchanges is to be abstraction, right layer to handle callbacks in general, this isn't just simpler use case, among APIs that are synchronous, we can move async stuff to exchanges (where all callback stuff moves).

[...]

PR #564, opened 2025-10-19
Add optional exchange step callback feature.

msporny:
This PR is an attempt to address issue #347 by defining the ability to call a callback when an exchange step is completed.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vcalm/pull/564.html" title="Last updated on Nov 18, 2025, 8:10 PM UTC (a8836ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vcalm/564/ee5b54a...a8836ef.html" title="Last updated on Nov 18, 2025, 8:10 PM UTC (a8836ef)">Diff</a>

### DIFF
--- a/components/parameters/path/LocalCallbackId.yml
+++ b/components/parameters/path/LocalCallbackId.yml
@@ -4,6 +4,4 @@ in: path
 required: true
 schema:
   type: string
-  # Pattern ensures at least 128 bits of entropy (minimum 22 base64url characters, or UUID format)
-  pattern: "^([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}|[A-Za-z0-9_-]{22,})$"
 


oas.yaml
```diff
@@ -632,6 +632,34 @@ paths:
             application/json:
               schema:
                 $ref: "./components/ProblemDetails.yml#/components/schemas/ProblemDetails"
+
+  /callback:
```
dlongley: I don't think we should spec this part out -- at least not like this. It's a goal for the callback to be a capability URL and this clearly wouldn't be, neither would "any arbitrary URL". Instead, we want the capability URL constraint as it provides authz without needing additional tokens or client registration. It could be an attack vector to use any arbitrary URL (that did not have the properties of a capability URL).

msporny: I can't easily spec the payload part out w/o adding something to the OAS file. 

The alternative would be to copy the bits of the section we want back into the spec after it has rendered. Is the issue that we don't say this is a capability URL? I could do so by changing `/callback` to `/capability-url-based-callback` or something similar.

Before I rip this part out of the OAS file, would like to understand what part you don't want to be spec'd out.

dlongley: If we want to name it "a-callback-capability-url" and include in the description that it must be a capability URL, that's  probably a fine approach.

dlongley: Maybe `/callbacks/<callbackId>` would suffice, where we say that the `callbackId` can be used to make the callback a capability URL by using a sufficiently large (128-bits+) random identifier.

TallTed: It seems to me that "any arbitrary URL" is to encompass the URLs that we recommend, which should either deliver some payload to a callback processor or themselves be served by the callback processor (which I think is what @dlongley means by a "capability URL"). I would expect that each "arbitrary URL" would be specific to the in-flight transaction, minted based on some element of the same in-flight transaction (and so, not easily targeted as an attack vector), and that some nonce-like data in the payload and/or the URL would help further minimize the attack surface.

index.html
```diff
@@ -2269,6 +2269,14 @@ <h4>Get Exchange State</h4>
           data-api-endpoint="get /workflows/{localWorkflowId}/exchanges/{localExchangeId}"></div>
       </section>
 
+      <section>
+        <h4>Exchange Step Callbacks</h4>
+        <p>
+        </p>
+
+        <div class="api-detail"
+          data-api-endpoint="post /callback"></div>
```
```suggestion
          data-api-endpoint="post /callbacks/{localCallbackId}"></div>
```

oas.yaml
```suggestion
  /callbacks/{localCallbackId}:
```

oas.yaml
```diff
@@ -632,6 +632,34 @@ paths:
...
+    post:
+      summary: A callback that can be any arbitrary URL that can be notified when an exchange step is performed.
```
```suggestion
      summary: A callback that can be any capability URL (i.e., a URL that is infeasible to guess) that can be notified when an exchange step is performed. The URL SHOULD be a capability URL to ensure that it can only be used by parties with whom the URL is shared without a need for authorization tokens, periodic token refresh, nor client registration. If the suggested URL format with "localCallbackId" is used, then the "localCallbackId" value must express at least 128-bits of random information in order to ensure the full callback URL can be treated as a capability URL. It is also recommended that a new capability URL be created per exchange and only used by that exchange.
```

oas.yaml
```diff
@@ -496,6 +496,183 @@ paths:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/VerifiablePresentationRequestBody"
+        "400":
+          description: Request is malformed.
+          content:
+            application/json:
+              schema:
+                $ref: "./components/ProblemDetails.yml#/components/schemas/ProblemDetails"
+
```
PatStLouis:
```suggestion
```

PatStLouis:
Linting somehow got broken after merging previous PR, will need to investigate and apply linting fixes.

components/parameters/path/LocalCallbackId.yml
```diff
@@ -0,0 +1,9 @@
+name: localCallbackId
+description: A local identifier for a callback that must express at least 128-bits of random information to ensure the full callback URL can be treated as a capability URL.
```
TallTed:
```suggestion
description: A local identifier for a callback. It must express at least 128-bits of random information to ensure the full callback URL can be treated as a capability URL.
```

components/parameters/path/LocalCallbackId.yml
```diff
@@ -0,0 +1,9 @@
+name: localCallbackId
+description: A local identifier for a callback. It must express at least 128-bits of random information to ensure the full callback URL can be treated as a capability URL.
+in: path
+required: true
+schema:
+  type: string
+  # Pattern ensures at least 128 bits of entropy (minimum 22 base64url characters, or UUID format)
+  pattern: "^([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}|[A-Za-z0-9_-]{22,})$"
```
dlongley:
```suggestion
```

Issue #542, opened 2024-07-16
Add `acceptedEnvelopes` and `acceptedCryptosuites` examples for `QueryByExample`


dlongley:
For example, these two examples show that the sender will accept a number of cryptosuites or at least one envelope. The second example also highlights that only a limited subset of the VC is requested, so if a selective disclosure mechanism is available and chosen, just that field can be revealed.

```js
{
  "type": "QueryByExample",
  "credentialQuery": {
    "reason": "Please present your PermanentResidentCard or Verifiable Credential(s) to complete the verification process.",
    "example": {
      "@context": [
        "https://www.w3.org/2018/credentials/v1",
        "https://w3id.org/citizenship/v1"
      ],
      "type": ["PermanentResidentCard"]
    },
    "acceptedCryptosuites": [
      "Ed25519Signature2020",
      "eddsa-rdfc-2022",
      "ecdsa-rdfc-2019",
      "bbs-2023",
      "ecdsa-sd-2023"
    ],
    "acceptedEnvelopes": [
      "application/jwt"
    ]
  }
}
```

```js
{
  "type": "QueryByExample",
  "credentialQuery": {
    "reason": "Please present your PermanentResidentCard or Verifiable Credential(s) to complete the verification process.",
    "example": {
      "@context": [
        "https://www.w3.org/2018/credentials/v1",
        "https://w3id.org/citizenship/v1"
      ],
      "type": ["PermanentResidentCard"],
      "credentialSubject": {
        "birthCountry": ""
      }
    },
    "acceptedCryptosuites": [
      "Ed25519Signature2020",
      "eddsa-rdfc-2022",
      "ecdsa-rdfc-2019",
      "bbs-2023",
      "ecdsa-sd-2023"
    ],
    "acceptedEnvelopes": [
      "application/jwt"
    ]
  }
}
```

PR #580, opened 2026-01-26
Add acceptedCryptosuites and acceptedEnvelopes examples for QueryByExample

bparth24:
Document how to express accepted cryptosuites or envelopes in a VPR with examples.

This PR makes it clear that:
- `example` is optional in QBE
- `acceptedCryptosuites` is optional
- `acceptedEnvelopes` is optional

Addresses #542.

### DIFF
--- a/components/VerifiablePresentationRequest.yml
+++ b/components/VerifiablePresentationRequest.yml
@@ -37,9 +37,9 @@ components:
                   "example":
                     type: object
                     description: An example of the credential being requested.
-                  "trustedIssuers":
+                  "acceptedIssuers":
                     type: array
-                    description: Trusted issuers for the requested credential.
+                    description: Accepted issuers for the requested credential.
                     items:
                       type: object
                       properties:
--- a/index.html
+++ b/index.html
@@ -1271,9 +1271,9 @@ or missing `group` values are processed as "OR" operations.
           "type": "Alumni"
         }
       },
-      <span class="comment">// (Optional) Specify credentials from particular trustedIssuers or
+      <span class="comment">// (Optional) Specify credentials from particular acceptedIssuers or
       // delegates of the authority</span>
-      "trustedIssuers": [{
+      "acceptedIssuers": [{
         "issuer": "did:web:authority.example"
       }]
     }
@@ -1309,9 +1309,9 @@ including its <code>@context</code>, <code>type</code>, and optionally
             </td>
           </tr>
           <tr>
-            <td>trustedIssuers</td>
+            <td>acceptedIssuers</td>
             <td>
-An OPTIONAL array of objects specifying trusted [=issuers=] for the requested
+An OPTIONAL array of objects specifying accepted [=issuers=] for the requested
 credential. Each object MAY contain an <code>issuer</code> property with a
 value that is an [=issuer=] identifier.
             </td>


index.html
```diff
@@ -1283,6 +1283,115 @@ <h3>Query By Example</h3>
...
+including its <code>@context</code>, <code>type</code>, and optionally
+<code>credentialSubject</code> fields to indicate which claims are needed.
+            </td>
+          </tr>
+          <tr>
+            <td>authority</td>
+            <td>
+An OPTIONAL array of objects specifying trusted [=issuers=] for the requested
+credential. Each object MAY contain an <code>issuer</code> property with a
+value that is an [=issuer=] identifier.
```
dlongley: I think `authority` would be a new property -- and we shouldn't name it `authority` as it could create the wrong impression about how VCs work and their decentralized ecosystem, I think. Maybe `trustedIssuers` would be ok -- which is what it was in the previous VPR spec modulo being plural.

bparth24: @dlongley 

Sure. I can update `authority` to `trustedIssuers`. **Note:** I got that `authority` from [`Example 2`](https://w3c-ccg.github.io/vcalm/#example-a-query-by-example-query).

bparth24: @dlongley 

Addressed in this commit - https://github.com/w3c-ccg/vcalm/pull/580/commits/5071eb0d1c342e00e53b3e68df3f63ca63d2e328.

dlongley: Ah, we should fix Example 2 then as well. That looks like a copy mistake when we moved over from the VPR spec.

bparth24: @dlongley 

Yes, I did that fix already in this commit - https://github.com/w3c-ccg/vcalm/commit/5071eb0d1c342e00e53b3e68df3f63ca63d2e328. Refer - https://github.com/w3c-ccg/vcalm/pull/580/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R1276

components/VerifiablePresentationRequest.yml
```diff
@@ -27,6 +27,36 @@ components:
...
+                      type: object
+                      properties:
+                        "issuer":
+                          type: string
+                  "acceptedCryptosuites":
+                    type: array
+                    description: Cryptographic proof suites the verifier will accept.
+                    items:
+                      type: string
+                    example: ["eddsa-rdfc-2022", "bbs-2023", "ecdsa-sd-2023"]
```
dlongley: Implementations today use an object for each array element to allow for future extensibility where other properties might need to be listed. I recommend the PR reflect what implementations are doing today, even if we want to allow a simpler version after group discussion. If we want to allow this shortcut, we could say that if a string value is given it is taken to be the cryptosuite name for a `DataIntegrity` proof.

```suggestion
                    example: [{"cryptosuite": "eddsa-rdfc-2022"}, {"cryptosuite": "bbs-2023"}, {"cryptosuite": "ecdsa-sd-2023"}]
```

components/VerifiablePresentationRequest.yml
```diff
@@ -27,6 +27,36 @@ components:
...
+                  "acceptedEnvelopes":
+                    type: array
+                    description: Envelope formats the verifier will accept.
+                    items:
+                      type: string
+                    example: ["application/jwt", "application/vc+sd-jwt"]
```
dlongley: While the `acceptedCryptosuites` array uses objects for the array elements, the `acceptedEnvelopes` array does not (in current implementations), so this example matches. However, this is an inconsistency I wanted to highlight if we want to have some further discussion to perhaps also support: `[{mediaType: "application/jwt"}, ...` instead of just using raw strings as the elements. This would allow for other parameters to be specified if necessary. For backwards compatibility we could say that if the value is a string, it is taken to be the media type for the envelope.

components/VerifiablePresentationRequest.yml
```diff
@@ -67,7 +97,15 @@ components:
                   "https://www.w3.org/2018/credentials/examples/v1"
                 ],
                 "type": "UniversityDegreeCredential"
-              }
+              },
+              "acceptedCryptosuites": [
+                "eddsa-rdfc-2022",
+                "ecdsa-rdfc-2019",
+                "bbs-2023"
+              ],
```
dlongley, line 121:
```suggestion
              "acceptedCryptosuites": [
                {"cryptosuite": "eddsa-rdfc-2022"},
                {"cryptosuite": "ecdsa-rdfc-2019"},
                {"cryptosuite": "bbs-2023"}
              ],
```

index.html
```diff
@@ -1283,6 +1283,115 @@ <h3>Query By Example</h3>
...
+          <tr>
+            <td>acceptedCryptosuites</td>
+            <td>
+An OPTIONAL array of [=strings=] specifying which cryptographic proof suites
+the [=verifier=] will accept. This enables the [=holder=] to select an
+appropriate proof format. Valid example values include:
+<ul>
+  <li><code>["eddsa-rdfc-2022", "ecdsa-rdfc-2019"]</code></li>
+  <li><code>["bbs-2023", "ecdsa-sd-2023"]</code></li>
+</ul>
```
dlongley: This will need adjustment per the comments above and any group discussion.

index.html
```diff
@@ -1283,6 +1283,115 @@ <h3>Query By Example</h3>
...
+        "credentialSubject": {
+          "birthCountry": ""
+        }
+      },
+      "acceptedCryptosuites": [
+        "eddsa-rdfc-2022",
+        "ecdsa-rdfc-2019",
+        "bbs-2023",
+        "ecdsa-sd-2023"
+      ],
```
dlongley, line 1377: See comments above and any group discussion.

index.html
```diff
@@ -1271,9 +1271,9 @@ <h3>Query By Example</h3>
           "type": "Alumni"
         }
       },
-      <span class="comment">// (Optional) Specify credentials from a particular authority or
+      <span class="comment">// (Optional) Specify credentials from a particular trustedIssuer(s) or
       // delegate of the authority</span>
```
TallTed: `a particular` means singular `trustedIssuer` or `delegate`. If plural `trustedIssuers` or `delegates` is possible, the edit will be more complex.
```suggestion
      <span class="comment">// (Optional) Specify credentials from a particular trustedIssuer(s) or
      // delegate of the authority</span>
```

index.html
```diff
@@ -1283,6 +1283,115 @@ <h3>Query By Example</h3>
...
+          </tr>
+        </table>
+
+        <p>
+The following example demonstrates a QueryByExample that specifies accepted
+cryptosuites and envelope formats, enabling the [=holder=] to choose an
+appropriate proof mechanism:
+        </p>
+
+        <pre class="example nohighlight" title="A Query By Example with accepted cryptosuites and envelopes">
```
```suggestion
        <pre class="example nohighlight" title="A Query By Example with acceptable cryptosuites and envelopes">
```

components/VerifiablePresentationRequest.yml
bparth24: > Implementations today use an object for each array element to allow for future extensibility where other properties might need to be listed. I recommend the PR reflect what implementations are doing today, even if we want to allow a simpler version after group discussion. If we want to allow this shortcut, we could say that if a string value is given it is taken to be the cryptosuite name for a `DataIntegrity` proof.

@dlongley @PatStLouis @TallTed 

Addressed the above changes in this commit - https://github.com/w3c-ccg/vcalm/pull/580/commits/28f3e32a9b3319ef610520d1b29d8a46ee6e33e7.

components/VerifiablePresentationRequest.yml
bparth24: > While the `acceptedCryptosuites` array uses objects for the array elements, the `acceptedEnvelopes` array does not (in current implementations), so this example matches. However, this is an inconsistency I wanted to highlight if we want to have some further discussion to perhaps also support: `[{mediaType: "application/jwt"}, ...` instead of just using raw strings as the elements. This would allow for other parameters to be specified if necessary. For backwards compatibility we could say that if the value is a string, it is taken to be the media type for the envelope.

@dlongley @PatStLouis @TallTed 

Addressed the above changes in this commit - https://github.com/w3c-ccg/vcalm/pull/580/commits/28f3e32a9b3319ef610520d1b29d8a46ee6e33e7.

components/VerifiablePresentationRequest.yml
bparth24: @dlongley @PatStLouis @TallTed 

Addressed the above changes in this commit - https://github.com/w3c-ccg/vcalm/pull/580/commits/28f3e32a9b3319ef610520d1b29d8a46ee6e33e7.

components/VerifiablePresentationRequest.yml
```diff
@@ -27,6 +27,44 @@ components:
...
+                type: object
+                description: For QueryByExample queries, specifies the credential request details.
+                properties:
+                  "reason":
+                    type: string
+                    description: A human-readable explanation for why the credential is being requested.
+                  "example":
+                    type: object
+                    description: An example of the credential being requested.
+                  "trustedIssuers":
```
dlongley: I think the plan was to leave this off for now and add an issue to the spec. Either that or people were amendable to `acceptedIssuers` towards the end of the call, so we could put that here and see if the group agrees:

```suggestion
                  "acceptedIssuers":
```

components/VerifiablePresentationRequest.yml
```diff
@@ -27,6 +27,44 @@ components:
...
+                    type: array
+                    description: Trusted issuers for the requested credential.
```
dlongley:
```suggestion
                    description: Accepted issuers for the requested credential.
```

components/VerifiablePresentationRequest.yml
```diff
@@ -27,6 +27,44 @@ components:
...
+                  "acceptedEnvelopes":
+                    type: array
+                    description: Envelope formats the verifier will accept. Each element SHOULD be an object with a `mediaType` property; however, for backwards compatibility, string values are also accepted and interpreted as the media type for the envelope.
+                    items:
+                      type: object
+                      properties:
+                        "mediaType":
+                          type: string
+                          description: The media type of the envelope format.
+                    example: [{"mediaType": "application/jwt"}, {"mediaType": "application/vc+sd-jwt"}]
```
dlongley: We'll need to accept either an array of objects or an array of strings, where if strings are used they are just syntactic sugar for objects with "mediaType".

components/VerifiablePresentationRequest.yml
```diff
@@ -27,6 +27,44 @@ components:
...
+                  "acceptedCryptosuites":
+                    type: array
+                    description: Cryptographic proof suites the verifier will accept. Each element SHOULD be an object with a `cryptosuite` property; however, for backwards compatibility, string values are also accepted and interpreted as the cryptosuite name for a Data Integrity proof.
+                    items:
+                      type: object
+                      properties:
+                        "cryptosuite":
+                          type: string
+                          description: The name of the cryptosuite.
+                    example: [{"cryptosuite": "eddsa-rdfc-2022"}, {"cryptosuite": "bbs-2023"}, {"cryptosuite": "ecdsa-sd-2023"}]
```
dlongley: We'll need to accept either an array of objects or an array of strings, where if strings are used they are just syntactic sugar for objects with "cryptosuite".

components/VerifiablePresentationRequest.yml
bparth24: > I think the plan was to leave this off for now and add an issue to the spec. Either that or people were amendable to `acceptedIssuers` towards the end of the call, so we could put that here and see if the group agrees:

@dlongley yes, I remember that I was thinking of reverting to `authority` at least for this PR at the end of all the changes. Then, create a separate issue as the group decided. Either way works for me.

components/VerifiablePresentationRequest.yml
bparth24: @dlongley 

Addressed in this commit - https://github.com/w3c-ccg/vcalm/pull/580/commits/b9c43819892f50efe26c5d4b587159cbfdfb3fea.

components/VerifiablePresentationRequest.yml
bparth24: @dlongley 

Addressed in this commit - https://github.com/w3c-ccg/vcalm/pull/580/commits/b9c43819892f50efe26c5d4b587159cbfdfb3fea.

index.html
bparth24: > This will need adjustment per the comments above and any group discussion.

@dlongley 

Adjustment addressed in this commit - https://github.com/w3c-ccg/vcalm/pull/580/commits/4a041ad3cbb11880206e6d3520bbc6ae15f4cf84.

index.html
bparth24: @dlongley 

Adjustment addressed in this commit - https://github.com/w3c-ccg/vcalm/pull/580/commits/4a041ad3cbb11880206e6d3520bbc6ae15f4cf84.

components/VerifiablePresentationRequest.yml
dlongley: Let's put `acceptedIssuers` in this PR and file an issue to have the group decide what we want to do. I think it's clear we definitely don't want `authority` so let's get that removed from the spec -- and that might be most easily accomplished by replacing it with `acceptedIssuers` vs. it disappearing entirely. The group seems amenable to `acceptedIssuers` but we should confirm that via a new issue and discussion in the group.

index.html
bparth24: > `a particular` means singular `trustedIssuer` or `delegate`. If plural `trustedIssuers` or `delegates` is possible, the edit will be more complex.

@TallTed 

Addressed in this commit - https://github.com/w3c-ccg/vcalm/pull/580/commits/60ca7620abd4920da95195fb90b30f050b747a8a.

index.html
bparth24: @TallTed 

Addressed in this commit - https://github.com/w3c-ccg/vcalm/pull/580/commits/60ca7620abd4920da95195fb90b30f050b747a8a.

components/VerifiablePresentationRequest.yml
bparth24: @dlongley 

Addressed in this commit - https://github.com/w3c-ccg/vcalm/pull/580/commits/170d9705ecccdfc82fb967ae92485e7ac2674872.

components/VerifiablePresentationRequest.yml
bparth24: @dlongley 

Addressed in this commit - https://github.com/w3c-ccg/vcalm/pull/580/commits/170d9705ecccdfc82fb967ae92485e7ac2674872.

dlongley (approved):

Issue #527, opened 2025-09-02
Specify optional "clientProfiles" in "openId" workflow step to support OID4VP multiplexing


dlongley:
Some versions or profiles of OID4VP require various different parameters to be used depending on the protocol or format requested, even with the same exchange. For example, an exchange that requests *either* an mDL or a VC would require two different schemes in the URLs given to wallets and different parameters, such as client IDs or schemes, etc. when using ISO 18013-7 as opposed to newer DQCL-based OID4VP queries.

In order to facilitate all of these variants, a `clientProfiles` option can be included in a workflow step that specifies any different parameters while working from the same base variables and steps for the exchange -- and allowing the issuer and / or verifier coordinator to continue to work with a single exchange while different wallets can make different protocol and format choices when interacting.

The suggestion here is to add an optional `clientProfiles` object to the `openId` object in a `step`, where the keys are the name of the client profile (which can be chosen to be anything useful by the workflow step designer), and where the value of each client profile has options that would match this JSON schema:

```
const oid4vpClientProfile = {
  title: 'OID4VP Client Profile',
  type: 'object',
  additionalProperties: false,
  // an authorization request or a directive to create one can be used,
  // but not both
  oneOf: [{
    required: ['createAuthorizationRequest'],
    // cannot also use `authorizationRequest`
    not: {
      required: ['authorizationRequest']
    }
  }, {
    required: ['authorizationRequest'],
    // cannot also use `createAuthorizationRequest`
    not: {
      required: ['createAuthorizationRequest']
    }
  }],
  properties: {
    // value is name of variable to store the created authz request in
    createAuthorizationRequest: {
      type: 'string'
    },
    // ... or full authz request to use
    authorizationRequest: {
      type: 'object'
    },
    // optional properties that will be used as overrides in any authz request
    client_id: {type: 'string'},
    client_id_scheme: {type: 'string'},
    client_metadata: {type: 'object'},
    nonce: {type: 'string'},
    presentation_definition: {type: 'object'},
    response_mode: {type: 'string'},
    response_uri: {type: 'string'},
    // optional parameters for signing authorization requests
    authorizationRequestSigningParameters: {
      type: 'object',
      required: ['x5c'],
      additionalProperties: false,
      properties: {
        x5c: {
          type: 'array',
          minItems: 1,
          items: {
            type: 'string'
          }
        }
      }
    },
    // optional protocol URL parameters
    protocolUrlParameters: {
      type: 'object',
[...]

PR #581, opened 2026-01-28
OpenID client profile exchange configuration

kezike:
This PR addresses Issue #527 by introducing OpenID client profile exchange configuration.

### DIFF
--- a/oas.yaml
+++ b/oas.yaml
@@ -288,16 +288,16 @@ paths:
   /credentials/{id}:
     get:
       tags:
-       - Credentials
+        - Credentials
       security:
-       - networkAuth: []
-       - oAuth2: []
-       - zCap: []
+        - networkAuth: []
+        - oAuth2: []
+        - zCap: []
       summary: Gets a credential or verifiable credential by ID. To get a credential that does not have credential.id set but has an associated credentialId value, pass credentialId instead.
       x-expectedCaller: 
-       - Issuer Coordinator
-       - Holder Coordinator
-       - Workflow Service
+        - Issuer Coordinator
+        - Holder Coordinator
+        - Workflow Service
       operationId: getCredential
       parameters:
         - $ref: "./components/parameters/path/ObjectId.yml"
@@ -321,11 +321,11 @@ paths:
           description: I'm a teapot - MUST not be returned outside of pre-arranged scenarios between both parties
     delete:
       tags:
-       - Credentials
+        - Credentials
       security:
-       - networkAuth: []
-       - oAuth2: []
-       - zCap: []
+        - networkAuth: []
+        - oAuth2: []
+        - zCap: []
       summary: Deletes a credential or verifiable credential by ID. To delete a credential that does not have credential.id set but has an associated credentialId value, pass credentialId instead.
       x-expectedCaller:
         - Issuer Coordinator
@@ -347,17 +347,17 @@ paths:
   /credentials/derive:
     post:
       tags:
-       - Credentials
+        - Credentials
       security:
-       - networkAuth: []
-       - oAuth2: []
-       - zCap: []
+        - networkAuth: []
+        - oAuth2: []
+        - zCap: []
       summary: Derives a credential and returns it in the response body.
       operationId: deriveCredential
       description: Derives a credential and returns it in the response body.
       x-expectedCaller: 
-       - Holder Coordinator
-       - Workflow Service
+        - Holder Coordinator
+        - Workflow Service
       requestBody:
         content:
           application/json:
@@ -377,16 +377,16 @@ paths:
   /presentations/{id}:
     get:
       tags:
-       - Presentations
+        - Presentations
       summary: Gets a presentation or verifiable presentation by ID
       security:
-       - networkAuth: []
-       - oAuth2: []
-       - zCap: []
+        - networkAuth: []
+        - oAuth2: []
+        - zCap: []
       operationId: getPresentation
       x-expectedCaller: 
-       - Holder Coordinator
-       - Workflow Service
+        - Holder Coordinator
+        - Workflow Service
       parameters:
         - $ref: "./components/parameters/path/ObjectId.yml"
       responses:
@@ -408,12 +408,12 @@ paths:
           description: Gone! There is no data here
     delete:
       tags:
-       - Presentations
+        - Presentations
       summary: Deletes a presentation or verifiable presentation by ID
       security:
-       - networkAuth: []
-       - oAuth2: []
-       - zCap: []
+        - networkAuth: []
+        - oAuth2: []
+        - zCap: []
       operationId: deletePresentation
       x-expectedCaller: Holder Coordinator
       parameters:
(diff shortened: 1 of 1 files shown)


dlongley:
@kezike,

> I introduced the clientProfileId field in the openId workflow step variable in PR https://github.com/w3c-ccg/vcalm/pull/549. Should I remove it now that we are adding the clientProfiles field or do we still expect that it will be needed, because there will only ever be one client profile that is activated (and hence only one associated result variable that is expected) in a given exchange?

It should remain for the reason you said: Only one client profile will be activated in an exchange. As a side note, though, I looked closer at the existing description and it isn't correct. The description seems to be for a `client_id` which a `clientProfileId` is not. The description should just say that it is the identifier for the chosen client profile. (In other words, it holds the JSON key value from the `clientProfiles` map for the chosen client profile).

> Note that presentation_definition and presentation_definition_uri were not included, despite their reference in Issue https://github.com/w3c-ccg/vcalm/issues/527. This is because Presentation Exchange has been removed from OID4VP. Should I leave it out? Also, should I add a DCQL field or some other presentation submission field? Finally, should I also remove the reserved presentationSubmission exchange variable (defined in the ExchangeStepResults.openId schema) for the same reason?

Hmm, I guess we can include `presentation_definition` and list it as an optional override for any `presentation_definition` that would be auto-generated from a VPR or in the specified / auto-generated authorization request. We don't need `presentation_definition_uri` since that, I don't believe, is even really used in legacy/deprecated software. It would be nice for people who need to still work with PE in the near term to see they can put a `presentation_definition` override in place. If we need to remove that later or mark it non-normative/reserved we can do that later as well.

We should add a `dcql_query` field, yes, that serves the same "override" function, but is for a DCQL query object.

Let's leave the `presentationSubmission` exchange variable there for now. We're going to need to figure out, later, if/how we want to mark these things as reserved/non-normative/etc. since they will not become standards but are useful for implementers who have to deal with interoperability issues with OID4*.

> Issue https://github.com/w3c-ccg/vcalm/issues/527 referenced response_uri, but this should be specified in the authorization_request. Assuming you meant this to be redirect_uri. Is that correct?

I think that would be there to allow for an override of what goes into an auto-generated authz request as well.

oas.yaml
```diff
@@ -1218,3 +1208,95 @@ components:
...
+        authorizationRequest:
+          type: object
+          description: If present, an OID4VP authorization request for subsequent use in the exchange. Either this or <code>createAuthorizationRequest</code> MUST be present, but not both.
+        # The following are optional properties that can be used as overrides in any authorization request:
+        # ====================================================================================================
+        client_id:
+          type: string
+          description: The identifier of the OID4VCI/OID4VP profile to use in an exchange (defined in RFC 6749).
```
dlongley: This isn't the profile / or about client profiles, but about the client itself (though each profile can reference a different client). Here's some text to try and explain how this works according to OID4VP:

```suggestion
          description: The identifier of the OID4VP client. This is defined in RFC 6749 as the application making protected resource requests on behalf of the resource owner and with its authorization. In OID4VP, the client acts on behalf of the party making a request for a verifiable presentation; for example, a server that is interacting with a digital wallet.
```

oas.yaml
```diff
@@ -1218,3 +1208,95 @@ components:
...
+        client_id_scheme:
+          type: string
+          description: The scheme of the identifier of the OID4VCI/OID4VP profile to use in an exchange (defined in OID4VP, drafts 20+).
```
dlongley: This was actually removed sometime after draft 18. It is not present in OID4VP 1.0 as the scheme is now part of the `client_id` itself. However, it is needed for backwards compatibility, so we should keep it here but here's some different suggested text:

```suggestion
          description: The scheme of the identifier of the OID4VP client. This is defined in OID4VP drafts 18, but was removed in OID4VP 1.0. It is made available for backwards compatibility purposes with the popular OID4VP draft 18.
```

oas.yaml
```diff
@@ -1218,3 +1208,95 @@ components:
...
+        client_metadata:
+          type: object
+          description: Important metadata about the OID4VCI/OID4VP profile to use in an exchange (defined in RFC 7591).
```
dlongley: All of these authorization request fields are for OID4VP.

```suggestion
          description: Important metadata about the OID4VP client to use in an exchange (defined in RFC 7591).
```

oas.yaml
```diff
@@ -1218,3 +1208,95 @@ components:
...
+        nonce:
+          type: string
+          description: A single-use value created by the OID4VCI/OID4VP profile to prevent replay attacks during an exchange (defined in OIDC 1.0).
```
```suggestion
          description: A single-use value created by the OID4VP client to prevent replay attacks during an exchange (defined in OIDC 1.0).
```

oas.yaml
```diff
@@ -1218,3 +1208,95 @@ components:
...
+        response_mode:
+          type: string
+          description: A directive from an OID4VCI/OID4VP instructing how to submit an authorization response during an exchange (defined in OIDC 1.0).
```
```suggestion
          description: A directive from an OID4VP instructing how to submit an authorization response during an exchange (defined in OIDC 1.0).
```

oas.yaml
```diff
@@ -1218,3 +1208,95 @@ components:
...
+        redirect_uri:
+          type: string
+          description: A URI registered by the OID4VCI/OID4VP profile to indicate where to redirect UX control after an exchange (defined in RFC 6749).
```
```suggestion
          description: A URI specified by the OID4VP client to indicate where to redirect UX control after an exchange (defined in RFC 6749).
```

oas.yaml
```diff
@@ -1218,3 +1208,95 @@ components:
...
+        # ====================================================================================================
+        authorizationRequestSigningParameters:
+          type: object
+          additionalProperties: false
+          description: Parameters needed for signing authorization requests submitted to an OID4VCI/OID4VP profile during an exchange.
```
```suggestion
          description: Parameters needed for signing authorization requests submitted via OID4VP during an exchange.
```

oas.yaml
```diff
@@ -1218,3 +1208,95 @@ components:
...
+            scheme:
+              type: string
+              description: Scheme used in OID4VCI/OID4VP URL
+          required:
+            - name
+            - scheme
+        zcapReferenceIds:
+          type: object
+          additionalProperties: false
+          description: References to ZCAPs needed when interacting with an OID4VCI/OID4VP profile during an exchange.
```
```suggestion
          description: References to ZCAPs needed during an exchange that is using OID4VCI/OID4VP.
```

oas.yaml
```diff
@@ -1218,3 +1208,95 @@ components:
...
+          properties:
+            signAuthorizationRequest:
+              type: string
+              description: Reference to a ZCAP needed for signing an authorization request submitted to an OID4VCI/OID4VP profile during an exchange.
```
```suggestion
              description: Reference to a ZCAP needed for signing an authorization request submitted during an exchange using OID4VP.
```

oas.yaml
```diff
@@ -1218,3 +1208,95 @@ components:
...
+              - authorizationRequest
+        - required:
+            - authorizationRequest
+          not:
+            required:
+              - createAuthorizationRequest
+    OpenIdExchangeConfig:
+      type: object
+      additionalProperties: false
+      description: Optional parameters to enable OID4VCI/OID4VP protocol for credential delivery/presentation.
```
```suggestion
      description: Optional parameters to enable OID4VCI/OID4VP protocols for credential delivery/presentation.
```

PatStLouis (approved):

Issue #538, opened 2023-07-07
Questions regarding `QueryByExample` structure and handling


geel9:
Hello,

I have some questions regarding the proper handling of `QueryByExample` objects.

I understand that many of these questions may not yet have answers, but I'd appreciate any insight that can be provided.

1. Are the `@context` and `type` fields required or optional?
2. What is the datatype of these fields? 
    - I am operating under the assumption that they are both `string | string[]` -- is this fair?
3. How are the `@context` and `type` fields to be handled?
	- I am operating under the assumption that every value in the example `@context` array (whether it be expressed as an array of strings or a string itself) must be present in the credential's `@context` array, but additional values may be present in the credential's `@context` array. Is this fair?
	- I am operating under an equal assumption for the `type` field.
4. Is `credentialSubject` limited to only `id` and `name` fields, or is it intended for arbitrary key/values to query against the credential's `credentialSubject`?
    - My intuition tells me that any key/value pairs are valid, but the comment above it in Example 2 states `You can request a specific subject id`, which implies that that is the _only_ intended usage.
5. Assuming `credentialSubject` is not limited to `id` and `name`, how is comparison meant to be handled?
	- Comparisons of primitives is intuitive, but stringy equality (does `5` match `"5"`?) should be clarified.
	- Comparisons of objects is also intuitive -- just recurse.
		- I am assuming, however, that all object comparisons (including the top-level `example.credentialSubject` to `credential.credentialSubject` comparison itself) are loose in the sense that the object on the credential may have additional fields not specified by the example object.
	- How should one perform a comparison of arrays?
		- See below

### Array Comparison in `credentialSubject`
_(This is all assuming that the `credentialSubject` field in the example query is not restricted to `id` and `name` fields)_

Given the following query:

```json
{
  "type": "QueryByExample",
  "credentialQuery": {
    "example": {
      "credentialSubject": {
        "primitive_array_one": [0, 1, 2],
        "primitive_array_two": [10, 11, 12],
        "complex_array": [
          {
            "a": 1
          },
          {
            "b": 2
          }
        ]
      }
    }
  }
}
```

And the following `credentialSubject` from a VerifiableCredential:

```json
{
  "credentialSubject": {
    "id": "whatever",
    "primitive_array_one": [0, 1, 2, 3],
    "primitive_array_two": [12, 11, 10],
    "complex_array": [
      {
        "a": 1,
        "b": 2,
      },
      {
        "b": 3
      }
    ]
  }
}
```

[...]

PR #593, opened 2026-02-10
Add JSON Schema for QueryByExample

msporny:
This PR is an attempt to address issue #538 by providing a JSON Schema for QueryByExample.

### DIFF
--- a/schemas/qbe.json
+++ b/schemas/qbe.json
@@ -163,10 +163,8 @@
       }
     },
     "AcceptedCryptosuite": {
-      "type": "object",
-      "description": "Specification of an accepted cryptographic suite",
-      "properties": {
-        "cryptosuite": {
+      "oneOf": [
+        {
           "type": "string",
           "description": "The name of the cryptographic suite",
           "examples": [
@@ -175,9 +173,25 @@
             "bbs-2023",
             "ecdsa-sd-2023"
           ]
+        },
+        {
+          "type": "object",
+          "description": "Specification of an accepted cryptographic suite",
+          "properties": {
+            "cryptosuite": {
+              "type": "string",
+              "description": "The name of the cryptographic suite",
+              "examples": [
+                "eddsa-rdfc-2022",
+                "ecdsa-rdfc-2019",
+                "bbs-2023",
+                "ecdsa-sd-2023"
+              ]
+            }
+          },
+          "required": ["cryptosuite"]
         }
-      },
-      "required": ["cryptosuite"]
+      ]
     },
     "AcceptedIssuer": {
       "oneOf": [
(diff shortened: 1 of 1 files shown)


schemas/qbe.json
```diff
@@ -0,0 +1,203 @@
...
+        "example": {
+          "$ref": "#/$defs/CredentialExample"
+        },
+        "acceptedIssuers": {
+          "type": "array",
+          "description": "List of accepted issuer DIDs or URIs",
+          "items": {
+            "type": "string",
+            "format": "uri"
+          },
```
dlongley, line 67: We will want an "AcceptedIssuer" type here as well, to allow an object with various properties.

schemas/qbe.json
```diff
@@ -0,0 +1,203 @@
...
+          "minItems": 1
+        },
+        "acceptedCryptosuites": {
+          "type": "array",
+          "description": "List of accepted cryptographic suites for credential proofs",
+          "items": {
+            "$ref": "#/$defs/Cryptosuite"
```
dlongley: We might want to call this an "AcceptedCryptosuite" type instead.

schemas/qbe.json
```diff
@@ -0,0 +1,203 @@
...
+          },
+          "minItems": 1
+        },
+        "acceptedEnvelopes": {
+          "type": "array",
+          "description": "List of accepted envelope formats for credentials",
+          "items": {
+            "$ref": "#/$defs/Envelope"
```
dlongley: We might want to call this an "AcceptedEnvelope" type instead.

schemas/qbe.json
```diff
@@ -0,0 +1,203 @@
...
+              },
+              "minItems": 1
+            }
+          ],
+          "description": "The type(s) of the credential being requested"
+        },
+        "credentialSubject": {
+          "type": "object",
+          "description": "Properties that the credential subject should have",
+          "properties": {
```
dlongley, line 134: We might need to explicitly set `additionalProperties: true`. I don't recall what the JSON schema default value is.

schemas/qbe.json
msporny: @dlongley noted on the call today that we should think about supporting this:

```acceptedIssuers: [did, {"recognizedIn": <here>}]```

File: schemas/qbe.json (hunk 4 above)
Comment (msporny) [reply schemas/qbe.json:134]: By default, `additionalProperties` is set to `true` in JSON Schema:

https://json-schema.org/understanding-json-schema/reference/object#additionalproperties

File: schemas/qbe.json (hunk 1 above)
Comment (msporny) [reply schemas/qbe.json:67]: Fixed in 166028abd72331097a5f1dd7f74ae8cf2647eb1d.

File: schemas/qbe.json (hunk 2 above)
Comment (msporny) [reply schemas/qbe.json]: Fixed in 166028abd72331097a5f1dd7f74ae8cf2647eb1d.

File: schemas/qbe.json (hunk 3 above)
Comment (msporny) [reply schemas/qbe.json]: Fixed in 166028abd72331097a5f1dd7f74ae8cf2647eb1d.

File: schemas/qbe.json [hunk 5]
```diff
@@ -0,0 +1,223 @@
...
+              "format": "uri"
+            },
+            {
+              "type": "object"
+            }
+          ],
+          "description": "The issuer of the credential"
+        }
+      },
+      "required": ["@context", "type", "credentialSubject"]
```
msporny:
```suggestion
      }
```

schemas/qbe.json
```diff
@@ -0,0 +1,223 @@
...
+        "acceptedEnvelopes": {
+          "type": "array",
+          "description": "List of accepted envelope formats for credentials",
+          "items": {
+            "$ref": "#/$defs/AcceptedEnvelope"
+          },
+          "minItems": 1
+        }
+      },
+      "required": ["reason", "example"]
```
msporny:
```suggestion
      }
```

PR #597, opened 2026-02-24
Fix DELETE /presentations/{id} component table link

eric-schuh:
The DELETE /presentations/{id} entry in the component tables was missing a link due to the missing section in the spec for the endpoint. Now that it is added the component table can be updated.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1837,7 +1837,7 @@ strongly advised to not assign semantics to any human-readable values.
       </section>
 
       <!--If this header is changed, also update the x-componentTableLink parameter for this endpoint in the oas.yaml file.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1836,6 +1836,8 @@ <h4>Get a Specific Presentation</h4>
           data-api-endpoint="get /presentations/{id}"></div>
       </section>
 
+      <!--If this header is changed, also update the x-componentTableLink parameter for this endpoint in the oas.yaml file.
+          If the x-componentTableLink isn't updated to match the corresponding Component Table link will break.-->
```
```suggestion
          If the x-componentTableLink isn't updated to match the corresponding Component Table, link will break.-->
```

Issue #599, opened 2026-03-02
Nonsensical display for "variables" on create exchange endpoint


ottonomy:
It's nice to have the auto-generation of the endpoint model code from data models, but the data we have for "variables" under the create exchange endpoint documentation is odd. 

> `variables` `[object]` Variables needed for an exchange. It MUST be an object of the following form:
> 
> results [object] ...

It looks like you have to pass in an object like:
```
{
  "results": {
    "someStep": {
      "verifiablePresentation": "..."
```

Whereas in real life it definitely doesn't make sense to pass `results` at the initiation of an exchange.

Perhaps we can sub out the variables schema with a different VariablesForCreate or something like that.

BigBlueHat:
Fixed in #603

PR #603, opened 2026-03-10
Use generic object without results for create exchange variables.

kezike:
This PR addresses Issue #599 by removing the `results` field from the `variables` object in the request body schema of the create exchange endpoint.


msporny (approved):
LGTM, as long as @dlongley's changes requested during the call are included.

Issue #574, opened 2025-12-16
Add normative algorithm for QueryByExample matching


msporny:
The specification needs to detail the normative matching algorithm, which is the algorithm that a digital wallet uses to match the query against the credentials it holds, for QueryByExample.

dlongley:
Some links that might be helpful for producing a matching algorithm:

https://github.com/digitalbazaar/oid4-client/blob/v5.13.2/lib/query/match.js#L6-L23
https://github.com/digitalbazaar/oid4-client/blob/v5.13.2/lib/query/queryByExample.js#L6-L8
https://github.com/digitalbazaar/oid4-client/blob/v5.13.2/lib/query/util.js#L35-L38

kezike:
@bparth24 Do you still want to keep this issue or would you like for me to take it over?

bparth24:
@kezike 

If you have time, feel free to take over. Here is the branch with updates: https://github.com/w3c/vcalm/tree/bparth24-issue-574.

What was pending:

- Mainly double-checking the algorithm written against the actual code links shared above by Dave Longley in the issue.

PR #572, opened 2025-12-16
Add selective disclosure example and conversion pseudo code

PatStLouis:
Addresses #545

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vcalm/pull/572.html" title="Last updated on Dec 16, 2025, 6:53 PM UTC (301c197)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vcalm/572/52372b2...301c197.html" title="Last updated on Dec 16, 2025, 6:53 PM UTC (301c197)">Diff</a>

PR #624, opened 2026-04-14
Add selective disclosure example and conversion pseudo code

### LINKED ISSUE #545: Add selective disclosure example to spec w/ conversion of QueryByExample to JSON pointers
We will want to eventually add an algorithm that can convert a `QueryByExample` "example" to a list of JSON pointers that could look something like this:

```js
import jsonpointer from 'json-pointer';

// first perform matching on `@context`, optionally performing
// JSON-LD transformations as desired, then...

// convert `vprQuery.credentialQuery.example` to JSON pointers, modulo
// `@context` field; any VCDM mandatory pointers can be automatically
// added based on VC version
// (e.g., `/issuer`, `/issuanceDate` for VC 1.x)
const object = {...example};
delete object['@context'];
const pointers = _adjustPointers(Object.keys(jsonpointer.dict(object)));

function _adjustPointers(pointers) {
  // ensure `credentialSubject` is included in any reveal because the VCDM
  // requires it; presume that if it isn't present that the entire credential
  // subject was requested and inform the user of this (as always)
  const hasCredentialSubject = pointers.some(
    pointer => pointers.includes('/credentialSubject/') ||
      pointer.endsWith('/credentialSubject'));
  if(!hasCredentialSubject) {
    pointers = pointers.slice();
    pointers.push('/credentialSubject');
  }

  pointers = _pruneShallowPointers(pointers);

  // make `type` pointers generic
  return pointers.map(pointer => {
    const index = pointer.indexOf('/type/');
    return index === -1 ? pointer : pointer.slice(0, index) + '/type';
  });
}

// gets only the deepest pointers from the given list of pointers, for example,
// `['/a/b', '/a/b/c', '/a/b/c/d']` will be pruned to: `['/a/b/c/d']`
function _pruneShallowPointers(pointers) {
  const deep = [];
  for(const pointer of pointers) {
    let isDeep = true;
    for(const p of pointers) {
      if(pointer.length < p.length && p.startsWith(pointer)) {
        isDeep = false;
        break;
      }
    }
    if(isDeep) {
      deep.push(pointer);
    }
  }
  return deep;
}
```

PatStLouis:
### What changes

Under **Query by example**, this adds a subsection (**Converting QueryByExample to JSON Pointers**) that explains how to turn a QueryByExample `example` into a list of JSON pointers for selective disclosure. It spells out the intended steps (including handling `@context`, mandatory VCDM pointers, `credentialSubject`, pruning to deepest pointers, and normalizing `type` pointers) and includes reference pseudo-code for the conversion and pruning logic.

### Context

- **#545** — tracks adding this material to the spec.
- **#572** — earlier draft with the same intent; this PR supersedes that work on the current `main` line.
- **#574** — CG direction to refine this into a clearer normative algorithm section later; this PR does not claim to complete that issue.

### Review feedback already incorporated (from #572)

- Prose: “it is **necessary** to …” (not “often necessary”).
- Pseudo-code in `<pre>`: use **`&lt;`** for the less-than comparison so the line is not parsed as HTML in viewers that syntax-highlight the file.

Closes #545

Refs #572

### DIFF
--- a/index.html
+++ b/index.html
@@ -1535,7 +1535,7 @@ specify which fields should be revealed. The following algorithm provides a
 standardized way to perform this conversion:
           </p>
 
-          <ol>
+          <ol class="algorithm">
             <li>
 First, perform matching on the `@context` field, optionally performing
 JSON-LD transformations as desired.
--- a/index.html
+++ b/index.html
@@ -1525,6 +1525,124 @@ formats are also acceptable.
           such as <code>bbs-2023</code> or <code>ecdsa-sd-2023</code> in the <code>acceptedCryptosuites</code> array.
         </p>
 
+        <section>
+          <h4>Converting QueryByExample to JSON Pointers</h4>
+
+          <p>
+When implementing [=selective disclosure=] features, it is necessary to
+convert a QueryByExample `example` object into a list of JSON pointers that
+specify which fields should be revealed. The following algorithm provides a
+standardized way to perform this conversion:
+          </p>
+
+          <ol>
+            <li>
+First, perform matching on the `@context` field, optionally performing
+JSON-LD transformations as desired.
+            </li>
+            <li>
+Convert the `example` object to JSON pointers, excluding the `@context` field.
+Any VCDM mandatory pointers can be automatically added based on the VC version
+(e.g., `/issuer`, `/issuanceDate` for VC 1.x).
+            </li>
+            <li>
+Ensure that `credentialSubject` is included in any reveal because the VCDM
+requires it. If it is not present, presume that the entire credential subject
+was requested.
+            </li>
+            <li>
+Prune shallow pointers to keep only the deepest pointers from the list. For
+example, `['/a/b', '/a/b/c', '/a/b/c/d']` will be pruned to `['/a/b/c/d']`.
+            </li>
+            <li>
+Make `type` pointers generic by removing array indices, so `/type/0` becomes
+`/type`.
+            </li>
+          </ol>
+
+          <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1525,6 +1525,124 @@ <h3>Query By Example</h3>
...
+          <h4>Converting QueryByExample to JSON Pointers</h4>
+
+          <p>
+When implementing [=selective disclosure=] features, it is necessary to
+convert a QueryByExample `example` object into a list of JSON pointers that
+specify which fields should be revealed. The following algorithm provides a
+standardized way to perform this conversion:
+          </p>
+
+          <ol>
```
TallTed: Should this be classed `algorithm` and use the algorithm-specific CSS used elsewhere?

TallTed (approved):
lgtm

index.html
```diff
@@ -1525,6 +1525,124 @@ <h3>Query By Example</h3>
...
+Make `type` pointers generic by removing array indices, so `/type/0` becomes
+`/type`.
+            </li>
+          </ol>
+
+          <p>
+The following is a reference implementation of this algorithm:
+          </p>
+
+          <pre class="example" title="Algorithm to convert QueryByExample example to JSON pointers">
```
PatStLouis, line 1567: convert to an actual algorithm (normative), using the algorithm class. Avoid pseudo code in favor of a normative algorithm.

PatStLouis: review: https://w3c.github.io/vc-data-integrity/#add-proof  for example of existing algorithms in respec

PatStLouis:
Closing in favor of #700

PR #636, opened 2026-05-12
Align with VC Data Model 2.0 (context URLs + validFrom/validUntil)

### LINKED ISSUE #635: Align examples and schemas with VC Data Model 2.0 (context + validFrom/validUntil)
## Summary
Several artifacts still referenced VC 1.1-style `@context` URLs (`https://www.w3.org/2018/credentials/v1`) and date properties (`issuanceDate` / `expirationDate`, and `issued` in one status-list example).

## Proposed change
- Use **VCDM 2.0** contexts: `https://www.w3.org/ns/credentials/v2` and `https://www.w3.org/ns/credentials/examples/v2` where appropriate.
- Prefer **`validFrom` / `validUntil`** over VC 1.1 `issuanceDate` / `expirationDate` / `issued` in OpenAPI components, examples, and the workflow narrative in `index.html` (plus transition snapshots).

## Scope
OpenAPI (`oas.yaml`, `components/*.yml`), editor’s draft `index.html`, and `transitions/2026/*/index.html` for consistency with published snapshots.

A PR will reference this issue.

PatStLouis:
Closes #635

## Changes
- **`components/Credential.yml`**: Replace `issuanceDate` / `expirationDate` with `validFrom` / `validUntil` (consistent with `UnsecuredCredential` and VCDM 2.0).
- **`components/VerifiablePresentationRequest.yml`**: Query-by-example example uses v2 `@context` URLs.
- **`oas.yaml`**: Status-list response example uses v2 VC context and `validFrom` instead of `issued`.
- **`index.html`** and **`transitions/2026/*/index.html`**: Workflow example + embedded API docs updated to v2 contexts and `validFrom`; narrative updated (removed VC 1.0 / issue note).

OpenAPI validated with `openapi lint oas.yaml`.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 26, 2026, 7:22 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://www.w3.org/publications/spec-generator/?type=respec&output=html&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fvcalm%2F75288d4d47bcd8d8cb28e91ebfb48c4849dbb61c%2Findex.html%3FisPreview%3Dtrue)

**Error output:**

```
EISDIR: illegal operation on a directory, open 'uploads/eGhACO/'
```

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20w3c/vcalm%23636.)._

</details>


dlongley (approved):

PatStLouis:
Discussed: its been mentioned that we might want to reduce the overly prescriptive data modelling for request  / response bodies when it comes to vcdm data modeling. We want to ensure that readers/implementers of the specification understand that any vcdm version can be used, including credential formats via enveloped VC. vcdm 2.0 is only there as an example but currently has some normative statements within OAS definitions.

We will still merge this as a first step next meeting and open an issue to capture the above.

iherman:
This was discussed during the [vcwg meeting on 26 May 2026](https://w3c.github.io/vc-wg/minutes/2026-05-26-alm.html#baf2).

<details><summary><i>View the transcript</i></summary>

<h4><a href="https://github.com/w3c/vcalm/pull/636" rel="noopener noreferrer">Align with VC Data Model 2.0 (context URLs + validFrom/validUntil) by PatStLouis · Pull Request #636 · w3c/vcalm · GitHub</a></h4>
<p><cite>Patrick_St-Louis:</cite> It's Yes,…</p>
<p><cite>Dave_Longley:</cite> I'm not quite sure how we communicate that. and we don't have to necessarily figure that out right now, but for the same people that might have erroneously thought, I can't use this for 2.0." After we make this change, we'll have 1.1 people perhaps thinking, I can't use it for 1.1." But this spec should allow any of those versions.</p>
<p><cite>Patrick_St-Louis:</cite> I think that makes a lot of sense. So the only thing that worries me is about these normative statements that we have in the response, these ones are fine…</p>
<p><cite>Patrick_St-Louis:</cite> because they don't change from one or another. but I think some maybe it's fine, right?</p>
<p><cite>Patrick_St-Louis:</cite> just to make sure that there's no normative statement and these kind of details that mandates one or the other. U Yes.</p>
<p><cite>Dave_Longley:</cite> Yeah, we're overly prescriptive right now.</p>
<p><cite>Dave_Longley:</cite> I would even say that all of this stuff that's on the screen, this has come up in other conversations where we talked about making this appear smaller so it's easier to digest for people, but really all that needs to be in the verifiable credential field is a valid verifiable credential with a link out to those specs. it's still useful to have JSON schema that people can use, and do stuff with, but we have to find some way to better express</p>
<p><cite>Patrick_St-Louis:</cite> And here you have an ore for this envelope credential. I said the VCOM is interesting because it's kind of an intersection right of many different specs.</p>
<p><cite>Patrick_St-Louis:</cite> we don't really want to do data model thing in VCOM but at the same time we kind of want to do data model but more from a rest API perspective because I don't want to get into redoing a bunch of data model testing in the vcom spec because we already have those extensively tested. so yeah basically I think this is maybe something we'll have to redisco oas field description how specific we want to be and that's all it touches right I think I'm pretty sure it only touches these kind of</p>
<p><cite>Patrick_St-Louis:</cite> little example bodies for credentials and verifiable presentation. yeah.</p>
<p><cite>Patrick_St-Louis:</cite> So, I think maybe a go ahead. Yeah.</p>
<p><cite>Dave_Longley:</cite> Yeah, the other comment I want to make is,…</p>
<p><cite>Dave_Longley:</cite> VCDM 2.1 is going to be out at some point. and we don't want people to think you can't use that. So we might want to even make a few places maybe even make them informative and say you need to put this such as these types future versions of them. We might want to loosen up some of the things that are normative today. I really feel like we need to explore that at some point. because we're duplicating work. We're also creating unnecessary restrictions. and…</p>
<p><cite>Dave_Longley:</cite> we don't want to have to rip to do a revision of this spec just cuz some new data model comes, tweaks are</p>
<p><cite>Patrick_St-Louis:</cite> Yeah, something that could be interesting maybe in the OS we only want to define the context ID and…</p>
<p><cite>Patrick_St-Louis:</cite> type and the rest people can put whatever they want as long as it's JSON LD conformant I know if you want to support nonJSONLD things in there. but I would be fairly confident to say that if there's the VCDM 2.1, 2.2, 2.3, they will all have a context ID and type, right? if this changes, I don't know what I'm going to do, but going to lose my mind. for other things like issuance date, right?</p>
<p><cite>Patrick_St-Louis:</cite> expiration date, even things like issuer, proof, this is all depending on the type, this would cover both enveloped and sort of inline stickering mechanism. so maybe something I can open as an issue following if we merge this PR I think that the PR is good let's update this so that at least it's what we have now is consistent and maybe loosen up the OAS file and remove everything that's kind of derived from the JSONLD stuff and just put words in it that you can use other credential format</p>
<p><cite>Patrick_St-Louis:</cite> I don't know we'll have to probably iron the wording around this. this will also significantly reduce the size of the specification and even this problem response right from valid until Yeah, good thing. I think we're on the same page here. and that. Any other comments regarding this? I'm going to capture this on the issue. I'm not going to merge this today. Maybe I'll leave it open the other issue.</p>
<p><cite>Patrick_St-Louis:</cite> maybe link the other issue here to say there's going to be a part two of deciding if we actually need these extensive example I could see arguments for both right some people sometime it's nice when you have the spec completely mapped out and no ended part because us in this group talking about yeah but you might do 2.1 2 or 1.1 or envelope</p>
<p><cite>Patrick_St-Louis:</cite> we're familiar with these things, but I'm thinking about the poor dev, right, that's being tasked to implement VCOM and he's never heard about verifiable credential or anything. having the spec being very prescriptive and not really saying yeah, you go to this spec, move around. I think there's value in that. so I think finding a solution that's kind of in the middle of this could go a long way.</p>
<p><cite>Patrick_St-Louis:</cite> Any other comments for this? discussed it's been mentioned we might want to reduce the restrictive model for response bodies when it comes to VCDM data modeling.</p>
<p><cite>Patrick_St-Louis:</cite> We want to ensure that there be used both formats.</p>
<p><cite>Patrick_St-Louis:</cite> BCDM 2.0 is only there as an example, but I can't see. Okay, so here's what I wrote. So, it's been mentioned that we might want to reduce the overly perspective data modeling for request response bodies when it comes to VCDM data modeling.</p>
<p><cite>Patrick_St-Louis:</cite> We want to ensure that implementers of the specification understand that any VCDM version can be used as well as other credential formats. VCDM 2.0 is only there as an example but currently has some statements within A's definitions. so I think this sums up a bit what we discussed. any objections any other information we might want to capture here?</p>
<p><cite>Patrick_St-Louis:</cite> Yes, Dave.</p>
<p><cite>Dave_Longley:</cite> We might just want to say instead of saying as well as other credential formats,…</p>
<p><cite>Dave_Longley:</cite> it could be a parenthetical that says this includes other credential formats via enveloped credential because that's still VCDM compliant.</p>
<p><cite>Patrick_St-Louis:</cite> Is it VCDM 1.1 compliant?</p>
<p><cite>Dave_Longley:</cite> Yeah, but we're saying any version of ECDM works and…</p>
<p><cite>Patrick_St-Louis:</cite> Debatable. Yes. Yes.</p>
<p><cite>Dave_Longley:</cite> so it's covered that</p>
<p><cite>Patrick_St-Louis:</cite> So I think there goes something we need to untangle. we will still merge this as a first step and open an issue to capture the above. There we go.</p>

<hr /></details>

TallTed (approved):
Looks OK as a stepping stone. Probably shouldn't merge until the issue for the next steps exists and there are live cross links.

PatStLouis:
Here's an issue to capture the simplification of the OAS file, relating to credentials and presentation payloads. This would be the logical next step.
#653

Issue #629, opened 2026-04-14
Add clarity around `acceptedEnvelopes` vs. `acceptedCryptosuites` usage


kezike:
There have been multiple recent reports by implementers about confusion around the usage of `acceptedEnvelopes` and `acceptedCryptosuites`. The confusion arose from the lack of clarity around what `acceptedEnvelopes` values to use for JSON-LD VCs. The offline discussion led to the confirmation that such VC requests should omit `acceptedEnvelopes`, as it is only relevant for enveloped VCs (e.g., with JWT encoding), although an empty array is also permissible. Additionally, JSON-LD VCs should generally use `acceptedCryptosuites` instead of `acceptedEnvelopes`.

CC @dmitrizagidulin @ottonomy

kezike:
This issues is addressed by PR #651.

Issue #605, opened 2026-03-10
Fix rendering issues that occur whenever `oneOf` is used with `required` constraint


kezike:
As I was writing PR #603, I noticed that there are a few instances where ReSpec is not rendering fields when they are not included in `oneOf[i].required`. Here is an example where `variables` is not rendering:

```
StepData:
  type: object
  additionalProperties: false
  description: Step data to be included if a template is not used.
  properties:
    createChallenge:
      type: boolean
      description: An optional step directive that tells the exchange to handle challenge management via a VCALM verifier service it has a zcap for.
    verifiablePresentationRequest:
      $ref: "./components/VerifiablePresentationRequest.yml#/components/schemas/VerifiablePresentationRequest"
    callback:
      type: object
      description: A callback that will be called after this step executes.
      properties:
        url:
          type: string
          description: A URL that will receive the callback data.
    issueRequests:
      type: array
      items:
        type: object
        properties:
          credentialTemplateId:
            type: string
            description: The ID of the item in workflow.credentialTemplates that captures the schema of a credential that is issued in a step.
          credentialTemplateIndex:
            type: number
            description: The index of the item in workflow.credentialTemplates that captures the schema of a credential that is issued in a step.
          # Optionally specify different variables from exchange.variables
          # (either a string that is the name of a variable in exchange.variables
          # to use, or a whole object with alternative variables)
          variables:
            oneOf:
              - type: string
              - $ref: "#/components/schemas/CreateExchangeVariables"
        oneOf:
          - required:
              - credentialTemplateId
          - required:
              - credentialTemplateIndex
        additionalProperties: false
    presentationSchema:
      oneOf:
        - type: object
          description: A JSON Schema describing the validation to be performed on the presentation.
          properties:
            type:
              type: string
              description: The value MUST be <code>JsonSchema</code>.
            jsonSchema:
              type: object
              description: The JSON Schema object to use when validating the presentation.
        - type: object
          description: An alternate presentation schema format to use when validating a presentation in the current step. Additional properties are expected for a specific presentation schema type, but are beyond the scope of this specification.
          properties:
            type:
              type: string
              description: The type of presentation schema mechanism to use when validating the presentation.
    verifyPresentationResponseSchema:
      oneOf:
        - type: object
          description: A JSON Schema describing the validation to be performed on the result of verifying the presentation.
          properties:
            type:
              type: string
              description: The value MUST be <code>JsonSchema</code>.
            jsonSchema:
              type: object
[...]

PR #651, opened 2026-06-22
Clarify when to use `acceptedEnvelopes` and `acceptedCryptosuites`

kezike:
This PR addresses Issue #629 by providing guidance on when a VPR should include `acceptedEnvelopes` or `acceptedCryptosuites`.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 30, 2026, 11:02 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://www.w3.org/publications/spec-generator/?type=respec&output=html&url=https%3A%2F%2Fraw.githubusercontent.com%2Fkezike%2Fvcalm%2F624b3dd89c8dec98c6653651e5d0c2b2e239b611%2Findex.html%3FisPreview%3Dtrue)

**Error output:**

```
EISDIR: illegal operation on a directory, open 'uploads/CMNNgv/'
```

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20w3c/vcalm%23651.)._

</details>

### DIFF
--- a/components/VerifiablePresentationRequest.yml
+++ b/components/VerifiablePresentationRequest.yml
@@ -47,7 +47,7 @@ components:
                           type: string
                   "acceptedCryptosuites":
                     type: array
-                    description: Cryptographic proof suites the verifier will accept. Each element SHOULD be an object with a `cryptosuite` property; however, for backwards compatibility, string values are also accepted and interpreted as the cryptosuite name for a Data Integrity proof.
+                    description: Cryptographic proof suites the verifier will accept. Each element SHOULD be an object with a <code>cryptosuite</code> property; however, for backwards compatibility, string values are also accepted and interpreted as the cryptosuite name for a Data Integrity proof. This enables the holder to select an appropriate proof format. The presence of this property SHOULD be mutually exclusive with that of <code>acceptedEnvelopes</code>, which is only applicable to enveloped credentials.
                     items:
                       oneOf:
                         - type: string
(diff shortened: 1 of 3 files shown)


components/VerifiablePresentationRequest.yml
kezike, line 1: Prior to raising this PR, I noticed that ReSpec is not rendering the element types of `acceptedEnvelopes` nor `acceptedCryptosuites` properly (e.g., _Each item in the acceptedCryptosuites array MUST be a undefined_ ). This is somewhat related to Issue #605, but I can't tell if it's coming from the same source. It shouldn't block this PR, but I have made [a note of it](https://github.com/w3c/vcalm/issues/605#issuecomment-4770883145) in the existing issue, encouraging the assignee to investigate and address the issue in the same PR.

dlongley (approved):

kezike:
@dlongley suggests that we allow `Ed25519Signature2020` as a `cryptosuite` value.
@dlongley suggests that we should not use "mutually exclusive" and instead specify that the two fields are independent and can be used separately from each other.

index.html
```diff
@@ -1422,9 +1422,11 @@ <h3>Query By Example</h3>
             <td>
 An OPTIONAL array specifying which cryptographic proof suites the [=verifier=]
 will accept. Each element SHOULD be an object with a <code>cryptosuite</code>
-property; however, for backwards compatibility, string values are also accepted
-and interpreted as the cryptosuite name for a Data Integrity proof. This enables
-the [=holder=] to select an appropriate proof format. Valid example values include:
+property referencing a Data Integrity cryptosuite name or <code>Ed25519Signature2020</code> for legacy
```
TallTed, line 1425: "for legacy" *what*? "for legacy support" maybe?

Note that "legacy support" is often synonymous with "backward compatibility", so it's weird to have that following parenthetical note.

Issue #633, opened 2026-04-30
Consider adding `<exchangeId>/request` endpoint to serve current VPR/request for DC API protocol


dlongley:
This endpoint would respond to an HTTP GET w/`accept: application/json` -- and return any initial VPR for the exchange. This might be a useful endpoint for integration with the DC API. To make a DC API request backed by the VCALM protocol, an interaction URL would be sent as the request data. The browser could do an HTTP GET on the interaction URL and get the supported subprotocols there and do an HTTP GET on the vcapi URL + `/request`. Getting this request would not advance the exchange (no moving to new steps/no issuance of VCs, etc.)

This would allow the browser to go through the RP's origin (anchoring the whole operation in the RP's origin), through to whatever delegate they have approved as a service provider (themselves or a third party). The browser could do this whether the request is on the same device or forwarded for a cross device request. It would all work without a need for "signed requests", it would use TLS and existing CA infrastructure, without any need for new CA infrastructure/certs, nor a need to incorporate those into the exchange protocol at all (other than in the usual HTTPS communication).

dlongley:
Additionally, if we find that requests MUST be signed or responses MUST be encrypted per some DC API rules (still in flux), then the request could be returned/negotiated as a signed object and it could include a public key for facilitating encryption for a response that would be provided over DC API (if desired by the responder).

Once a DC API (potentially encrypted) response is received by a coordinator at whatever endpoint it provides to communicate with its frontend, it could optionally decrypt and send the response along to the exchange per usual -- or support for performing that decryption could be added to the exchange protocol. The latter brings the complexity of the DC API to workflow services but would be repeatable / pluggable infrastructure vs. requiring coordinators to implement it.

We can keep this issue open while the DC API rules evolve.

kezike:
Addressed by PR #663

PR #663, opened 2026-07-12
Add support for retrieving current exchange VPR

kezike:
This PR addresses Issue #633 by adding an endpoint to retrieve the current VPR in an exchange, a feature we expect will be needed for DC API and potentially other protocols in the future.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Aug 11, 2026, 9:55 PM UTC)_.

<details>
<summary>More</summary>

**Error output:**

```
EISDIR: illegal operation on a directory, open 'uploads/bt1tqm/'
```

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20w3c/vcalm%23663.)._

</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2828,7 +2828,7 @@ and perform an HTTP GET on the `vcapi` URL concatenated with `/request`.
         <p>
 This feature allows the browser to route their requests through the
 [=coordinator=]'s origin (thereby anchoring the entire operation in said origin),
-through to the delegate they have approved as a service provider.
+to the delegate they have approved as a service provider.
 This convenience exists whether the request is made on the same device or
 forwarded for a cross-device request.
         </p>
--- a/index.html
+++ b/index.html
@@ -2829,7 +2829,7 @@ and perform an HTTP GET on the `vcapi` URL concatenated with `/request`.
 
         <p>
 This feature allows the browser to route their requests through the
-[=coordinator=]'s origin (thereby anchoring the entire operation in said origin),
+[=coordinator's=] origin (thereby anchoring the entire operation in said origin),
 to the delegate they have approved as a service provider.
 This convenience exists whether the request is made on the same device or
 forwarded for a cross-device request.
--- a/index.html
+++ b/index.html
@@ -2806,6 +2806,52 @@ links to resources from other domains to render a web page.
       <section>
         <!--If this header is changed, also update the x-componentTableLink parameter for this endpoint in the oas.yaml file.
             If the x-componentTableLink isn't updated to match the corresponding Component Table link will break.-->
+        <h4>Get Initial Exchange VPR</h4>
+        <p>
+This endpoint MUST return the initial [=verifiable presentation request=] to be returned by an exchange
+when accessed with an `Accept` header value of `application/json`. Processing this
+request SHOULD not advance the exchange to a new step nor perform any
+state-changing side effects, such as credential issuance.
+        </p>
+
+        <div class="api-detail"
+          data-api-endpoint="get /workflows/{localWorkflowId}/exchanges/{localExchangeId}/request"></div>
+
+        <p>
+This might be a useful endpoint for integration with the [[[DCAPI]]].
+To make a [[DCAPI]] request backed by the VCALM protocol, the [=coordinator=]
+MUST provide the [=interaction URL=] as the request data. Subsequently, the
+browser MUST fetch the [=interaction URL=], retrieve the supported protocols,
+and perform an HTTP GET on the `vcapi` URL concatenated with `/request`.
+        </p>
+
+        <p>
+This feature allows the browser to route their requests through the
+[=coordinator=]'s origin (thereby anchoring the entire operation in said origin)
+through to the delegate they have approved as a service provider.
+This convenience exists whether the request is made on the same device or
+forwarded for a cross-device request.
+        </p>
+
+        <p>
+Finally, this endpoint operates without the need for signed requests, as it MAY
+rely on TLS and existing CA infrastructure, without the need for new certificates
+nor a need to incorporate those into the exchange protocol in an additive way.
+However, if we find that the [[DCAPI]] or other supported protocols require
+signed requests or encrypted responses, then the request MAY be returned as a
+signed object referencing a public key that SHOULD be used to encrypt the response.
+Upon receipt of the (potentially encrypted) response, the [=coordinator=] could
+decrypt and deliver the response to the appropriate exchange or delegate decryption
+support to the exchange service altogether. The official decision on which interface
+VCALM prescribes for performing this functionality is in flux as the [[DCAPI]] rules
+continue to evolve.
+        </p>
+
+      </section>
+
+      <section>
+        <!--If this header is changed, also update the x-componentTableLink parameter for this endpoint in the oas.yaml file.
+            If the x-componentTableLink isn't updated to match the corresponding Component Table link will break.-->
         <h4>Participate in an Exchange</h4>
         <p>
 A server MAY include a <code>referenceId</code> property in an exchange
@@ -3276,7 +3322,7 @@ on how to start an interaction with the remote system.
         </p>
 
         <p>
-When the [=interaction URL=] is fetched using an `Accept` header of
+When the [=interaction URL=] is fetched using an `Accept` header value of
 `application/json`, a single JSON object containing a `protocols` [=map=] MUST
 be returned where each [=map/key=] is a protocol identifier and each
 [=map/value=] is a URL that can be used to initiate the interaction. For
@@ -3313,7 +3359,7 @@ the `saas.example` domain in the list of protocols.
         </p>
 
         <p>
-When the [=interaction URL=] is fetched using any unrecognized `Accept` header,
+When the [=interaction URL=] is fetched using any unrecognized `Accept` header value,
 a `text/html` document MUST be returned with directions instructing a human
 being to use specific software that understands how to process interaction URLs.
         </p>
--- a/index.html
+++ b/index.html
@@ -2806,9 +2806,9 @@ links to resources from other domains to render a web page.
       <section>
         <!--If this header is changed, also update the x-componentTableLink parameter for this endpoint in the oas.yaml file.
             If the x-componentTableLink isn't updated to match the corresponding Component Table link will break.-->
-        <h4>Get Initial Exchange VPR</h4>
+        <h4>Get Current Exchange VPR</h4>
         <p>
-This endpoint MUST return the initial [=verifiable presentation request=] to be returned by an exchange
+This endpoint MUST return the current [=verifiable presentation request=] to be returned by an exchange
 when accessed with an `Accept` header value of `application/json`. Processing this
 request SHOULD not advance the exchange to a new step nor perform any
 state-changing side effects, such as credential issuance.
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -2598,6 +2598,52 @@ <h4>Get Exchange Protocols</h4>
 
       </section>
 
+      <section>
+        <!--If this header is changed, also update the x-componentTableLink parameter for this endpoint in the oas.yaml file.
+            If the x-componentTableLink isn't updated to match the corresponding Component Table link will break.-->
+        <h4>Get Initial Exchange VPR</h4>
```
kezike: @dlongley When we discussed this on a recent VCALM TF call, you alluded to the possibility that this endpoint might return the current VPR (not necessarily the initial one). How do we feel about this now?

oas.yaml
```diff
@@ -346,6 +346,35 @@ paths:
...
+      summary: Gets the initial VPR for a specific exchange.
+      tags:
+        - Exchanges
+      security:
+        - networkAuth: []
+        - oAuth2: []
+        - zCap: []
+      operationId: getInitialExchangeVpr
+      description: Gets the initial VPR for a specific exchange. This might be a useful endpoint for integration with the DC API. Processing this request SHOULD not advance the exchange to a new step nor perform any state-changing side effects, such as credential issuance.
+      x-expectedCaller: Anyone with provided URL
```
kezike, line 361: @dlongley Do we expect this to be a public endpoint, since we expect the DC API browser frontend to call it?

oas.yaml
```diff
@@ -346,6 +346,35 @@ paths:
...
+        "200":
+          description: Initial VPR to be returned by the exchange.
+          content:
+            application/json:
+              schema:
+                $ref: "./components/VerifiablePresentationRequest.yml#/components/schemas/VerifiablePresentationRequest"
+        "400":
+          description: Invalid input
+        "401":
+          description: Not Authorized
```
kezike: Are these response statuses necessary for this endpoint?

oas.yaml
dlongley: Yes, it is "public" in the same sense that anyone with the capability URL can retrieve it (just like anyone with the capability URL can participate in the exchange).

oas.yaml
dlongley: Not based on their descriptions.

index.html
dlongley: I think it would be good to return the current VPR, yes. For the browser use case (or any "initiating use case"), it would only care about the initial one. But there might be other uses for clients that are stepping through exchanges to just ask for the next VPR.

kezike: OK, addressed in a recent commit.

PatStLouis (approved):
LGTM

PatStLouis:
discussed: The group decided that this might be preferred as an optional endpoint with normative instructions if implemented. Leaving it as REQUIRED for the time being, we may revisit and make it optional.

oas.yaml
```diff
@@ -346,6 +346,31 @@ paths:
...
+      parameters:
+        - $ref: "./components/parameters/path/LocalWorkflowId.yml"
+        - $ref: "./components/parameters/path/LocalExchangeId.yml"
+      responses:
+        "200":
+          description: Current VPR to be returned by the exchange.
+          content:
+            application/json:
+              schema:
+                $ref: "./components/VerifiablePresentationRequest.yml#/components/schemas/VerifiablePresentationRequest"
```
dlongley: I think this should return `{verifiablePresentationRequest}`. This also allows us to augment this information in standard and non-standard (incubation) ways. It might be useful to add other meta information like announcing what VC types might be issued if you provide a VP that satisfies the VPR.

We might find that we want to (optionally) put that information directly inside a VPR, which means we wouldn't need another top-level field here, but it still seems more future proof to do it this way for something else.

index.html
```diff
@@ -2598,6 +2598,52 @@ <h4>Get Exchange Protocols</h4>
...
       </section>
 
+      <section>
+        <!--If this header is changed, also update the x-componentTableLink parameter for this endpoint in the oas.yaml file.
+            If the x-componentTableLink isn't updated to match the corresponding Component Table link will break.-->
+        <h4>Get Current Exchange VPR</h4>
+        <p>
+This endpoint MUST return the current [=verifiable presentation request=] to be returned by an exchange
+when accessed with an `Accept` header value of `application/json`. Processing this
+request SHOULD not advance the exchange to a new step nor perform any
```
```suggestion
request SHOULD NOT advance the exchange to a new step nor perform any
```

index.html
```diff
@@ -2598,6 +2598,52 @@ <h4>Get Exchange Protocols</h4>
...
+This might be a useful endpoint for integration with the [[[DCAPI]]].
+To make a [[DCAPI]] request backed by the VCALM protocol, the [=coordinator=]
+MUST provide the [=interaction URL=] as the request data. Subsequently, the
+browser MUST fetch the [=interaction URL=], retrieve the supported protocols,
+and perform an HTTP GET on the `vcapi` URL concatenated with `/request`.
+        </p>
+
+        <p>
+This feature allows the browser to route their requests through the
+[=coordinator=]'s origin (thereby anchoring the entire operation in said origin)
```
```suggestion
[=coordinator=]'s origin (thereby anchoring the entire operation in said origin),
```

index.html
```diff
@@ -2598,6 +2598,52 @@ <h4>Get Exchange Protocols</h4>
...
+through to the delegate they have approved as a service provider.
+This convenience exists whether the request is made on the same device or
+forwarded for a cross-device request.
+        </p>
+
+        <p>
+Finally, this endpoint operates without the need for signed requests, as it MAY
+rely on TLS and existing CA infrastructure, without the need for new certificates
```
```suggestion
rely on TLS and existing CA infrastructure, without a need for new certificates
```

TallTed (changes requested):
lgtm with minor tweaks

oas.yaml
kezike: Yes, good catch. Updated in 861f18c.

kezike: I have a few more questions regarding the response to this endpoint:
1. In the event that an exchange never returns a VPR (e.g., possession of the exchange URL is enough to earn a VP or redirect to another site or interaction URL, without fulfilling an additional request as a prerequisite), should this be an empty object, null, omitted, or something else?
2. In the event that a VP is delivered in the first step, should that be returned in a separate `verifiablePresentation` property? If so, should that also accept both object and array formats, per the update included in #666?

dlongley (approved):

oas.yaml
dlongley: > In the event that an exchange never returns a VPR (e.g., possession of the exchange URL is enough to earn a VP or redirect to another site or interaction URL, without fulfilling an additional request as a prerequisite), should this be an empty object, null, omitted, or something else?

I'd think this should be represented, I think, as an empty VPR: `{verifiablePresentationRequest: {}}` (an empty object would signal the exchange is terminated and an empty VPR has been used in implementations to signal nothing specific is requested but the exchange continues).

> In the event that a VP is delivered in the first step, should that be returned in a separate `verifiablePresentation` property?

No, because the purpose of this endpoint is to only advertise what will be asked for -- not provide anything (like VCs) to the caller. As mentioned in a comment above, we could advertise metadata about what will be provided if the VPR is fulfilled, but that is perhaps best added to the VPR itself, so we don't need to worry about that here.

PatStLouis:
discussed: need to address new comments about empty VPR

index.html
```diff
@@ -2598,6 +2598,52 @@ <h4>Get Exchange Protocols</h4>
...
+To make a [[DCAPI]] request backed by the VCALM protocol, the [=coordinator=]
+MUST provide the [=interaction URL=] as the request data. Subsequently, the
+browser MUST fetch the [=interaction URL=], retrieve the supported protocols,
+and perform an HTTP GET on the `vcapi` URL concatenated with `/request`.
+        </p>
+
+        <p>
+This feature allows the browser to route their requests through the
+[=coordinator=]'s origin (thereby anchoring the entire operation in said origin),
+through to the delegate they have approved as a service provider.
```
```suggestion
to the delegate they have approved as a service provider.
```

oas.yaml
kezike: > I'd think this should be represented, I think, as an empty VPR

Hmm, I have seen empty exchange response messages, but not empty VPR responses. But technically the `verifiablePresentationRequest` field references the `VerifiablePresentationRequest` schema, which does not have any `required` properties, so I suppose it's allowed.

> No, because the purpose of this endpoint is to only advertise what will be asked for...As mentioned in a comment above, we could advertise metadata about what will be provided if the VPR is fulfilled, but that is perhaps best added to the VPR itself...

Coincidentally, this seems to be addressed by the `credentialOffer` property introduced in Issue #689 that you recently opened.

index.html
```diff
@@ -2606,7 +2606,9 @@ <h4>Get Current Exchange VPR</h4>
 This endpoint MUST return the current [=verifiable presentation request=] to be returned by an exchange
 when accessed with an `Accept` header value of `application/json`. Processing this
 request SHOULD NOT advance the exchange to a new step nor perform any
-state-changing side effects, such as credential issuance.
+state-changing side effects, such as credential issuance. If there is no [=verifiable presentation request=]
+available for the exchange in the current step, the value of the `verifiablePresentationRequest` property
+returned by this endpoint MUST be an empty JSON object.
```
kezike: @dlongley Per [your comment here](https://github.com/w3c/vcalm/pull/663#discussion_r3616081103), should this be a `SHOULD` or is `MUST` fine here?

TallTed (approved):

index.html
```diff
@@ -2803,6 +2803,54 @@ <h4>Get Exchange Protocols</h4>
...
+This might be a useful endpoint for integration with the [[[DCAPI]]].
+To make a [[DCAPI]] request backed by the VCALM protocol, the [=coordinator=]
+MUST provide the [=interaction URL=] as the request data. Subsequently, the
+browser MUST fetch the [=interaction URL=], retrieve the supported protocols,
+and perform an HTTP GET on the `vcapi` URL concatenated with `/request`.
+        </p>
+
+        <p>
+This feature allows the browser to route their requests through the
+[=coordinator=]'s origin (thereby anchoring the entire operation in said origin),
```
```suggestion
[=coordinator's=] origin (thereby anchoring the entire operation in said origin),
```

TallTed (commented):
lgtm, modulo minor tweaks

index.html
kezike: Hmm, I have seen it both ways in the spec. For example, there are currently 6 other instances of this in the spec. I am surprised that ReSpec is able to do this kind of fuzzy matching, since these are references to `<dfn></dfn>` elements defined elsewhere in the spec. Do you suggest that I make this update for all instances of `=]s`, `=]'s`, etc. or should we just leave it as is everywhere?

TallTed: I find the prettified `'s=]`, `s=]`, etc., read better. This is part of why the ReSpec authors worked hard on fuzzy matching of (regular) [plurals](https://github.com/speced/respec/wiki/pluralize), [possessives](https://github.com/search?q=repo%3Aspeced%2Frespec+possessive&type=code), and similar. (Irregular constructs may require [manual aliasing](https://github.com/speced/respec/wiki/data-lt) in the `<dfn data-lt="...">`.) That said, it's up to the group.

Issue #631, opened 2026-04-22
Mermaid diagrams for interactions are incorrect


dlongley:
In https://w3c.github.io/vcalm/#initiating-interactions, the part of these flows where the exchange happens (e.g., "initiate workflow exchange") should show the holder coordinator communicating with a workflow service instead of with the issuer/verifier coordinator.

kezike:
Addressed by PR #665 ✅

PR #665, opened 2026-07-12
Fix exchange interaction diagrams

kezike:
This PR addresses Issue #631 by fixing the exchange interaction diagrams such that the holder coordinator engages with the workflow service for the exchange, instead of the coordinator.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jul 14, 2026, 12:51 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://www.w3.org/publications/spec-generator/?type=respec&output=html&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fvcalm%2F0c9107780845849348fec67af207505a229982b9%2Findex.html%3FisPreview%3Dtrue)

**Error output:**

```
EISDIR: illegal operation on a directory, open 'uploads/Z8DKxB/'
```

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20w3c/vcalm%23665.)._

</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2861,6 +2861,9 @@ sequenceDiagram
   autonumber
   H->>B: Click "Receive credential"
   B->>I: Request interaction QR Code
+  I->>WS: Create workflow exchange
+  Note left of WS: POST https://workflow.example/workflows/abc/exchanges
+  WS->>I: Return workflow exchange
   I->>I: Generate QR Code
   I->>B: Display QR Code
   Note left of I: https://issuer.example/interactions/123?iuv=1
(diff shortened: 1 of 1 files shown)


index.html
kezike, line 1: Should I also add the following details to these diagrams or are they best left out for simplicity:

1. A connection between the coordinator and the workflow service for the QR code generation (in order to illustrate that this involves creating an exchange in the backend)
2. A connection between the coordinator and the workflow service for the interaction protocol retrieval (in order to illustrate that this involves fetching the protocols object from the backend)

dlongley: I think item one should happen (coordinator hitting the workflow service to create the exchange prior to generating the QR code for it). I don't think we should do item two because it doesn't necessarily have to be done that way and is a level of detail that's not needed here, I don't think.

Otherwise, this looks good (as best as I can tell from just reading the mermaid as text).

kezike: Addressed in recent commit.

kezike: If it helps review, this is what the issuer diagram looks like now (the verifier one is essentially the same):

VCALM Issuer Initiated Exchange Interaction

PatStLouis (approved):
LGTM

dlongley (approved):

TallTed (approved):
lgtm

Issue #649, opened 2026-06-16
Allow multiple presentations to be sent in a response via an array value for `verifiablePresentation`


dlongley:

kezike:
Since this changes the response schema in an important way, it would be nice to get the reasoning behind this proposal, especially as we are classifying issues in preparation for pre-CR horizontal review.

dlongley:
This is in support of being able to submit different VPs for the same exchange, potentially enabling simpler UX for users that might need to provide different VP formats (e.g., one or more Enveloped VPs and a non-enveloped VP) in a single flow, because of format restrictions that are outside of the purview of VCALM.

A single request might need to ask for a digital credential that requires a presentation format that is different or that should not or cannot be signed at the same time or using the same mechanism as another digital credential (and other presentation format). It's unclear how many of these use cases might develop over time, but it could be the case that, for example, the DC API behaves in this fashion in the future -- and providing this affordance could make VCALM use easier (or it might be required to make it possible based on decisions made elsewhere).

At the very least, we would not want the existing spec to prevent experimentation or incubation in this direction, which means, I think, some kind of call out -- which may as well be specification of a simple "there maybe more than one `verifiablePresentation` provided in response". Of course, specific exchanges can refuse this because they didn't ask or don't permit more than one.

kezike:
That is useful context. I have follow-up questions:

1. Would you expect that both parties of the exchange interaction would need this? If so, would the holder potentially need to be able to unpack multiple VCs from multiple VPs?
2. Is this a feature that could potentially be delivered with the existing schema via multiple workflow steps that each expect a single VP or do you anticipate that other protocols, like DC API, might expect a multi-VP response from a single query?
3. The way I see it, this feature would not impact existing workflow implementations, since they are more likely the ones to control whether the `verifiablePresentation` value is an array or an object based on the protocol they are using for credential requests. Is this true or do you envision a scenario where a wallet can send an array-valued `verifiablePresentation` to a workflow service that does not yet support it? This would give me a sense of which components would be most impacted by this update. 

Before we make a decision about whether to include this in version 1.0, I would be interested in hearing thoughts from other VCALM workflow implementers (e.g., @ottonomy, @dmitrizagidulin, @PatStLouis) about whether this feature is worth the effort. I say this because I expect that at least a few (if not all?) of our implementations would need to add support for this in time for CR. But your response to question 3 might clear this up for me.

kezike:
[...]

Issue #664, opened 2026-07-12
No way for an exchange participant to decline/terminate an exchange


dmitrizagidulin:
After an exchange is created, a participant's only operations are reading its state (GET) and continuing it (POST a message). There is no way to signal "I will not complete this exchange" — no DELETE, no decline message. `ProblemDetails` only covers servers reporting errors to clients.

So when a wallet user cancels a presentation request (or holds nothing that satisfies it), the wallet can only abandon the exchange silently, and the coordinator's exchange sits pending until expires. The relying party cannot distinguish a definitive decline from a crashed wallet or a user still deciding.

OpenID4VP defines a refusal (`error=access_denied` to the `response_uri`); the native VC API protocol has no equivalent.

Suggestion: define participant-initiated termination — either a decline message POSTed to the exchange URL, or DELETE on the exchange, transitioning it to a terminal state.

Related: #637 (error handling for workflows) covers server-side errors, not this.

kezike:
I agree that this is a glaring issue. Not sure if you will be on the next call due to Badge Summit and JFF Horizons, but we will take your suggestions as input when we discuss this issue on the next call. In the meantime, I have applied the `v1.0` label to this issue.

kezike:
Per [my comment](https://github.com/w3c/vcalm/pull/666#discussion_r3610753874) in PR #666, the solution to this issue should apply to client-initiated and server-initiated exchange termination. At the moment, server-initiated termination is signaled by an empty object, which could be better.

dlongley:
Perhaps we should add `"done": true` as an explicit key+value that can be included by client or server to signal that the exchange is terminated by their end.

PR #666, opened 2026-07-12
Add support for multiple VPs in exchange participation messages

kezike:
This PR addresses issue #649 by adding support for multiple VPs in exchange participation messages.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jul 19, 2026, 2:17 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://www.w3.org/publications/spec-generator/?type=respec&output=html&url=https%3A%2F%2Fraw.githubusercontent.com%2Fkezike%2Fvcalm%2F8b852bc77f4a2a05614de8560a4037b31608fbd5%2Findex.html%3FisPreview%3Dtrue)

**Error output:**

```
EISDIR: illegal operation on a directory, open 'uploads/w9P51K/'
```

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20w3c/vcalm%23666.)._

</details>

### DIFF
--- a/oas.yaml
+++ b/oas.yaml
@@ -281,9 +281,10 @@ paths:
     post:
       summary: Participate in an exchange. Posting an empty body will start the exchange or return
         what the exchange is expecting to complete the next step. Posting a Verifiable Presentation Request
-        will result in either a 4xx error or one of the following - a Verifiable Presentation
-        or Verifiable Presentation Request that conforms to the client's request. If a Verifiable Presentation
-        is sent, an additional Verifiable Presentation Request may also be sent to continue the exchange.
+        will result in either a 4xx error or one of the following - 1. one or more Verifiable Presentations
+        or 2. a Verifiable Presentation Request that conforms to the client's request. If one or more
+        Verifiable Presentations are sent, an additional Verifiable Presentation Request may also be sent
+        to continue the exchange.
       tags:
         - Credentials
       security:
@@ -1026,7 +1027,7 @@ components:
               $ref: "./components/VerifiablePresentationRequest.yml#/components/schemas/VerifiablePresentationRequest"
             referenceId:
               type: string
-              description: When the server previously sent a referenceId, the client SHOULD include it here to aid debugging and message correlation. The value SHOULD be a urn:uuid value.
+              description: When the server previously sent a referenceId, the client SHOULD include it here to aid debugging and message correlation. The value SHOULD be a <code>urn:uuid</code> value.
         - type: object
           required:
             - verifiablePresentation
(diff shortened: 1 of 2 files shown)


oas.yaml
```diff
@@ -1077,13 +1102,13 @@ components:
...
-              description: An optional identifier to correlate exchange messages. A server MAY include this value; if present, the client SHOULD include it in its next message. The value SHOULD be a urn:uuid value.
+              description: An optional identifier to correlate exchange messages. A server MAY include this value; if present, the client SHOULD include it in its next message. The value SHOULD be a <code>urn:uuid</code> value.
         - type: object
           description: An object empty of exchange continuation properties indicating the server wishes to terminate the exchange.
           properties:
             referenceId:
               type: string
-              description: An optional identifier to correlate exchange messages. A server MAY include this value; if present, the client SHOULD include it in its next message, if there is a next message to send. The value SHOULD be a urn:uuid value.
+              description: An optional identifier to correlate exchange messages. A server MAY include this value; if present, the client SHOULD include it in its next message, if there is a next message to send. The value SHOULD be a <code>urn:uuid</code> value.
```
kezike: I don't know if this option belongs here. The empty object case should probably be represented the same way in the server exchange message as it is in the client message:

```
- type: object
   description: An empty object to initiate the exchange or request what the exchange is expecting to complete the next step.
   properties: {}
   additionalProperties: false
```

dlongley: Maybe, though there is a little asymmetry because the client always starts the exchange, so the server couldn't be initiating anything. I do think, via addressing another issue (#664), we might also get some benefit from having something to include in the object that means an explicit "i'm done with this exchange".

For backwards compatibility, an empty exchange message from a client should still mean "I'm initiating the exchange" and an empty exchange message from a server should still mean "I'm done with the exchange", but enabling both sides to be explicit about terminating the exchange might be nice.

dlongley (approved):

oas.yaml
kezike: Yeah, the part I was confused about is whether there MUST be a `referenceId` included in the server response. Currently, the "empty" case for the server response is really an object with `referenceId` in it.

dlongley: No, there shouldn't be a MUST on sending `referenceId`. However, I do agree that "empty object" actually isn't even the right condition for signaling "the exchange is complete" from the server, it's "no `verifiablePresentationRequest`" (i.e., the server saying "i'm not asking/won't accept anything more"). This can look like an empty object, but it doesn't have to be. So we should correct that.

Once again, having an explicit signal that the exchange is terminated would also help here so we could extend in the future (if we wanted) by having messages that continue the exchange in some new way without needing any VPR.

PatStLouis (approved):
LGTM

TallTed (approved):
lgtm

oas.yaml
kezike: For now, I have brought the server's behavior to parity with the client's by using an empty object to signal exchange termination (8b852bc). At the moment, I think that signal, along with the HTTP status code, is reliable enough. Perhaps, in the future we could add an extra field to the exchange participation body, like `state` or `status`, but I would like to avoid adding new schema properties at this time.

kezike: Now that I am thinking about it, this is related to Issue #664 about client-initiated exchange termination. Since I have brought that in scope for version 1.0, I would be okay using a symmetrical approach for the server (even it includes adding a new property). This is not a blocker for this PR, but I have left [a note](https://github.com/w3c/vcalm/issues/664#issuecomment-5016140953) in the issue to suggest that the solution should address both client-initiated termination and server-initiated termination.

Issue #642, opened 2026-05-29
Add `protocols` object key table


BigBlueHat:
The spec currently explains `inviteRequest` and `vcapi`. The example in https://w3c.github.io/vcalm/#interaction-protocols-response point to the existence of an `OID4VP` key, but there's no spec text about it. Presumably there's also an `OID4VCI` key. The community will certainly want to know if there are (or how to add) endpoints for DIDComm, etc.

BigBlueHat:
There is a list of keys in https://w3c.github.io/vcalm/#get-exchange-protocols so "lifting" that into this other section plus adding information on how extensible it is and how to extend it would be helpful.

PatStLouis:
discussed: lets start with a table in the spec to cover vcalm (vcapi), oid4,  didcomm, inviteRequest, http (see what is in the examples)

eric-schuh:
I just wanted to note that the question of "vcalm" or "vcapi" when discussing the protocols object came up in the threat model work as well: https://github.com/w3c/vcalm/issues/677

kezike:
@eric-schuh 

> I just wanted to note that the question of "vcalm" or "vcapi" when discussing the protocols object came up in the threat model work as well: [#677](https://github.com/w3c/vcalm/issues/677)

Just in case you weren't in the few calls where this has come up, as much as I would have loved to migrate to the `vcalm` protocol key, we have decided to keep `vcapi` for backwards compatibility.

kezike:
Closing, as addressed by PR #680

Issue #537, opened 2023-03-22
Section 3.2 should refer to OID4


Sakurann:
[Section 3.2](https://w3c-ccg.github.io/vp-request-spec/#oidc-credential-provider) should probably be updated to[ OpenID for Verifiable Credentials Issuance Draft](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html)

dlongley:
This open source library now has some conversion code for QueryByExample <=> JSON pointers <=> DCQL: https://github.com/digitalbazaar/oid4-client ... so we might want to crib from it.

kezike:
I will try to address this issue after I address Issue #585, since it is another request for examples.

PR #680, opened 2026-07-19
Add interaction protocol table

kezike:
This PR addresses Issue #642 by adding an interaction protocol registry.

***
<a href="https://pr-preview.s3.amazonaws.com/kezike/vcalm/pull/680.html" title="Last updated on Aug 13, 2026, 2:45 AM UTC (5cd72d6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vcalm/680/11fcd37...kezike:5cd72d6.html" title="Last updated on Aug 13, 2026, 2:45 AM UTC (5cd72d6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2267,11 +2267,11 @@ The exchange client is expected to initiate the exchange using a protocol that
 is compatible with how the client received the exchange URL. For example, the
 exchange URL could have been provided over CHAPI with a protocol identifier
 indicating that this API protocol ought to be used. Alternatively, the
-exchange URL could be included as the "credential_issuer" in an OID4VCI
-credential offer, or as the "client_id" of an OID4VP authorization request,
+exchange URL could be included as the `credential_issuer` in an OID4VCI
+credential offer, or as the `client_id` of an OID4VP authorization request,
 indicating that OID4VCI or OID4VP, respectively, ought to be used. This section
 focuses on how an exchange client uses this API to interact with the exchange; see
-<span class="issue">Appendix TBD</span> to see how to combine exchanges
+Section [[[#interaction-protocols-response]]] to see how to combine exchanges
 with other protocols such as OID4VCI, OID4VP, and DIDComm.
       </p>
       <p>
--- a/index.html
+++ b/index.html
@@ -3371,6 +3371,62 @@ a `text/html` document MUST be returned with directions instructing a human
 being to use specific software that understands how to process [=interaction URLs=].
         </p>
 
+        <p>
+Below is a registry of interaction protocols that are supported today.
+There is broad interest in extracting this registry into a separate
+document as a dependency of this specification. The reasoning behind
+this interest is that the idea of a generic, extensible, reference
+of credential management protocols has a natural appeal across various
+communities and use cases. In the event that this migration occurs,
+implementaters SHOULD use that reference as the authoritative guide
+for protocol usage. In the event that there is a conflict between the
+registry listed below and the external registry, implementers SHOULD
+be aware of versioning differences of this specification and proceed
+with caution.
+        </p>
+
+        <table class="simple">
+          <thead>
+            <tr>
+              <th>Property</th>
+              <th>Usage</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td>`inviteRequest`</td>
+              <td>
+The `inviteRequest` interaction protocol is used by a local system to signal to
+the remote system that it would like an invitation to a remote system via a
+specific URL, such as a website where an individual can engage in a use-case
+specific interaction. More details are provided in Section [[[#inviterequest-interaction-protocol]]].
+              </td>
+            </tr>
+            <tr>
+              <td>`vcapi`</td>
+              <td>
+The `vcapi` interaction protocol is used to initiate a specific [=exchange=]
+as described in Section [[[#participate-in-an-exchange]]].
+More details are provided in Section [[[#vcapi-interaction-protocol]]].
+              </td>
+            </tr>
+            <tr>
+              <td>`OID4VCI`</td>
+              <td>
+The `OID4VCI` interaction protocol is used to initiate a credential issuance
+flow in accordance with the latest version of [[OID4VCI]].
+              </td>
+            </tr>
+            <tr>
+              <td>`OID4VP`</td>
+              <td>
+The `OID4VP` interaction protocol is used to initiate a credential presentation
+flow in accordance with the latest version of [[OID4VP]].
+              </td>
+            </tr>
+          </tbody>
+        </table>
+
         <p class="note" title="Mapping to exchange URLs">
 Some coordinator implementations will implement the protocols endpoint as a pass
 through to a protocols endpoint for an exchange instance. For example, a GET on
--- a/index.html
+++ b/index.html
@@ -3421,33 +3421,41 @@ More details can be found in Section [[[#vcapi-interaction-protocol]]].
             <tr>
               <td>`OID4VCI`</td>
               <td>
+                <p>
 The `OID4VCI` interaction protocol is used to initiate a credential issuance
 flow in accordance with the latest version of [[OPENID4VCI]].
 The value of this entry should be the Authorization Endpoint, against which
 the wallet should perform an Authorization Request for the exchange.
-More details can be found in [[[OPENID4VCI]]]. Implementers should be aware of
-the drift that has occurred with this specification and the impact that this
-has made on the semantics of this property. Specifically, due to the material
-differences between popular draft versions and the stable 1.0 version, it is
-difficult for a wallet to know how to process an inbound request without
-coordination in advance. As a result of this, implementers should consider
-using the versioned variant of this property as described in the protocol versioning note below.
+More details can be found in [[[OPENID4VCI]]].
+                </p>
+                <p>
+Implementers should be aware of the drift that has occurred with this specification
+and the impact that this has made on the semantics of this property.
+Specifically, due to the material differences between popular draft versions and the
+stable 1.0 version, it is difficult for a wallet to know how to process an inbound request
+without coordination in advance. As a result of this, implementers should consider using
+the versioned variant of this property as described in the protocol versioning note below.
+                </p>
               </td>
             </tr>
             <tr>
               <td>`OID4VP`</td>
               <td>
+                <p>
 The `OID4VP` interaction protocol is used to initiate a credential presentation
 flow in accordance with the latest version of [[OPENID4VP]].
 The value of this entry should be the Authorization Endpoint, against which
 the wallet should perform an Authorization Request for the exchange.
-More details can be found in [[[OPENID4VP]]]. Implementers should be aware of
-the drift that has occurred with this specification and the impact that this
-has made on the semantics of this property. Specifically, due to the material
-differences between popular draft versions and the stable 1.0 version, it is
-difficult for a wallet to know how to process an inbound request without
-coordination in advance. As a result of this, implementers might consider
-using the versioned variant of this property as described in the protocol versioning note below.
+More details can be found in [[[OPENID4VP]]].
+                </p>
+                <p>
+Implementers should be aware of the drift that has occurred with this specification
+and the impact that this has made on the semantics of this property.
+Specifically, due to the material differences between popular draft versions and the
+stable 1.0 version, it is difficult for a wallet to know how to process an inbound request
+without coordination in advance. As a result of this, implementers might consider using
+the versioned variant of this property as described in the protocol versioning note below.
+                </p>
               </td>
             </tr>
             <tr>
--- a/index.html
+++ b/index.html
@@ -77,6 +77,12 @@
         status: "WD",
         publisher: "W3C"
       },
+      "DIDCOMM": {
+        title: "DIDComm Messaging",
+        href: "https://identity.foundation/didcomm-messaging/spec/",
+        status: "Editor's Draft",
+        publisher: "DIF"
+      },
       "ZCAP": {
         title: "Authorization Capabilities for Linked Data v0.3",
         href: "https://w3c-ccg.github.io/zcap-spec/",
@@ -3397,9 +3403,10 @@ with caution.
               <td>`inviteRequest`</td>
               <td>
 The `inviteRequest` interaction protocol is used by a local system to signal to
-the remote system that it would like an invitation to a remote system via a
-specific URL, such as a website where an individual can engage in a use-case
-specific interaction. More details are provided in Section [[[#inviterequest-interaction-protocol]]].
+the remote system that it would like an invitation to a remote system.
+The value of this entry should be the URL of a website where an individual can
+engage in a use-case-specific interaction.
+More details can be found in Section [[[#inviterequest-interaction-protocol]]].
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -3303,6 +3303,62 @@ <h4>Interaction Protocols Response</h4>
...
+of credential management protocols has a natural appeal across various
+communities and use cases. In the event that this migration occurs,
+implementaters SHOULD use that reference as the authoritative guide
+for protocol usage. In the event that there is a conflict between the
+registry listed below and the external registry, implementers SHOULD
+be aware of versioning differences of this specification and proceed
+with caution.
+        </p>
+
+        <table class="simple">
```
kezike, line 3394: Outstanding questions regarding this registry:
1. Should we add an entry for DIDComm?
2. There was recent mention that DC API is not an appropriate protocol to include here because it is captured at an earlier context in the browser. Could someone confirm this for me?
3. Are there any other protocols that we should include in this registry in version 1.0?
4. For the `[[OID4VCI]]` and `[[OID4VP]]` references, these point to the listed normative references, which are locked to specific versions. However, the expectation as of now is that the keys should always point to the latest version of the protocol. Is there a way to reference a resource/document without tying it to a specific version (e.g., the way that https://www.w3.org/TR/vc-data-model/ always points to the latest published recommendation for VCDM)?
5. Should we add a mechanism for implementers to provide versioned protocols in addition to the latest protocol properties currently specified. For example, an implementer that wants to use version 1.1 of `aGoodProtocol` might provide a property like `aGoodProtocol_1_1`/`aGoodProtocol-1_1`/`aGoodProtocol_1.1`/`aGoodProtocol-1.1`/`\*something_else\*`? This would address the concern raised by @BigBlueHat on our last call regarding the ambiguity of protocol versioning. It would place the onus on implementers to know that they should either 1) update their implementation to the latest version if they want to use the generic keys or 2) update their keys to use the versioned keys if they are not ready to take on the development cost of a version upgrade, but this may be the best way to avoid ambiguity. Open to opinions/rejections/modifications on this approach.

kezike: Additionally, I noticed the `mdoc-openid4vp` protocol in Issue #537. Does that need to be added as well?

index.html
```diff
@@ -3303,6 +3303,62 @@ <h4>Interaction Protocols Response</h4>
 being to use specific software that understands how to process interaction URLs.
         </p>
 
+        <p>
+Below is a registry of interaction protocols that are supported today.
+There is broad interest in extracting this registry into a separate
```
dlongley: I think we should avoid using the term "registry" here -- this is just a table expressing protocols known to be in use today and there isn't any formal registry or registration process (and no gating). Some kind of registry (formal or information) or "directory" is what might be created later as mentioned.

```suggestion
Below is a table of interaction protocols that are supported today.
There is broad interest in extracting this table into a separate
```

index.html
```diff
@@ -3303,6 +3303,62 @@ <h4>Interaction Protocols Response</h4>
...
+document as a dependency of this specification. The reasoning behind
+this interest is that the idea of a generic, extensible, reference
+of credential management protocols has a natural appeal across various
+communities and use cases. In the event that this migration occurs,
+implementaters SHOULD use that reference as the authoritative guide
+for protocol usage. In the event that there is a conflict between the
+registry listed below and the external registry, implementers SHOULD
+be aware of versioning differences of this specification and proceed
+with caution.
```
dlongley: Removes "registry" language and tweaks to normative language that can't be tested:

```suggestion
implementers are expected to use that reference instead of this table
for protocol usage. In the event that there is a conflict between the
table below and information in the new reference, implementers
ought to note versioning differences of this specification and proceed
with caution.
```

jandrieu:
Let's not call this a registry. Simply list it as a set of related resources.

PatStLouis:
Discussed: omit versioning for now, likely provide non normative guidance that protocols can include versioning in the protocol key with a specific format.

Limit initial "registry" (might not want to call it a registry) to current "mainstream" protocols

Mention warnings around the oid4* protocol keys and their associated versions/drafts.

index.html
PatStLouis: @kezike I can propose a didcomm format after this pr landed

index.html
```diff
@@ -3303,6 +3309,107 @@ <h4>Interaction Protocols Response</h4>
...
+the wallet should perform an Authorization Request for the exchange.
+More details can be found in [[[OPENID4VCI]]].
+                </p>
+                <p>
+Implementers should be aware of the drift that has occurred with this specification
+and the impact that this has made on the semantics of this property.
+Specifically, due to the material differences between popular draft versions and the
+stable 1.0 version, it is difficult for a wallet to know how to process an inbound request
+without coordination in advance. As a result of this, implementers should consider using
+the versioned variant of this property as described in the protocol versioning note below.
```
kezike: @msporny @dlongley Let me know what you think about this warning I used for the OID4* protocols (used the same one for `OID4VCI` here and `OID4VP` below).

index.html
```diff
@@ -3303,6 +3309,107 @@ <h4>Interaction Protocols Response</h4>
...
+            </tr>
+          </tbody>
+        </table>
+
+        <p class="note" title="Versioning exchange protocols">
+In order to remove ambiguity around the version of a protocol that
+an exchange supports, implementers might consider adding information
+to the protocol properties. One way to do this would be to use the
+pattern "`PROTOCOL`-`MAJOR`.`MINOR`". For example, instead of `OID4VP`,
+they could use `OID4VP-draft28` or `OID4VP-1.0`.
```
kezike: @dlongley Let me know what you think about this protocol versioning guidance.

index.html
```diff
@@ -3303,6 +3309,107 @@ <h4>Interaction Protocols Response</h4>
...
+              </td>
+            </tr>
+            <tr>
+              <td>`didcomm`</td>
+              <td>
+The `didcomm` interaction protocol is used to initiate a credential presentation
+flow in accordance with the latest version of [[DIDCOMM]].
+The value of this entry should be the Out-of-Band (OOB) Invitation URL for the exchange.
+More details can be found in [[[DIDCOMM]]].
+              </td>
```
kezike: @PatStLouis I decided to add the DIDComm protocol entry to the table so that we can just contain all supported protocols in one PR without worry of forgetting or running out of time. Let me know what you think of this description.

index.html
```diff
@@ -3303,6 +3309,107 @@ <h4>Interaction Protocols Response</h4>
...
+              <td>`vcapi`</td>
+              <td>
+The `vcapi` interaction protocol is used to initiate a specific [=exchange=]
+as described in Section [[[#participate-in-an-exchange]]].
+The value of this entry should be the fully-qualified URL used as the ID for the exchange.
+More details can be found in Section [[[#vcapi-interaction-protocol]]].
+              </td>
+            </tr>
+            <tr>
+              <td>`OID4VCI`</td>
```
kezike: @dlongley I asked about this a while back, but I want to try one more time: should we just make an executive decision to make all protocols lowercase (i.e., including `OID4VCI` and `OID4VP`)? I know that there are implementations in the wild that already expect uppercase, but 1. there is already sufficient enough drift in the OID4* specs to justify a clean start and 2. it would establish a clear standard for future protocol entries, which should reduce confusion for protocol architects and implementers. Thoughts?

index.html
msporny: At a high-level sounds fine to me.

Generally, you will want to avoid using the words "must", "should", "may" when you don't mean them in the capitalized RFC language sense. A small number of readers tend to get very confused with lowercase RFC language vs. uppercase, and what they need to implement. Use language such as "Implementers are advised to understand the drift that has... As a result of this, implementers are advised to use the..."

index.html
msporny: We should probably provide much stronger guidance here and say they SHOULD use the OID4VP-1.0 string.

index.html
msporny: I would prefer all lowercase protocol names. We can tell implementers that have shipped uppercase names that they should treat the uppercase values as lowercase (and that's fine and good as a backwards compatible thing to do).

index.html
kezike: Agreed, avoided it in a few places, but missed those ones. Just updated it in [9d75cae](https://github.com/w3c/vcalm/pull/680/commits/9d75cae488f0c0e210d45979a8aeab46862233e0).

index.html
kezike: Done in [9d75cae](https://github.com/w3c/vcalm/pull/680/commits/9d75cae488f0c0e210d45979a8aeab46862233e0)

index.html
```diff
@@ -3318,6 +3324,108 @@ <h4>Interaction Protocols Response</h4>
...
+        </p>
+
+        <table class="simple">
+          <thead>
+            <tr>
+              <th>Property</th>
+              <th>Usage</th>
+            </tr>
+          </thead>
+          <tbody>
```
kezike, line 3401: @dlongley I am just realizing that the OpenAPI spec for the [exchange protocol endpoint](https://w3c.github.io/vcalm/#get-exchange-protocols) includes `interact`, but does not include `inviteRequest` (and now `didcomm`) and the reverse is true for the [interaction protocols response](https://w3c.github.io/vcalm/#interaction-protocols-response) section. I suspect that the `interact` protocol is an artifact of a previous iteration, as this property is primarily now used in CHAPI land to reference/deliver an interaction URL. Am I correct in my understanding? I know I will have to change at least one of these two sections in order to achieve parity between them, but I want to be sure I am changing the right thing.

dlongley: @kezike,

> I suspect that the interact protocol is an artifact of a previous iteration...

I wouldn't call it an artifact; it's not "historical" or an error, if that's what you mean. It's how you express an interaction URL in a "protocols" object.

The `exchange protocol endpoint` OpenAPI spec should be updated to include `inviteRequest` and `didcomm`.

The `interaction protocols response` section looks like it's just an example. Examples might be targeted at a subset of protocols mentioned there, but a fully comprehensive example would, I guess, have them all). That might be a pretty unusual use case though.

index.html
dlongley: When adding the version numbers and so on, +1 let's lowercase things. We can keep the old uppercase values around for backwards compatibility, but let's encourage SHOULD with the new lowercase, versioned ones.

index.html
kezike: > It's how you express an interaction URL in a "protocols" object.

By this, do you mean that it references the interaction URL (`?iuv=1`)? If so, isn't this redundant/self-referential, since the GET response from that URL is the protocols object?

> Examples might be targeted at a subset of protocols mentioned there

Fair, but I would have to at least add it to the new table that this PR introduces. So, I will do that soon if it is not redundant (per my comment above).

dlongley: > By this, do you mean that it references the interaction URL (?iuv=1)? If so, isn't this redundant/self-referential, since the GET response from that URL is the protocols object?

No, the use case here is going to another place with a different interaction URL (and it might even have a different origin, establishing delegation); nothing self-referential.

> Fair, but I would have to at least add it to the new table that this PR introduces. So, I will do that soon if it is not redundant (per my comment above).

+1

index.html
```diff
@@ -3323,6 +3329,142 @@ <h4>Interaction Protocols Response</h4>
...
+            </tr>
+            <tr>
+              <td>`OID4VCI`</td>
+              <td>
+                <p>
+A deprecated version of the newer `oid4vci` interaction protocol property
+that is indifinitely retained here for backwards compatibility.
+Implementers are advised to use the `oid4vci` property for OID4VCI exchanges
+and allow for a reasonable transition period during which they interpret
+`OID4VCI` from peer systems as `oid4vci`.
```
```suggestion
A deprecated and unversioned variant of the `oid4vci` versioned interaction
protocol properties that is indefinitely retained here for backwards
compatibility. Implementers are advised to upgrade to one or more of the
versioned variants of the `oid4vci` properties, only using the deprecated
`OID4VCI` property for a reasonable transition period during which they interpret
`OID4VCI` from peer systems as an unversioned or multiversion variant.
```

index.html
```diff
@@ -3323,6 +3329,142 @@ <h4>Interaction Protocols Response</h4>
...
+            </tr>
+            <tr>
+              <td>`OID4VP`</td>
+              <td>
+                <p>
+A deprecated version of the newer `oid4vp` interaction protocol property
+that is indifinitely retained here for backwards compatibility.
+Implementers are advised to use the `oid4vp` property for OID4VP exchanges
+and allow for a reasonable transition period during which they interpret
+`OID4VP` from peer systems as `oid4vp`.
```
```suggestion
A deprecated and unversioned variant of the `oid4vp` versioned interaction
protocol properties that is indefinitely retained here for backwards
compatibility. Implementers are advised to upgrade to one or more of the
versioned variants of the `oid4vp` properties, only using the deprecated
`OID4VP` property for a reasonable transition period during which they interpret
`OID4VP` from peer systems as an unversioned or multiversion variant.
```

index.html
```diff
@@ -3323,6 +3329,142 @@ <h4>Interaction Protocols Response</h4>
...
+          </tbody>
+        </table>
+
+        <p class="note" title="Versioning exchange protocols">
+In order to remove ambiguity around the protocol versions that
+an exchange supports, implementers might consider adding information
+to the protocol properties. One way to do this would be to use the
+pattern "`PROTOCOL`-`MAJOR`.`MINOR`". For example, instead of `oid4vp`/`OID4VP`,
+they could use `oid4vp-draft28` or `oid4vp-1.0`. However, in the case of
+`oid4vci`/`OID4VCI` and `oid4vp`/`OID4VP`, implementers SHOULD use `oid4vci-1.0` and `oid4vp-1.0` respectively.
```
```suggestion
For protocols where version ambiguity could cause interoperability
issues, implementers might consider adding information to the protocol
properties. One way to do this would be to use the
pattern "`PROTOCOL`-`MAJOR`.`MINOR`". For example, instead of `oid4vp`,
they ought to use `oid4vp-draft28` or `oid4vp-1.0`.
```

oas.yaml
```diff
@@ -1118,16 +1118,28 @@ components:
...
             vcapi:
               type: string
               description: The URL to use when initiating a VCALM exchange.
-            OID4VP:
+            oid4vci:
+              type: string
+              description: The URL to use when initiating an OID4VCI issuance.
+            oid4vp:
               type: string
               description: The URL to use when initiating an OID4VP presentation.
```
```suggestion
```

dlongley (changes requested):
No one presently supports the unversioned, lowercase variants of `oid4vp` and `oid4vci` -- and we should not introduce them given that we know that using versions is important for interop for those protocols. We should abandon unversioned variants for those two protocols rather than introduce lowercase versions of the uppercase, unversioned variants (which are only preserved for backwards compatibility).

dlongley (approved):
Approving after my suggestions were applied, but there's an issue with this PR deleting and replacing everything, perhaps a whitespace, EOL problem like we've seen recently with other specs.

msporny:
Yes, please fix whatever broke in the PR, merging a PR that touches every line in the spec will break maintainability in a  variety of bad ways for the Editors (like not being able to tell anymore, who changed which line in the spec and when, which Editors need to do from time to time when trying to track down which PR changed a particular line in the spec, and the people that were involved in that conversation).

dlongley:
It's the second-to-last commit (at the moment anyway): https://github.com/w3c/vcalm/pull/680/changes/68c9f4afe603956893372e37a17c07e8c38db949 that changes everything, the one that applies suggestions in github.

Maybe that one just needs to be pulled out and reapplied manually. If you look at the diff for that one, click the settings cog next to the green "Submit Review" button, and hide whitespace changes, you can see what actually needs to be changed (it's small). (Remember to turn it back on when you're done).

It could also potentially be applied as "patches" where only the important changes are grabbed, if it reordered after the `Add .gitattributes file` commit.

dlongley (approved):

PR #690, opened 2026-08-04
Fix oneOf+required respec render issue

eric-schuh:
This PR attempts to address https://github.com/w3c/vcalm/issues/605 which details how OAS components that make use of both _oneOf_ and _required_ were not being rendered correctly. 

To fix, the oas.yaml was updated for the IssueRequests and OpenIDClientProfile components were updated to match the pattern of oneOf+required found in the ExchangeParticipationClientMessage component. I don't believe that these changes modify the underlying meaning of the OAS for these components but I would be nice to have that double checked. 

As part of this the IssueRequests:credentialTemplateIndex property was updated from the type: "number" to type: "string". This was primarily because I could not find any other instance of type: "number" being used. If we want to keep this as type: "number" a respec-oas update will be needed as currently it has no case handling for "number" types. 

To fix this in other ways, a changes to respec-oas will be needed. Currently this PR is not dependent on any respec-oas changes.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Aug 4, 2026, 6:47 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://www.w3.org/publications/spec-generator/?type=respec&output=html&url=https%3A%2F%2Fraw.githubusercontent.com%2Feric-schuh%2Fvcalm%2F3ef2f08636f7f2a82b509df08b2f342d3e69291d%2Findex.html%3FisPreview%3Dtrue)

**Error output:**

```
EISDIR: illegal operation on a directory, open 'uploads/tqJyHK/'
```

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20w3c/vcalm%23690.)._

</details>


eric-schuh:
Forgot some cleanup, will do that and reopen this PR

PR #692, opened 2026-08-04
Fix oneOf + required rendering issues

eric-schuh:
This PR attempts to address #605 which details how OAS components that make use of both oneOf and required were not being rendered correctly.

To fix, the oas.yaml was updated for the IssueRequests and OpenIDClientProfile components were updated to match the pattern of oneOf+required found in the ExchangeParticipationClientMessage component. I don't believe that these changes modify the underlying meaning of the OAS for these components but I would be nice to have that double checked.

As part of this the IssueRequests:credentialTemplateIndex property was updated from the type: "number" to type: "string". This was primarily because I could not find any other instance of type: "number" being used. If we want to keep this as type: "number" a respec-oas update will be needed as currently it has no case handling for "number" types.

To fix this in other ways, changes to respec-oas will be needed. Currently this PR is not dependent on any respec-oas changes.


PatStLouis:
discussed: we should favor the OAS accuracy, rendered OAS comes second. Lets review the semantics of the data models to ensure they are representative of the expected request/response bodies.

TallTed (commented):
lgtm

kezike:
@eric-schuh Does PR #706 obviate this PR (you mentioned that it overrides PR #691, but not sure if it does the same for this PR)?

PatStLouis:
discussed: Merge this as is and then review the spec holistically for pattern alignment based on the following observation


Investigate the possible duplication related to `oneOf` object nesting. @dlongley suggested expressing which properties are acceptable by referencing the property name under `oneOf` and hope no 2 configuration share the same properties as `required`

PR #700, opened 2026-08-11
Add QueryByExample matching and JSON pointer conversion algorithms

### LINKED ISSUE #545: Add selective disclosure example to spec w/ conversion of QueryByExample to JSON pointers

### LINKED ISSUE #574: Add normative algorithm for QueryByExample matching

PatStLouis:
## Summary
- Adds a normative **Matching a Credential to a QueryByExample** algorithm (and Matching Object Properties helper) so holders can determine whether a held credential satisfies a `credentialQuery` ([#574](https://github.com/w3c/vcalm/issues/574)).
- Adds a normative **Converting QueryByExample to JSON Pointers** algorithm (plus Adjust Reveal Pointers / Prune Shallow Pointers) for selective disclosure, with a MUST same-output conformance bar and one worked `selectivePointers` example ([#545](https://github.com/w3c/vcalm/issues/545)).
- Written in [vc-di-ecdsa §3 Algorithms](https://www.w3.org/TR/vc-di-ecdsa/#algorithms) style. `@context` uses ordered prefix matching; `acceptedIssuers` matches the current property table (string / `{id}` / `recognizedIn`). Supersedes the approach in [#624](https://github.com/w3c/vcalm/pull/624); builds on prior draft work in `bparth24-issue-574`.

Closes #574
Closes #545
Refs #624

## Test plan
- [ ] Preview Respec rendering of the new subsections under Query By Example (algorithm list numbering, cross-links).
- [ ] Confirm `@context` prefix match and `acceptedIssuers` cases against implementer expectations / oid4-client.
- [ ] Confirm the PermanentResidentCard → `selectivePointers` example (`/type`, `/issuer`, `/credentialSubject/birthCountry`) is clear for VC 2.0 + derive.
- [ ] Assignee review from #574 / #545 as appropriate (`kezike`, `ottonomy`, `PatStLouis`).

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Aug 11, 2026, 6:57 PM UTC)_.

<details>
<summary>More</summary>

**Error output:**

```
EISDIR: illegal operation on a directory, open 'uploads/I8jWyo/'
```

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20w3c/vcalm%23700.)._

</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1730,6 +1730,397 @@ formats are also acceptable.
           such as <code>bbs-2023</code> or <code>ecdsa-sd-2023</code> in the <code>acceptedCryptosuites</code> array.
         </p>
 
+        <section>
+          <h4 id="matching-a-credential-to-a-querybyexample">Matching a Credential to a QueryByExample</h4>
+
+          <p>
+The following algorithm specifies how a conforming [=holder=] implementation
+determines whether a [=verifiable credential=] matches the selection criteria
+in a QueryByExample <code>credentialQuery</code> object. Required inputs are a
+[=verifiable credential=] ([=map=] <var>credential</var>) and a
+<code>credentialQuery</code> object ([=map=] <var>credentialQuery</var>). A
+boolean match result is produced as output. A conforming implementation MUST
+produce the same output as this algorithm.
+          </p>
+
+          <p>
+Matching proceeds in two stages. First, if an <code>example</code> is present,
+each of its properties is treated as a requirement against
+<var>credential</var>. Second, if <code>acceptedIssuers</code> is present, the
+credential's issuer MUST be acceptable. The credential matches only if both
+stages succeed when present.
+          </p>
+
+          <ol class="algorithm">
+            <li>
+If <var>credential</var> is not a [=map=], return <code>false</code>.
+            </li>
+            <li>
+If <var>credentialQuery</var> contains an <code>example</code> property,
+initialize <var>example</var> to its value and perform the following steps:
+              <ol class="algorithm">
+                <li>
+If <var>example</var> contains an <code>@context</code> property, initialize
+<var>expectedContexts</var> to its value normalized to a [=list=], and
+initialize <var>credentialContexts</var> to the value of the
+<code>@context</code> property of <var>credential</var> normalized to a
+[=list=]. If the length of <var>credentialContexts</var> is less than the
+length of <var>expectedContexts</var>, return <code>false</code>. For each
+index <var>i</var> from <code>0</code> to the length of
+<var>expectedContexts</var> minus one, if
+<var>expectedContexts</var>[<var>i</var>] is not strictly equal to
+<var>credentialContexts</var>[<var>i</var>], return <code>false</code>.
+                </li>
+                <li>
+If <var>example</var> contains a <code>type</code> property, initialize
+<var>expectedTypes</var> to its value normalized to a [=list=], and initialize
+<var>credentialTypes</var> to the value of the <code>type</code> property of
+<var>credential</var> normalized to a [=list=]. If no value in
+<var>expectedTypes</var> is strictly equal to any value in
+<var>credentialTypes</var>, return <code>false</code>.
+                </li>
+                <li>
+Initialize <var>remaining</var> to a copy of <var>example</var> with the
+<code>@context</code> and <code>type</code> properties removed, if present. If
+<var>remaining</var> contains any properties, set <var>match</var> to the
+result of running the algorithm in Section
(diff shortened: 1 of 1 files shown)


PatStLouis:
Follow-up push addressing self-review notes:

1. **List-of-maps matching** — when \`expectedValue\` is a list, map elements now recurse through Matching Object Properties (plus strict equality / numeric coercion for non-maps), instead of strict equality only.
2. **Mandatory pointers** — conversion always appends \`/type\` and \`/issuer\` when missing, so the MUST same-output bar aligns with the PermanentResidentCard \`selectivePointers\` example. Extra cryptosuite mandatory pointers remain out of scope for this conversion.

index.html
```diff
@@ -1730,6 +1730,422 @@ <h3>Query By Example</h3>
...
+produce the same output as this algorithm.
+          </p>
+
+          <p>
+Matching proceeds in two stages. First, if an <code>example</code> is present,
+each of its properties is treated as a requirement against
+<var>credential</var>. Second, if <code>acceptedIssuers</code> is present, the
+credential's issuer MUST be acceptable. The credential matches only if both
+stages succeed when present.
+          </p>
```
TallTed: What if either is not present? It's not clear whether a missing `<code>example</code>` or `<code>acceptedIssuers</code>` is a success or a failure. Either would allow making this paragraph clearer, as below.
```suggestion
Matching proceeds in two stages. The credential matches only if both stages
succeed. First, if <code>example</code> is present, each of its properties is
treated as a requirement against <var>credential</var>; if <code>example</code>
is not present, this stage is a success. Second, if <code>acceptedIssuers</code>
is present, the credential's issuer MUST be acceptable; if
<code>acceptedIssuers</code> is not present, this stage is a success. 
          </p>
```

dlongley:
An existing implementation to compare against this algorithm:

Conversion from JSON to a flat or a "deep" map of JSON pointers:
(keys are pointers that point at maps/sets of other such "deep" maps):
https://github.com/digitalbazaar/oid4-client/blob/v5.13.2/lib/query/util.js

Matching a credential against a flat/"deep" map of JSON pointers:
https://github.com/digitalbazaar/oid4-client/blob/v5.13.2/lib/query/match.js

Perhaps a task that Claude could do to check for gaps/inconsistencies.

PatStLouis:
discussed: @PatStLouis will have a look at the proposed code from @dlongley to identify gaps in the pr algorithm

PR #626, opened 2026-04-14
docs: add informative reference workflow step processing algorithm

PatStLouis:
Add an informative subsection under Workflows and Exchanges with the exchange processing loop and hooks described in GitHub discussion, plus editorial fixes (grammar, parenthesis) from the issue text.

Closes #596.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 28, 2026, 7:33 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/publications/spec-generator/) - Spec Generator is the web service used to build bikeshed/ReSpec specs

:link: [Related URL](https://www.w3.org/publications/spec-generator/?type=respec&output=html&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fvcalm%2F880ae389e216de9b160d0e104d9e3cbf949c5a72%2Findex.html%3FisPreview%3Dtrue)

**Error output:**

```
EISDIR: illegal operation on a directory, open 'uploads/8ER9gU/'
```

_This seems to be an issue with the [Spec Generator](https://www.w3.org/publications/spec-generator/) service. PR Preview doesn't manage this service and so has no control over it. If you've identified an issue with it, you can [report the issue to the maintainers of Spec Generator](https://github.com/w3c/spec-generator/issues/new) directly. Please be courteous. Thank you!_

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20w3c/vcalm%23626.)._

</details>

PR #701, opened 2026-08-11
Add informative reference workflow step processing appendix

### LINKED ISSUE #596: Consider defining normative/non-normative workflow processing algorithm
We could specify an algorithm for processing a workflow (walking through its steps until either information is needed from the client or until the server has generated a response to give to the client). I think we'd want to leave some space for implementations to differ as needed here, but it would be helpful for implementers to know of at least one way to process workflow steps.

Here's a description of a workflow step processing algorithm that is in an existing implementation. This same algorithm is run on exchanges that communicate with the client using VC-API, OID4VP, OID4VCI, or inviteRequest -- though there are places for specific hooks to run where protocol customization might be needed. The resulting `response` output can be sent directly when using VC-API or transformed into some other protocol messages as needed for other protoocols.

```
0. Receive `workflow` and `exchange` as inputs and `receivedPresentation` and
  `receivedPresentationRequest` as optional inputs.
1. Initialize `step` and `response` to `null`.
2. If `exchange.state` is `complete` or `invalid`, throw a `NotAllowedError`.
3. If `exchange.state` is `pending`, set it to `active`.
4. Continuously loop to process exchange steps, optionally saving any error
  thrown as `exchange.lastError`. Other algorithm steps will return out of the
  loop when a full response is generated, input from the exchange client is required
  or the exchange times out. An implementation specific maximum step count be
  optionally enforced to handle misconfigured workflows.
  4.1. Set `step` to the current step (evaluating a step template as needed).
  4.2. Call sub-algorithm `prepareStep` hook, passing `workflow`, `exchange`, `step`,
    `receivedPresentation`, and `receivedPresentationRequest` to perform any
    protocol-specific custom step preparation. If `prepareStep` returns a
    `prepareResult` with `receivedPresentation` and/or
    `receivedPresentationRequest` set, then update `receivedPresentation` and/or
    `receivedPresentationRequest` accordingly.
  4.3. If `receivedPresentation` is set, then call the `validateReceivedPresentation`
    sub-algorithm, passing `workflow`, `exchange`, `step`, and `receivedPresentation`.
  4.4. If `receivedPresentationRequest` is set, call the
    `validateReceivedPresentationRequest` sub-algorithm, passing `exchange`,
    `step`, and `receivedPresentationRequest`.
  4.5. If the implementation supports blocking callbacks that can return
    results to be added to exchange variables (or return errors), call the
    callback and store its results in
    `exchange.variables.results[exchange.step].callbackResults` or throw any
    error received.    
  4.6. Set `isInputRequired` to the result of calling the `inputRequired` sub-algorithm
    hook passing `step` and `receivedPresentation` to perform any protocol-specific
    input checks.
  4.7. If `isInputRequired` is `true`:
    4.7.1. If `response` is `null`, set it to an empty object.
    4.7.2. If `step.verifiablePresentationRequest` is set, call the
      `createVerifiablePresentationRequest` sub-algorithm, passing `workflow`,
      `exchange`, `step`, and `response`.
    4.7.3. Save the exchange (and call any non-blocking callback in the step) and
      return `response`.
  4.8. Set `issueToClient` to `true` if `step.issueRequests` includes any issuer
    requests for VCs that are to be sent to the client (`issueRequest.result` is
    NOT set), otherwise set it to `false`.
  4.9. If `step.verifiablePresentation` is set or `issueToClient` is `true`:
    4.9.1. If `response` is not `null`:
      4.9.1.1. If `response.verifiablePresentationRequest` is not set, set it to an empty
        object (to indicate that the exchange is not yet complete).
      4.9.1.2. Save the exchange (and call any non-blocking callback in the step).
      4.9.1.3. Return `response`.
    4.9.2. Set `response` to an empty object.
    4.9.3. If `step.verifiablePresentation` is set, set `response.verifiablePresentation` to a
      copy of it, otherwise set `response.verifiablePresentation` to a new, empty,
      Verifiable Presentation (using VCDM 2.0 by default, but a custom configuration could
      specify another version).
  4.10. Perform every issue request (optionally in parallel), returning an error response to
    the client if any fails (note: implementations can optionally implement failure recovery
    or retry issue requests at their own discretion):
    4.10.1. For each issue request where `result` is set to an exchange variable path or name,
      save the issued credential in the referenced exchange variable.
    4.10.2. For each issue request where `result` is not specified, save the issued credential in
      `response.verifiablePresentation`, i.e., for a VCDM presentation, append the issued
      credential to `response.verifiablePresentation.verifiableCredential`.
  4.11. If `response.verifiablePresentation` is set and the step configuration indicates it
      should be signed, sign the presentation (e.g., by using a VCALM holder instance's
      `/presentations/create` endpoint).
  4.12. Call subalgorithm `isStepComplete`, passing `workflow`, `exchange`,
    `step`, `receivedPresentation`, and `receivedPresentationRequest` to
    perform any protocol-specific behavior to determine if the step is
    complete. Set `stepComplete` to the result of `isStepComplete`, defaulting
    to `true`.
  4.13. If `stepComplete` is `true`:
    4.13.1. If `step.redirectUrl` is set:
      4.13.1.1. If `response` is `null` then set it to an empty object.
      4.13.1.2. Set `response.redirectUrl` to `step.redirectUrl`.
    4.13.2. If `step.nextStep` is not set then set `exchange.state` to `complete`.
    4.13.3. Otherwise, delete `exchange.variables.results[step.nextStep]` if it exists, and set
      `exchange.step` to `step.nextStep`.
  4.14. Save the exchange ((and call any non-blocking callback in the step).
  4.15. If `exchange.state` is `complete`, return `response` if it is not `null`, otherwise
    return an empty object.
  4.16. Set `receivedPresentation` to `null` and loop.
```

EDIT: Fixed typo, thanks @TallTed!
EDIT2: Added optional blocking callback step for implementations and indicated
where non-blocking callbacks are called.

PatStLouis:
## Summary
- Adds an **informative** appendix documenting one reference algorithm for walking exchange/workflow steps until client input is needed or a response is ready ([#596](https://github.com/w3c/vcalm/issues/596)).
- Uses `ol.algorithm` style (same approach as Data Integrity / ECDSA cryptosuites and [#700](https://github.com/w3c/vcalm/pull/700)); includes protocol hooks (`prepareStep`, `inputRequired`, `isStepComplete`, etc.) so the loop stays agnostic to VC-API / OID4VP / OID4VCI / invite bindings.
- Cross-links from Workflows and Exchanges. Supersedes the `<pre>`-based approach in [#626](https://github.com/w3c/vcalm/pull/626).

**Stack:** Branched from [#700](https://github.com/w3c/vcalm/pull/700) (`OpSecId:docs/qbe-matching-and-json-pointers`). This PR currently targets `main`, so the GitHub diff may also show #700 commits until that PR merges—review the tip commit `docs: add informative reference workflow step processing appendix` for this change. After #700 merges, rebase this branch onto `main`.

Closes #596
Supersedes #626

## Test plan
- [ ] Respec preview: appendix marked informative; nested `ol.algorithm` numbering renders.
- [ ] Cross-link from Workflows and Exchanges resolves to the appendix.
- [ ] Spot-check algorithm steps against Dave’s text in #596 / prior #626 content.
- [ ] After #700 merges: rebase onto `main` and confirm the diff is appendix-only.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Aug 11, 2026, 6:57 PM UTC)_.

<details>
<summary>More</summary>

**Error output:**

```
EISDIR: illegal operation on a directory, open 'uploads/uVOIek/'
```

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20w3c/vcalm%23701.)._

</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1730,6 +1730,397 @@ formats are also acceptable.
           such as <code>bbs-2023</code> or <code>ecdsa-sd-2023</code> in the <code>acceptedCryptosuites</code> array.
         </p>
 
+        <section>
+          <h4 id="matching-a-credential-to-a-querybyexample">Matching a Credential to a QueryByExample</h4>
+
+          <p>
+The following algorithm specifies how a conforming [=holder=] implementation
+determines whether a [=verifiable credential=] matches the selection criteria
+in a QueryByExample <code>credentialQuery</code> object. Required inputs are a
+[=verifiable credential=] ([=map=] <var>credential</var>) and a
+<code>credentialQuery</code> object ([=map=] <var>credentialQuery</var>). A
+boolean match result is produced as output. A conforming implementation MUST
+produce the same output as this algorithm.
+          </p>
+
+          <p>
+Matching proceeds in two stages. First, if an <code>example</code> is present,
+each of its properties is treated as a requirement against
+<var>credential</var>. Second, if <code>acceptedIssuers</code> is present, the
+credential's issuer MUST be acceptable. The credential matches only if both
+stages succeed when present.
+          </p>
+
+          <ol class="algorithm">
+            <li>
+If <var>credential</var> is not a [=map=], return <code>false</code>.
+            </li>
+            <li>
+If <var>credentialQuery</var> contains an <code>example</code> property,
+initialize <var>example</var> to its value and perform the following steps:
+              <ol class="algorithm">
+                <li>
+If <var>example</var> contains an <code>@context</code> property, initialize
+<var>expectedContexts</var> to its value normalized to a [=list=], and
+initialize <var>credentialContexts</var> to the value of the
+<code>@context</code> property of <var>credential</var> normalized to a
+[=list=]. If the length of <var>credentialContexts</var> is less than the
+length of <var>expectedContexts</var>, return <code>false</code>. For each
+index <var>i</var> from <code>0</code> to the length of
+<var>expectedContexts</var> minus one, if
+<var>expectedContexts</var>[<var>i</var>] is not strictly equal to
+<var>credentialContexts</var>[<var>i</var>], return <code>false</code>.
+                </li>
+                <li>
+If <var>example</var> contains a <code>type</code> property, initialize
+<var>expectedTypes</var> to its value normalized to a [=list=], and initialize
+<var>credentialTypes</var> to the value of the <code>type</code> property of
+<var>credential</var> normalized to a [=list=]. If no value in
+<var>expectedTypes</var> is strictly equal to any value in
+<var>credentialTypes</var>, return <code>false</code>.
+                </li>
+                <li>
+Initialize <var>remaining</var> to a copy of <var>example</var> with the
+<code>@context</code> and <code>type</code> properties removed, if present. If
+<var>remaining</var> contains any properties, set <var>match</var> to the
+result of running the algorithm in Section
+[[[#matching-object-properties]]], passing <var>remaining</var> as
+<var>expectedObject</var> and <var>credential</var> as
+<var>actualObject</var>. If <var>match</var> is <code>false</code>, return
+<code>false</code>.
+                </li>
+              </ol>
+            </li>
+            <li>
+If <var>credentialQuery</var> contains an <code>acceptedIssuers</code>
+property, initialize <var>acceptedIssuers</var> to its value and perform the
+following steps:
+              <ol class="algorithm">
+                <li>
+Initialize <var>credentialIssuer</var> to the value of the
+<code>issuer</code> property of <var>credential</var>. If
+<var>credentialIssuer</var> is a [=map=], initialize
+<var>credentialIssuer</var> to the value of its <code>id</code> property.
+                </li>
+                <li>
+Initialize <var>issuerItems</var> to an empty [=list=]. For each
+<var>item</var> in <var>acceptedIssuers</var>, if <var>item</var> is a
+[=list=], append each of its elements to <var>issuerItems</var>; otherwise,
+append <var>item</var> to <var>issuerItems</var>.
+                </li>
+                <li>
+If no <var>item</var> in <var>issuerItems</var> is acceptable for
+<var>credentialIssuer</var>, return <code>false</code>. An <var>item</var> is
+acceptable if any of the following is true:
+                  <ol class="algorithm">
+                    <li>
+<var>item</var> is a [=string=] that is strictly equal to
+<var>credentialIssuer</var>.
+                    </li>
+                    <li>
+<var>item</var> is a [=map=] that contains an <code>id</code> property whose
+value is strictly equal to <var>credentialIssuer</var>.
+                    </li>
+                    <li>
+<var>item</var> is a [=map=] that contains a <code>recognizedIn</code>
+property, and <var>credentialIssuer</var> is recognized according to the
+[[[VC-RECOG-ENT]]] specification using that property's value.
+                    </li>
+                  </ol>
+                </li>
+              </ol>
+            </li>
+            <li>
+Return <code>true</code>.
+            </li>
+          </ol>
+
+          <p class="note" title="Properties not used during matching">
+The <code>reason</code>, <code>acceptedCryptosuites</code>, and
+<code>acceptedEnvelopes</code> properties of a <code>credentialQuery</code>
+object are not used during credential matching. The <code>reason</code>
+property is intended for display to the [=holder=] by their software. The
+<code>acceptedCryptosuites</code> and <code>acceptedEnvelopes</code> properties
+inform the [=holder=]'s software about which proof mechanisms and envelope
+formats the [=verifier=] will accept, and are used during presentation
+construction rather than credential selection.
+          </p>
+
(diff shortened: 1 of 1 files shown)


PatStLouis:
### Carry-over from #626 / #596 review feedback

This PR is meant to supersede #626 while keeping the review asks addressed:

**TallTed (#626 inline)** — Deeper overhang / hanging indents on the `<pre>` numbered list  
→ Resolved by replacing the `<pre>` with nested `<ol class="algorithm">` lists and the existing VCALM algorithm CSS counters (`1)`, `1.1)`, …), which is also what Manu asked for in the [26 May 2026 discussion](https://w3c.github.io/vc-wg/minutes/2026-05-26-alm.html#9117).

**TallTed (#596)** — Grammar (“information is”) and clearer nesting  
→ Intro rewritten (that phrase is gone); nesting handled by `ol.algorithm` rather than lettered outline tiers (`a.` / `i.`), so we stay aligned with Data Integrity / ECDSA algorithm style used elsewhere in this stack (#700).

**Manu (#596 + meeting)** — Non-normative, appendix, protocol-agnostic, algorithm `ol` markup  
→ Informative appendix (`class="appendix informative"`), protocol hooks kept underspecified, ECDSA-style “Required inputs… produced as output” + `ol.algorithm`.

**#626 editorial** — Balanced parentheses; implementation-specific max step count MAY  
→ Retained in the algorithm prose.

**TallTed (conversation)** — Retry PR Preview  
→ Process note only; this new PR can be previewed on its own when preview is available.

No further code changes planned for these items unless reviewers want lettered outline numbering instead of the nested algorithm counters.

PatStLouis:
Rebased onto updated #700 (includes list-of-maps matching + deterministic `/type`/`/issuer` mandatory pointers).

index.html
```diff
@@ -1730,6 +1730,422 @@ <h3>Query By Example</h3>
...
+If <var>example</var> contains an <code>@context</code> property, initialize
+<var>expectedContexts</var> to its value normalized to a [=list=], and
+initialize <var>credentialContexts</var> to the value of the
+<code>@context</code> property of <var>credential</var> normalized to a
+[=list=]. If the length of <var>credentialContexts</var> is less than the
+length of <var>expectedContexts</var>, return <code>false</code>. For each
+index <var>i</var> from <code>0</code> to the length of
+<var>expectedContexts</var> minus one, if
+<var>expectedContexts</var>[<var>i</var>] is not strictly equal to
+<var>credentialContexts</var>[<var>i</var>], return <code>false</code>.
```
ottonomy, line 1772: It might be nice to provide a human summary of the effect of context comparison, like "Ensure the contexts in the credential match each of the required list of contexts, allowing subsequent contexts in the credential after all required contexts.

index.html
```diff
@@ -1730,6 +1730,422 @@ <h3>Query By Example</h3>
...
+                </li>
+                <li>
+If <var>example</var> contains a <code>type</code> property, initialize
+<var>expectedTypes</var> to its value normalized to a [=list=], and initialize
+<var>credentialTypes</var> to the value of the <code>type</code> property of
+<var>credential</var> normalized to a [=list=]. If no value in
+<var>expectedTypes</var> is strictly equal to any value in
+<var>credentialTypes</var>, return <code>false</code>.
```
ottonomy, line 1780: The summary of this comparison is what, that we match ANY of the types that appear in the example. I would have assumed the intent would be more like the other properties in that all of the types that appear in `expectedTypes` must be present in `credentialTypes`. I read the current wording as "Any one type that appears in `expectedTypes` must appear in `credentialTypes`. If a query includes `VerifiableCredential` as a type, that would broadly match. 

Either we instruct people to not include `VerifiableCredential` here or we change the wording here to require all `expectedTypes` entries to appear in `credentialTypes` not just any.

Or I'm reading this incorrectly, feel free to correct my understanding.

index.html
```diff
@@ -1730,6 +1730,422 @@ <h3>Query By Example</h3>
...
+boolean match result is produced as output. A conforming implementation MUST
+produce the same output as this algorithm.
+          </p>
+
+          <p>
+Matching proceeds in two stages. First, if an <code>example</code> is present,
+each of its properties is treated as a requirement against
+<var>credential</var>. Second, if <code>acceptedIssuers</code> is present, the
+credential's issuer MUST be acceptable. The credential matches only if both
+stages succeed when present.
```
TallTed, line 1751: I believe I commented on this block, in one of the previous PRs. There will be a number of other comments in that same PR, which would be best replicated here. Please let me know if you will handle that migration, or if I need to put the PRs side by side and do it.

PatStLouis:
discussed: @PatStLouis needs to address comments

index.html
TallTed: I found one of my missing comments, deep in a mailbox -- https://github.com/w3c/vcalm/pull/700#pullrequestreview-4953401805

I thought I'd made more comments on that/this block, but if so, they're better at hiding.

Issue #682, opened 2026-07-21
Update threat_model folder name to threat-model


eric-schuh:
Suggested by @kezike on the July 21 call.

kezike:
Closing, as addressed by PR #707

PR #707, opened 2026-08-19
Fix threat model folder and landing page

kezike:
This PR applies the following updates:

- Rename threat model folder (addresses Issue #682).
- Remove "DRAFT" from threat model document.
- Use relative path to reference threat model document.

I decided to create a PR instead of pushing directly to `main` due to uncertainties around the following:

1. I decided to remove the parenthetical `DRAFT` from the threat model landing page in order to be consistent with the latest commit on `main`, which removed the explanatory paragraph that used “draft”. However, we are labeling this document as an Editor’s Draft and we are also referring to this as a draft in the [document status section](https://w3c.github.io/vcalm/#sotd). I believe this is fine, but I wanted to confirm that as well as what we want to do with the version (currently `v1`, but wondering if it should have a draft `v0.1` version or the same version as VCALM - `v1.0`).
2. I switched to using a relative path as the reference to the threat model (instead of the fully qualified ED URL, which may not be appropriate for the published `/TR/` spec). The issue with this (which I also noticed in other specs, like VC Barcodes) is that the published `/TR/` threat model document is returning a 404 error. Will this continue until we officially publish this as a WG note or something? In the meantime, should I switch back to the fully-qualified URL, or should we just keep it in, with the expectation of properly publishing it for CR, so as not to forget to switch it back?
3. For the _More details about this document_ accordion at the top of the threat model landing page, is it appropriate to use links to the VCALM spec, or should we be linking to the threat model document instead?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vcalm/pull/707.html" title="Last updated on Aug 21, 2026, 12:27 AM UTC (a0a36db)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vcalm/707/c34e8dc...a0a36db.html" title="Last updated on Aug 21, 2026, 12:27 AM UTC (a0a36db)">Diff</a>


msporny:
> I decided to remove the parenthetical DRAFT from the threat model landing page in order to be consistent with the latest commit on main, which removed the explanatory paragraph that used “draft”. 

This is fine for now. What's most important is all the links work.

> I switched to using a relative path as the reference to the threat model (instead of the fully qualified ED URL, which may not be appropriate for the published /TR/ spec).

This is the right thing to do

> The issue with this (which I also noticed in other specs, like VC Barcodes) is that the published /TR/ threat model document is returning a 404 error. Will this continue until we officially publish this as a WG note or something? 

It will continue until you add relative links from the /index.html to /threat-model/index.html. You will also have to update Echidna, I believe:

https://github.com/w3c/vc-recognized-entities/blob/main/echidna-manifest.txt

I have no idea if the above is the right thing to do, @iherman and I are trying to work on this.

> should I switch back to the fully-qualified URL, or should we just keep it in, with the expectation of properly publishing it for CR, so as not to forget to switch it back?

No, don't change back to the fully qualified URL.

> For the More details about this document accordion at the top of the threat model landing page, is it appropriate to use links to the VCALM spec, or should we be linking to the threat model document instead?

TBD, leave it the way it is for now.

msporny (approved):

PR #713, opened 2026-09-01
Add ExchangeVerifiablePresentation option to support exchange specific data

eric-schuh:
This PR attempts to address https://github.com/w3c/vcalm/issues/396

It adds a new option when responding in an exchange. The ExchangeVerifiablePresentation includes the exchangeData field that allows for exchange specific data to be included in a response to a VPR.

From the discussion in the issue the ExchangeVerifiablePresentation only appears and on option when participating in an exchange. As mentioned in the issue we may want to include more concrete examples, but in leu of having those ready this PR includes the exchangeData field in the example as a undefined/null JSON object.

### DIFF
--- a/components/ExchangeVerifiablePresentation.yml
+++ b/components/ExchangeVerifiablePresentation.yml
@@ -1,0 +1,85 @@
+openapi: 3.0.0
+info:
+  version: "1.0.0"
+  title: Simple W3C Verifiable Presentation Schema
+  description: This specifies a simplified W3C verifiable presentation schema sufficient for the purposes of VCALM.
+  license:
+    name: W3C Software and Document License
+    url: http://www.w3.org/Consortium/Legal/copyright-software.
+  contact:
+    name: GitHub Source Code
+    url: https://github.com/w3c/vcalm
+paths:
+components:
+  schemas:
+    ExchangeVerifiablePresentation:
+      description: >
+        A JSON-LD presentation document that also contains some exchange
+        specific data request. Used in an exchange as a response to a 
+        VPR that contains a request for data specific to the exchange, but
+        that is not contained within the response verifiable credentials 
+        themselves.
+      allOf:
+        - $ref: "./Presentation.yml#/components/schemas/PresentationDocument"
+        - type: object
+          properties:
+            exchangeData:
+              type: object
+              description: >
+                A JSON object that enables exchange specific data to be sent when a
+                VPR makes a request for some data not normally contained in a verifiable
+                presentation. To give some examples: if a VPR requests a VC using BBS, 
+                this field would contain the commitment information to create a pseudonym,
+                if a VPR asks for zcaps, this field would contain the reference IDs 
+                for each zcap to be generated, if a VPR request needs additional payment
+                information, this field would contain the requested data.
+      example:
+        {
+          "@context":
+            [
+                "https://www.w3.org/ns/credentials/v2",
+                "https://www.w3.org/ns/credentials/examples/v2",
+            ],
+          "holder": "did:example:123",
+          "type": "VerifiablePresentation",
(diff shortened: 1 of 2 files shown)


components/ExchangeVerifiablePresentation.yml
```diff
@@ -0,0 +1,85 @@
...
+      example:
+        {
+          "@context":
+            [
+                "https://www.w3.org/ns/credentials/v2",
+                "https://www.w3.org/ns/credentials/examples/v2",
+            ],
+          "holder": "did:example:123",
+          "type": "VerifiablePresentation",
+          "exchangeData": {...},
```
dlongley, line 45: We'll need a new context to use this new term with VP 2.0 -- but a VP 2.1 context could automatically include the term.

ottonomy: Is this data supposed to be signed as part of the VP, or would it be possible to deliver it as a top level JSON property in parallel to the `verifiablePresentation` property?

dlongley: It is intended to be (optionally) signed, yes.

kezike:
Outstanding tasks:

1. Create a new context for `exchangeData` (example: https://github.com/w3c/vc-bitstring-status-list/tree/main/contexts)
2. Add an issue in `vc-data-model` to add `exchangeData` property to `VerifiablePresentation` type in VC v2.1 context

msporny:
I raised a new issue for vc-data-model here: https://github.com/w3c/vc-data-model/issues/1649

eric-schuh:
@dlongley I have added a context file, I expect the URL I used is likely not the correct one for this but I was unsure what URL should be used to point to the definition of the exchangeData property at this time so tried to follow patterns I saw from other specs in hopes it would be close. 

Let me know what is wrong and I will update again!

contexts/2026/v1.jsonld
```diff
@@ -0,0 +1,9 @@
+{
```
dlongley, line 1: Minimally, this should be `v1rc1.jsonld` ... I'm not sure if it needs to have `vcalm` in the name. I guess just `v1rc1.jsonld` is ok for now, we'll have to figure out what the full context URL looks like.

contexts/2026/v1.jsonld
msporny, line 1: We want to release this as a `v1rc1.jsonld` and no need to date-stamp the directory... so, full path needs to be: `/contexts/v1rc1.jsonld`.

contexts/2026/v1.jsonld
```diff
@@ -0,0 +1,9 @@
...
+  "@context": {
+    "@protected": true,
+    "exchangeData": {
+      "@id": "https://w3id.org/vcalm#exchangeData",
```
dlongley, line 5:
```suggestion
      "@id": "https://www.w3.org/2018/credentials#exchangeData",
```

eric-schuh:
Update the URL base to: https://www.w3.org/2018/credentials#

Also need to put in a PR to the credentials repo adding this term.

components/ExchangeVerifiablePresentation.yml
```diff
@@ -0,0 +1,86 @@
...
+                if a VPR asks for zcaps, this field would contain the reference IDs 
+                for each zcap to be generated, if a VPR request needs additional payment
+                information, this field would contain the requested data.
+      example:
+        {
+          "@context":
+            [
+                "https://www.w3.org/ns/credentials/v2",
+                "https://www.w3.org/ns/credentials/examples/v2",
+                "https://w3id.org/vcalm#exchangeData"
```
dlongley, line 42: Something like this for now:

```suggestion
                "https://w3id.org/vcalm/v1rc1"
```

Issue #719, opened 2026-09-09
Add QR Code shoulder surfing to threat model


msporny:
When using Interaction URLs embedded in QR Codes, or any credential mechanism that uses a QR Code or optical scan to start a credential issuance process, there is a risk that an attacker will capture the QR Code and finish the issuance before the original intended recipient. This threat should be documented and responses provided to ensure that implementers understand what can prevent the attack.

PR #711, opened 2026-08-25
Resolve tension between Threat Model and Privacy & Security Considerations

eric-schuh:
This PR is in response to: https://github.com/w3c/vcalm/issues/687

Adding threats created from transforming the privacy and security consideration sections into threats. Also modifying the privacy and security consideration sections to inform readers of the transition to using the threat model approach in place of these sections.

Any sections from the security considerations that did not have threats that directly applied to this threat model but still wanted to make a recommendation were move to appendixes in the threat model.

***
<a href="https://pr-preview.s3.amazonaws.com/eric-schuh/vcalm/pull/711.html" title="Last updated on Sep 14, 2026, 4:33 PM UTC (48fd3d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vcalm/711/f91b9a2...eric-schuh:48fd3d1.html" title="Last updated on Sep 14, 2026, 4:33 PM UTC (48fd3d1)">Diff</a>

PR #720, opened 2026-09-09
Document QR Code shoulder surfing threat and mitigations

msporny:
This PR is an attempt to address issue #719 by documenting the QR Code shoulder surfing threat and adding mitigations and prose around what does and does not work.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vcalm/pull/720.html" title="Last updated on Sep 9, 2026, 12:28 AM UTC (3dd0195)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vcalm/720/f91b9a2...3dd0195.html" title="Last updated on Sep 9, 2026, 12:28 AM UTC (3dd0195)">Diff</a>


index.html
```diff
@@ -3254,6 +3254,39 @@ <h4>Interaction QR Code Format</h4>
...
+holder-initiated `inviteRequest` example above, so an observed code does not by
+itself carry an interaction forward.
+        </p>
+
+        <p class="note" title="Why wallet authorization bounce-backs are not recommended">
+Sending the [=holder=] back to the originating website to authorize their
+wallet is not recommended as a remedy. It costs every interaction usability to
+address a case that arises only where the screen is observable, and it does not
+close the gap, since software on the device can still intercept the handoff
+(<a href="threat-model/index.html#t13-interaction-url-read-from-qr-code-is-opened-by-unexpected-or-malicious-application">T13:
```
dlongley, line 3279:
```suggestion
wallet is not recommended as a remedy. It costs every interaction usability to
address a case that arises only where the screen is observable, creates additional
complexity and state management for websites and wallets, and it does not close
the gap, since software on the device can still intercept the handoff
(<a href="threat-model/index.html#t13-interaction-url-read-from-qr-code-is-opened-by-unexpected-or-malicious-application">T13:
```

dlongley (approved):

simoneonofri (commented):
I agree that an “approved wallet” bounce-back would add friction and could introduce gatekeeping. T30 exposes a narrower problem, though: possession of the interaction URL is sufficient for the first wallet to complete the pickup.

Related work suggests several useful controls:
 - [Alipay+](https://docs.alipayplus.com/alipayplus/alipayplus/code_scanning_payment_standards_mpp/security) recommends dynamic, time-limited codes, one-time session parameters, and results on both terminals.
 - [USENIX QRLogin study](https://www.usenix.org/conference/usenixsecurity25/presentation/zhang-xin) identifies unbound QR and browser sessions as a cause of authorization hijacking, while [POSAUTH](https://www.usenix.org/conference/usenixsecurity17/technical-sessions/presentation/bai) addresses a related first-use race by binding a payment token to the intended terminal. In the VC context
 - [Hoops and Matthes](https://arxiv.org/html/2401.09488v1) similarly use a per-login identifier both to map the browser and wallet and as the VP challenge.

Could T30 include an additional, wallet-neutral response where scanning registers only a candidate wallet, while delivery requires explicit confirmation binding the originating browser session, short-lived interaction, and wallet key? This would not solve T13, compromised endpoints, or inattentive confirmation, but possession of the URL alone would no longer be sufficient.

msporny:
> T30 exposes a narrower problem, though: possession of the interaction URL is sufficient for the first wallet to complete the pickup.

Not always -- remember that an Interaction URL can be used for more than just pickup and is displayable by more than just the issuer... it can be used for a simple re-direct (holder-supplied QR Code). It doesn't remove the point you're making, but just highlighting that Interaction URLs and QR Codes work differently from QR Codes like OID4 and DC API.

> * [Alipay+](https://docs.alipayplus.com/alipayplus/alipayplus/code_scanning_payment_standards_mpp/security) recommends dynamic, time-limited codes, one-time session parameters, and results on both terminals.

Interaction URL QR Codes are dynamic and time limited with one-time session parameters and results on both terminals. IOW, Interaction URLs + VCALM seem to be almost identical to Alipay which process many hundreds of billions of dollars in transactions a year (but I'm skeptical as I expect they're doing something else in the protocol to do further binding).

> * [USENIX QRLogin study](https://www.usenix.org/conference/usenixsecurity25/presentation/zhang-xin) identifies unbound QR and browser sessions as a cause of authorization hijacking, while [POSAUTH](https://www.usenix.org/conference/usenixsecurity17/technical-sessions/presentation/bai) addresses a related first-use race by binding a payment token to the intended terminal. In the VC context

This seems like an incomplete sentence?

> * [Hoops and Matthes](https://arxiv.org/html/2401.09488v1) similarly use a per-login identifier both to map the browser and wallet and as the VP challenge.

Our expectation is that if you can shoulder surf the QR Code, you can shoulder surf the per-login identifier. IOW, if the credential is high-stakes enough, just don't provide the issuer-provided QR Code engagement procedure... force it to a same-device flow.

> Could T30 include an additional, wallet-neutral response where scanning registers only a candidate wallet, while delivery requires explicit confirmation binding the originating browser session, short-lived interaction, and wallet key? 

Can you be more concrete about how this would happen? I think you might be suggesting something like what is done when doing code-based Bluetooth pairing?

eric-schuh:
I think this is a good new threat, I just want to be careful about numbering. The outstanding PR https://github.com/w3c/vcalm/pull/711 adds up to a T40 and R44. I believe that PR is ready to be merged as I just resolved the last comments there, so the new threat added here should probably be T41. 

I don't believe GitHub lets me suggest changes to a filename on a PR but I can push updates to the branch this was created from to make sure none of the numbering here conflicts with #711 if that is ok process wise.

msporny:
> I think this is a good new threat, I just want to be careful about numbering. 

Hmm, if we have to worry about numbering here, the tooling is probably not doing its job. :)

That's a side-quest/issue. I can update the numbering to align w/ whatever just went in, just noting that the numbering problem will be much larger when we go to a registry-based approach (maybe, I don't know what the plans there are).

In the meantime, yes, please feel free to push changes to this branch to do the renumbering, thanks @eric-schuh!

index.html
TallTed: Minor fixes to @dlongley's suggestion
```suggestion
wallet is not recommended as a remedy. It costs every interaction usability to
address a case that arises only where the screen is observable; it creates additional
complexity and state management for websites and wallets; and it does not close
the gap, since software on the device can still intercept the handoff
(<a href="threat-model/index.html#t13-interaction-url-read-from-qr-code-is-opened-by-unexpected-or-malicious-application">T13:
```

index.html
```diff
@@ -3254,6 +3254,39 @@ <h4>Interaction QR Code Format</h4>
           </figcaption>
         </figure>
 
+        <p class="note" title="Cross-device pickup and observed QR codes">
+A displayed [=interaction QR Code=] is readable by anyone with line of sight,
+and initiating an [=exchange=] requires no authorization beyond the exchange
+URL, so an observer can complete the exchange with their own wallet first. See
```
TallTed, line 3260:
```suggestion
URL, so an observer can complete the exchange with their own wallet before
the user who has displayed the [=interaction QR Code=]. See
```

index.html
```diff
@@ -3254,6 +3254,39 @@ <h4>Interaction QR Code Format</h4>
...
+<a href="threat-model/index.html#t30-shoulder-surfed-interaction-url-enables-cross-device-pickup">T30:
+Shoulder Surfed Interaction URL Enables Cross-Device Pickup</a>, which
```
TallTed, line 3262:
```suggestion
Shoulder-Surfed Interaction URL Enables Cross-Device Pickup</a>, which
```

index.html
```diff
@@ -3254,6 +3254,39 @@ <h4>Interaction QR Code Format</h4>
...
+recommends preferring pickup on the device already running the [=holder
```
TallTed, line 3263:
```suggestion
recommends preferring pickup by the device already running the [=holder
```

index.html
```diff
@@ -3254,6 +3254,39 @@ <h4>Interaction QR Code Format</h4>
...
+coordinator=], confirming receipt afterwards, and checking at presentation time
+that the presenter is the [=subject=]. Where cross-device engagement is used, it
```
TallTed, line 3265:
```suggestion
coordinator=], confirming receipt afterwards, and confirming at presentation time
whether the presenter is the [=subject=] or otherwise intended presenter. Where
cross-device engagement is used, it
```

eric-schuh:
I have a push ready to update the threat numbering here, however while doing so I found a rendering bug in respec-threats that I don't have time to track down before the call: https://github.com/LegReq/respec-threats/issues/10 

In short, if we want the preview to work so people can reason over the new content here the new threat needs to stay at index "30" for now as respec-threats won't render non-sequentially numbered threats. Once https://github.com/w3c/vcalm/pull/711 is pushed and the threats up to T40 have been added to the main spec the numbering changes can happen here and the rendering should work.

index.html
```diff
@@ -3254,6 +3254,39 @@ <h4>Interaction QR Code Format</h4>
...
+Sending the [=holder=] back to the originating website to authorize their
+wallet is not recommended as a remedy. It costs every interaction usability to
+address a case that arises only where the screen is observable, and it does not
+close the gap, since software on the device can still intercept the handoff
+(<a href="threat-model/index.html#t13-interaction-url-read-from-qr-code-is-opened-by-unexpected-or-malicious-application">T13:
+Interaction URL Read from QR Code is Opened by Unexpected or Malicious
+Application</a>). Doing so requires distinguishing approved wallets from
+unapproved ones, with the effects described in
+<a href="threat-model/index.html#t12-use-of-protocol-to-centralize-digital-credentials-ecosystem-and-lockout-competition">T12:
+Use of Protocol to Centralize Digital Credentials Ecosystem and Lockout
```
TallTed, line 3284:
```suggestion
<a href="threat-model/index.html#t12-use-of-protocol-to-centralize-digital-credential-ecosystem-and-lock-out-competition">T12:
Use of Protocol to Centralize Digital Credential Ecosystem and Lock Out
```

TallTed (changes requested):
Small tweaks, including renaming of T12, which renaming must be carried through to other locations/documents.

eric-schuh:
> I have a push ready to update the threat numbering here, however while doing so I found a rendering bug in respec-threats that I don't have time to track down before the call: [LegReq/respec-threats#10](https://github.com/LegReq/respec-threats/issues/10)
> 
> In short, if we want the preview to work so people can reason over the new content here the new threat needs to stay at index "30" for now as respec-threats won't render non-sequentially numbered threats. Once #711 is pushed and the threats up to T40 have been added to the main spec the numbering changes can happen here and the rendering should work.

Ignore this last comment, I forgot to update the threat list in the index.html which was causing this issue. Files should be updated with indexes that will not conflict with #711

index.html
ottonomy: It is an application on the device that will execute this, not the hardware of the device itself, so `on` seems correct to me here.

index.html
```diff
@@ -3254,6 +3254,39 @@ <h4>Interaction QR Code Format</h4>
           </figcaption>
         </figure>
 
+        <p class="note" title="Cross-device pickup and observed QR codes">
+A displayed [=interaction QR Code=] is readable by anyone with line of sight,
+and initiating an [=exchange=] requires no authorization beyond the exchange
+URL, so an observer can complete the exchange with their own wallet first. See
+<a href="threat-model/index.html#t30-shoulder-surfed-interaction-url-enables-cross-device-pickup">T30:
```
eric-schuh, line 3261:
```suggestion
<a href="threat-model/index.html#t41-shoulder-surfed-interaction-url-enables-cross-device-pickup">T41:
```

Updating index to match the new one to prevent conflicts with #711

PR #705, opened 2026-08-18
Adding `credentialOffer` to VPR

eric-schuh:
This PR attempts to address https://github.com/w3c/vcalm/issues/689:

It adds a "credentialOffers" optional parameter to VPRs that allows a recipient to advertise which credential formats may result from the current interaction. It also provides the initial requester of a credential the ability to inform the receiving party which of the advertised formats they would like to receive the credential in, if possible. 

As of right now this PR does not include any additional parameters such as "requestedGroup", which was mentioned in one of the comments on the original issues.

PR #721, opened 2026-09-15
Apply git renormalize to mixed line ending files

ottonomy:
Fixes a line ending on 3 files. It was hard even to get confirmation of what exactly `git add --renormalize .` had changed here. Here to attempt to enable #705 to be merged.

```
echo "=== OLD blob line 18 (last bytes) ===" && git show 5838384 | sed -n '18p' | xxd | tail -2 && echo "" && echo "=== NEW blob line 18 (last bytes) ===" && git show 17a6204 | sed -n '18p' | xxd | tail -2
=== OLD blob line 18 (last bytes) ===
00000030: 206f 7220 6465 7269 7665 206f 7065 7261   or derive opera
00000040: 7469 6f6e 732c 206f 7220 696e 0d0a       tions, or in..

=== NEW blob line 18 (last bytes) ===
00000030: 206f 7220 6465 7269 7665 206f 7065 7261   or derive opera
00000040: 7469 6f6e 732c 206f 7220 696e 0a         tions, or in.
```

Concrete example (one line)
In components/VerifiableCredential.yml, this line:

`verifiable credential, such as in verify or derive operations, or in`

Before — the line ended with two bytes 0d 0a (\r\n, i.e. carriage return + line feed):

`00000040: 7469 6f6e 732c 206f 7220 696e 0d0a       tions, or in..`

After — the line ends with a single byte 0a (\n, line feed only):

`00000040: 7469 6f6e 732c 206f 7220 696e 0a         tions, or in.`

So the only difference on that line is the removal of the `0x0D` carriage return byte immediately before the newline.

Interestingly, these files were mixed line endings before the commit. Notice the unchanged context line right above ended in a plain `0a` (LF) already:


dlongley (approved):

TallTed (approved):

kezike (approved):
Thanks @ottonomy! I know we try not to modify files in the `transitions` folder, since it contains locked snapshots of the spec at a particular point in the past, but it makes sense in this case.

msporny (approved):

